PR #10, opened 2019-05-07
Add osconfig agent packaging scripts and docker file

iamsubratp:


agent-packaging/packaging-scripts/90-google-osconfig-agent.preset
```diff
@@ -0,0 +1 @@
+enable google-osconfig-agent.service
```
hopkiw: this file already exists in root of repo; why are you adding a new one instead of moving?
why rename packaging/ to agent-packaging/ ?

agent-packaging/packaging-scripts/build_packages_deb.sh
```diff
@@ -15,22 +15,15 @@
...
-BASE_REPO=$(curl -f -H Metadata-Flavor:Google ${URL}/base-repo)
+cd osconfig
 
-apt-get install -y git-core 
-git clone "https://github.com/${BASE_REPO}/compute-image-tools.git"
-cd compute-image-tools/cli_tools/google-osconfig-agent 
-packaging/setup_deb.sh 
-gsutil cp /tmp/debpackage/google-osconfig-agent*.deb "${GCS_PATH}/" 
+source ./agent-packaging/packaging-scripts/setup_deb.sh
+gsutil cp /tmp/debpackage/google-osconfig-agent*.deb "gs://osconfig-agent-package/"
```
hopkiw: two nits, the hardcoded gs path and a top level bucket just for osconfig. i think we will probably want some hierarchy for builds so we can get every built artifact while also having a 'destination' bucket with folders for things which should be injected to repos. as written we'd always overwrite for new builds of same version.

agent-packaging/packaging-scripts/build_packages_deb.sh
```diff
@@ -15,22 +15,15 @@
...
-{
-  echo "build failed"
-  exit 1
-}
+echo "started build..."
 
-trap exit_error ERR
+apt-get -y update && apt-get -y upgrade
+apt-get install -y git-core
+git clone --branch agent-build "https://github.com/GoogleCloudPlatform/osconfig.git"
```
hopkiw: why is the branch specified? i think you need a way to get this info from the build, if we're going to use this in presubmits

agent-packaging/packaging-scripts/common.sh
```diff
@@ -21,11 +23,13 @@ export GOPATH=/usr/share/gocode
 export GOCACHE=/tmp/.cache
 
 working_dir=${PWD}
-if [[ $(basename "$working_dir") != $NAME ]]; then
+if [[ $(basename "$working_dir") != "osconfig" ]]; then
```
hopkiw: you can delete this check..it was specific to the c-i-p repo to ensure you ran it from the right location for rel paths to work.

agent-packaging/packaging-scripts/google-osconfig-agent.conf
```diff
@@ -0,0 +1,5 @@
+description "Google OSConfig Agent"
```
hopkiw: same deal here, this is an existing file

agent-packaging/packaging-scripts/setup_deb.sh
```diff
@@ -13,31 +13,31 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-source packaging/common.sh 
 
-working_dir=${PWD}
+set -e
 
-sudo apt-get update
+echo "running common script..."
```
hopkiw: dont think we need this

agent-packaging/packaging-scripts/90-google-osconfig-agent.preset
iamsubratp: this came from my initial idea where i wanted to keep the packaging for all agents in a single place and not just osconfig-agent. I should have reverted it, since we not using that model. will revert it

agent-packaging/packaging-scripts/build_packages_deb.sh
iamsubratp: ah... this should have been removed. it was just to test my changes.. will remove it

agent-packaging/packaging-scripts/common.sh
iamsubratp: yeah... i thought i will delete it. but then i realized that to someone who is pretty new to this package build, this check actually provides useful information from where to kick off an agent build. This is only for manual testing of builds (I myself fell into this, i did not know from where to invoke the build scripts. this was specially because the subsequent scripts use the relative path from ${PWD}).

agent-packaging/packaging-scripts/google-osconfig-agent.conf
iamsubratp: so, the packaging already had 2 version of these files, i actually did not add/remove anything from the packaging code. i just moved it from `packaging` directory to `agent-packaging/packaging-scripts`. I had a conversation with Andrew few days ago, and i remember we agreed on removing these duplicates.
anyhow, for now, i am reverting back everything to `packaging`.  i will send out a different PR to remove the duplicates.

agent-packaging/packaging-scripts/setup_deb.sh
iamsubratp: i think i added this for debugging. i think i added to mark the starting of the `setup_deb.sh`

packaging/build_packages_deb.sh
```diff
@@ -26,11 +26,21 @@ trap exit_error ERR
...
 BASE_REPO=$(curl -f -H Metadata-Flavor:Google ${URL}/base-repo)
+REPO=$(curl -f -H Metadata-Flavor:Google ${URL}/repo)
+PULL_REF=$(curl -f -H Metadata-Flavor:Google ${URL}/pull-ref)
+
+
+echo "started build..."
+
+apt-get -y update && apt-get -y upgrade
+apt-get install -y git-core
+git clone --branch ${PULL_REF} "https://github.com/${BASE_REPO}/${REPO}.git"
```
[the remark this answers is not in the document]
hopkiw: ex if you wanted to check out the code associated with a pull request, which is not yet merged to any real branch, you'd run something like:

```
git init
git fetch https://github.com/${BASE_REPO}/${REPO}.git
git fetch https://github.com/${BASE_REPO}/${REPO}.git pull/37/head:pull/37/head
git checkout pull/37/head
```

packaging/build_packages_deb.sh
```diff
@@ -26,11 +26,21 @@ trap exit_error ERR
...
 
-apt-get install -y git-core 
-git clone "https://github.com/${BASE_REPO}/osconfig.git"
 cd osconfig
-packaging/setup_deb.sh 
-gsutil cp /tmp/debpackage/google-osconfig-agent*.deb "${GCS_PATH}/" 
+
+packaging/setup_deb.sh
+source ./packaging/setup_deb.sh
```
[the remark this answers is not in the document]
iamsubratp: bad merge :(.  removed it

packaging/build_packages_goo.sh
```diff
@@ -26,11 +26,15 @@ trap exit_error ERR
...
+PULL_REF=$(curl -f -H Metadata-Flavor:Google ${URL}/pull-ref)
 
 apt-get install -y git-core
-git clone "https://github.com/${BASE_REPO}/osconfig.git" 
+
+git clone --branch ${PULL_REF} "https://github.com/${BASE_REPO}/${REPO}.git"
+
 cd osconfig
-packaging/setup_goo.sh
+source ./packaging/setup_goo.sh
```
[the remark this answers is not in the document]
iamsubratp: thanks for asking. i just learnt read through the difference between sourcing and executing.
source'ing is not required here. will just remove it

packaging/build_packages_deb.sh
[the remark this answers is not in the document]
iamsubratp: at first; it is actually PULL_BASE_REF(not sure why i used PULL_REF; may have been just a typo); It  is the "Ref name of the base branch" according to prow. The example on the prow page, led me to the conclusion that it will be the branch name of the repo that we check'd out. i really did not test this part on the prow which is why i missed the ${PULL_REF} failure.

all we want here is to checkout to the branch that was used to send PR.

i think this might not work with BASE_REPO being set to GoogleCloudPLatform since the PR is sent from a forked repo.
https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md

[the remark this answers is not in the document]
hopkiw: all of the refs will be relative to our repo, don't worry. when you file a PR, a special ref is created on your repo to point to the commit-to-be-merged. prow knows how to get these, and we need to support something somewhat similar for our builds. maybe we can use the clone util from prow directly, or with slight modification?

anyway here's an example of what's set in a PR for refs:
```
  refs:
    base_link: https://github.com/GoogleCloudPlatform/guest-logging-go/commit/5d4770fc537a2976761bcc1a78a3c41085246284
    base_ref: master
    base_sha: 5d4770fc537a2976761bcc1a78a3c41085246284
    org: GoogleCloudPlatform
    pulls:
    - author: hopkiw
      author_link: https://github.com/hopkiw
      commit_link: https://github.com/GoogleCloudPlatform/guest-logging-go/pull/3/commits/f2714eb104c424b90ecb0b9e07a87e49f4ab1b68
      link: https://github.com/GoogleCloudPlatform/guest-logging-go/pull/3
      number: 3
      sha: f2714eb104c424b90ecb0b9e07a87e49f4ab1b68
```

packaging/google-osconfig-agent.conf
```diff
@@ -0,0 +1,5 @@
+description "Google OSConfig Agent"
```
hopkiw: you were going to remove these files from the PR, right?

packaging/build_packages_deb.sh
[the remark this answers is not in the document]
iamsubratp: played around with git PR refs.
I made the packaging scripts agnostic to prow job type. Based on the prow job type, the "pull ref" will be added to the daisy args.

packaging/google-osconfig-agent.conf
iamsubratp: removed it

packaging/build_packages_goo.sh
```diff
@@ -26,11 +26,14 @@ trap exit_error ERR
...
 BASE_REPO=$(curl -f -H Metadata-Flavor:Google ${URL}/base-repo)
+REPO=$(curl -f -H Metadata-Flavor:Google ${URL}/repo)
+PULL_REF=$(curl -f -H Metadata-Flavor:Google ${URL}/pull-ref)
 
 apt-get install -y git-core
-git clone "https://github.com/${BASE_REPO}/osconfig.git" 
-cd osconfig
-packaging/setup_goo.sh
+
+packaging/pull_repository.sh
```
hopkiw: by putting the pull steps in a pull_repository script instead of a function, the variables set in *this* script will not be available to the pull_repository script. you either have to `export` all needed variables here, or use functions defined in and sourced from pull_repository

packaging/pull_repository.sh
```diff
@@ -0,0 +1,37 @@
...
+
+trap exit_error ERR
+
+# pull the repository from github - start
+mkdir -p $REPO
+cd $REPO
+git init
+
+# fetch only the branch that we want to build
+GIT_CMD="git fetch https://github.com/${BASE_REPO}/${REPO}.git $PULL_REF"
```
hopkiw: 'fetch' is only the second step in setting up a repo, you should give it a local name, and then checkout. even this is not quite robust but is probably good enough for simply building artifacts

packaging/build_packages_goo.sh
iamsubratp: exporting all the variables since they can be used in other subsequent scripts

packaging/pull_repository.sh
iamsubratp: changed the script to fetch the pull refs

PR #12, opened 2019-05-10
Fix a few bugs in patch

adjackura:
Actually add new patch run to state when they are acked
Fix issue with needs-restarting on el6
Be smarter about running shutdown on Windows.

### DIFF
--- a/ospatch/updates_linux.go
+++ b/ospatch/updates_linux.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"syscall"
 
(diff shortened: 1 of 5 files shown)


ospatch/updates_linux.go
```diff
@@ -68,22 +67,27 @@ func systemRebootRequired() (bool, error) {
...
 	case packages.ZypperExists:
-		// TODO: implement
 		logger.Errorf("systemRebootRequired not implemented for zypper")
 		return false, nil
 	}
+	// TODO: implement something like this for rpm based distros to fall back to:
+	// https://bugzilla.redhat.com/attachment.cgi?id=1187437&action=diff
 
-	return false, fmt.Errorf("no recognized package manager installed, can't if reboot is required")
+	return false, fmt.Errorf("no recognized package manager installed, can't determin if reboot is required")
```
ryanwe: typo: determine

PR #16, opened 2019-05-13
Update to shared logger library

hopkiw:
Update to use guest logging library


go.mod
```diff
@@ -4,6 +4,7 @@ go 1.12
 
 require (
 	cloud.google.com/go v0.38.0
+	github.com/GoogleCloudPlatform/compute-image-tools v0.0.0-20190513155852-4d0828004f36
```
adjackura: run go mod tidy, github.com/GoogleCloudPlatform/compute-image-tools shouldn't be a needed import

PR #18, opened 2019-05-13
Add basic retries, update inventory after a patch run

adjackura:


ospatch/updates_windows.go
```diff
@@ -239,7 +241,7 @@ func runUpdates(r *patchRun) error {
 			return err
 		}
 
-		if err := packages.InstallGooGetUpdates(); err != nil {
+		if err := retry(3*time.Minute, "installing package updates", packages.InstallGooGetUpdates); err != nil {
```
ryanwe: We'll want to ReportInstanceDetails between reruns at some point. But this is fine for now.

PR #19, opened 2019-05-15
Update the gpg key file

iamsubratp:


packaging/build_packages_rpm.sh
```diff
@@ -46,7 +46,7 @@ case ${RELEASE} in
   6*) el6_install;;
   7*) yum -y install https://rhel7.iuscommunity.org/ius-release.rpm;;
 esac
-rpm --import /etc/pki/rpm-gpg/IUS-COMMUNITY-GPG-KEY
+rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-IUS-7
```
adjackura: I assume this will fail for el6 (I bet theres a /etc/pki/rpm-gpg/RPM-GPG-KEY-IUS-6 key?), you probably need to add this into the el6_install function and add a el7_install function as well

iamsubratp: check the rpm gpg import on rhel-6. refactoring the current script to accommodate the changes

packaging/build_packages_rpm.sh
```diff
@@ -37,16 +41,30 @@ el6_install(){
     n=$[$n+1]
     sleep 5
   done
+
+  import_gpg "6"
+}
+
+el7_install_import(){
+  n=0
+  while ! yum -y install https://rhel7.iuscommunity.org/ius-release.rpm; do
```
adjackura: This was a typo on my part, these should be https://centos7.iuscommunity.org/ius-release.rpm and https://centos6.iuscommunity.org/ius-release.rpm

iamsubratp:
fixed in https://github.com/GoogleCloudPlatform/osconfig/commit/9471fba891cd90f70adbc2122bf2bf2050930ca1

PR #20, opened 2019-05-16
 Respect apt and yum special options for patch + bug fixes

adjackura:
Yum and Apt settings are now respected for patch runs
Inventory logs debug instead of info
Inventory reads from apt-get full-update
Cleaned up Windows WUA logging
Patch run steps now automatically log
Only retry on certain API errors

### DIFF
--- a/ospatch/patch_run.go
+++ b/ospatch/patch_run.go
@@ -30,6 +30,7 @@ import (
 	"github.com/GoogleCloudPlatform/osconfig/tasker"
 	"github.com/golang/protobuf/jsonpb"
 	"google.golang.org/api/option"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
 	osconfigpb "github.com/GoogleCloudPlatform/osconfig/_internal/gapi-cloud-osconfig-go/google.golang.org/genproto/googleapis/cloud/osconfig/v1alpha1"
@@ -43,7 +44,7 @@ const (
 	acked          = "acked"
 	prePatchReboot = "prePatchReboot"
 	patching       = "patching"
-	complete       = "complete"
+	reportSucceded = "reportSucceded"
 )
 
 var (
@@ -54,6 +55,10 @@ func initPatch(ctx context.Context) {
 	cancelC = make(chan struct{})
 	disableAutoUpdates()
 	go Run(ctx, cancelC)
+	// Sleep just long enough for Run to register any pending patches.
+	// TODO: Find a cleaner way to ensure any pending patch runs start before
+	// other tasks immediately after startup.
+	time.Sleep(1 * time.Second)
 }
 
 // Configure manages the background patch service.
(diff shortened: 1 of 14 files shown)


ospatch/patch_run.go
```diff
@@ -415,7 +422,15 @@ func (r *patchRun) reportPatchDetails(patchState osconfigpb.Instance_PatchState,
 		res, err := r.client.ReportPatchJobInstanceDetails(r.ctx, &request)
 		if err != nil {
 			if s, ok := status.FromError(err); ok {
-				return fmt.Errorf("code: %q, message: %q, details: %q", s.Code(), s.Message(), s.Details())
+				err := fmt.Errorf("code: %q, message: %q, details: %q", s.Code(), s.Message(), s.Details())
+				switch s.Code() {
+				// Errors we should retry.
+				case codes.DeadlineExceeded, codes.Unavailable, codes.Aborted, codes.Internal:
```
ryanwe: We should also retry ResourceExhausted errors, ideally with an extended, jittered wait interval.

ospatch/updates_linux.go
```diff
@@ -92,10 +94,49 @@ func systemRebootRequired() (bool, error) {
 	return false, fmt.Errorf("no recognized package manager installed, can't determine if reboot is required")
 }
 
+func updatePackages(r *patchRun) error {
+	var errs []string
+	if packages.AptExists {
+		var opts []packages.AptGetUpgradeOption
+		if r.Job.ReportPatchJobInstanceDetailsResponse.PatchConfig.Apt != nil {
```
ryanwe: Can this be combined with `if packages.AptExtists && ...` to remove an extra level of nesting? (Same for Yum & Zypper).

ospatch/updates_windows.go
```diff
@@ -185,33 +179,38 @@ func installWUAUpdates(r *patchRun) error {
...
+		logger.Debugf("Filtering by classifications: %q\n", r.Job.PatchConfig.WindowsUpdate.Classifications)
 
 		for _, e := range r.Job.PatchConfig.WindowsUpdate.Excludes {
 			excludes[e] = struct{}{}
 		}
+		logger.Debugf("Filtering out KBs: %q\n", r.Job.PatchConfig.WindowsUpdate.Excludes)
 	}
 
 	for i := 0; i < int(count); i++ {
+		if err := r.reportContinuingState(osconfigpb.Instance_APPLYING_PATCHES); err != nil {
```
ryanwe: What are the chances this leads to DOSing? If that's a possibility, it would be better to do this once before the loop. We can do this at each step when we know it's safe and users want that level of responsiveness.

ospatch/updates_linux.go
adjackura: APT existing on the system means we will patch APT, that doesn't mean there are any settings under PatchConfig.Apt, so it could be nil

ospatch/updates_windows.go
adjackura: Windows updates are slow to install, I don't see this breaking 1QPS

ospatch/updates_linux.go
ryanwe: Ah, this is only for adding additional args. Got it :+1:

ospatch/updates_windows.go
ryanwe: Chatted offline and decided that we should be fine.

PR #22, opened 2019-05-24
Packaging

hopkiw:
This is based on @iamsubratp PR, but moved common code to functions and tested on all platforms.


hopkiw:
misnamed secret, misnamed workflow file, yada yada
/packagebuild

hopkiw:
hmm..invalid service account key..?
/packagebuild

hopkiw:
maybe wrong cacerts.
/packagebuild

hopkiw:
getting somewhere. fixed ca-certs, found insufficient permissions. updated role for service account
/packagebuild

hopkiw:
update service account with service account user role...
/packagebuild

hopkiw:
seems to be running this time. v time consuming, though.. :( 
probably shouldn't capture the output from daisy, you don't get any updates in the build log until completion.

hopkiw:
gotten hung in some way, gcs artifacts aren't being updated anymore and without stdout it's difficult to see why.. updated container image with new logic, new perms
/packagebuild

hopkiw:
big progress!
```
[Daisy] Workflow "build-packages" finished
[Daisy] All workflows completed successfully.
```

hopkiw:
actually getting use out of this now and found some issue..

/packagebuild

hopkiw:
there was one case where try_command failed and that somehow didn't trigger the trap on ERR, so the instance was hung forever. updated overall timeout relative to how long i've observed this to take on successful runs, but will want to figure out that case later.

packaging/googet/google-osconfig-agent.goospec
```diff
@@ -30,7 +30,6 @@
       ]
   }],
   "build": {
-    "linux": "/bin/bash",
-    "linuxArgs": ["-c", "GOOS=windows /tmp/go/bin/go build -ldflags='-X main.version={{.version}}' -mod=readonly -o google_osconfig_agent.exe"]
```
adjackura: This shouldn't need to be changed as long as you goopack line from packaging/build_goo.sh

packaging/build_goo.sh
```diff
@@ -13,8 +13,18 @@
...
+. packaging/common.sh
+
+echo "Installing go"
+install_go
+
+# Pull go deps
+$GO mod download
+
+# Install dependencies.
+$GO get github.com/google/googet/goopack
```
adjackura: Keep the github.com/google/googet/v2/goopack

PR #24, opened 2019-05-28
 Fix up reboot indicators and logging

adjackura:


ospatch/updates_windows.go
```diff
@@ -28,16 +28,20 @@ import (
 	osconfigpb "github.com/GoogleCloudPlatform/osconfig/_internal/gapi-cloud-osconfig-go/google.golang.org/genproto/googleapis/cloud/osconfig/v1alpha1"
 )
 
-func systemRebootRequired() (bool, error) {
+func (r *patchRun) systemRebootRequired() (bool, error) {
+	reg := `HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\WindowsUpdate\Auto Update\RebootRequired`
```
ryanwe: consider making `SOFTWARE\Microsoft\Windows\CurrentVersion\WindowsUpdate\Auto Update\RebootRequired` a variable that you reuse instead of copy/pasting that long string. It'll be easier to verify they're the same string.

PR #25, opened 2019-06-04
Don't restart service on package upgrade

adjackura:
Agent now reacts to restart required signal and stops, the various service managers handle the restart automatically.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -48,8 +48,12 @@ const (
 	osPatchEnabledDefault     = false
 	debugEnabledDefault       = false
 
-	osPatchStateFileWindows = `C:\Program Files\Google\OSConfig\osconfig_patch.state`
-	osPatchStateFileLinux   = "/etc/osconfig/osconfig_patch.state"
+	configDirWindows        = `C:\Program Files\Google\OSConfig`
+	configDirLinux          = "/etc/osconfig"
+	osPatchStateFileWindows = configDirWindows + `\osconfig_patch.state`
+	osPatchStateFileLinux   = configDirLinux + "/osconfig_patch.state"
+	restartFileWindows      = configDirWindows + `\restart_required`
+	restartFileLinux        = configDirLinux + "/restart_required"
 
 	osConfigPollIntervalDefault = 10
 )
@@ -389,4 +393,13 @@ func PatchStateFile() string {
 	}
 
 	return osPatchStateFileLinux
+}
+
+// RestartFile is the location of the restart required file.
+func RestartFile() string {
+	if runtime.GOOS == "windows" {
+		return restartFileWindows
+	}
+
+	return restartFileLinux
 }
--- a/config/config.go
+++ b/config/config.go
@@ -52,8 +52,8 @@ const (
 	configDirLinux          = "/etc/osconfig"
 	osPatchStateFileWindows = configDirWindows + `\osconfig_patch.state`
 	osPatchStateFileLinux   = configDirLinux + "/osconfig_patch.state"
-	restartFileWindows      = configDirWindows + `\restart_required`
-	restartFileLinux        = configDirLinux + "/restart_required"
+	restartFileWindows      = `C:\Windows\Temp\osconfig_agent_restart_required`
+	restartFileLinux        = "/tmp/osconfig_agent_restart_required"
 
 	osConfigPollIntervalDefault = 10
 )
--- a/google-osconfig-agent.service
+++ b/google-osconfig-agent.service
@@ -2,6 +2,8 @@
 Description=Google OSConfig Agent
 After=local-fs.target network-online.target network.target
 Wants=local-fs.target network-online.target network.target
+StartLimitInterval=120
+StartLimitBurst=3
 
 [Service]
 ExecStart=/usr/bin/google_osconfig_agent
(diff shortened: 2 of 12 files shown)


config/config.go
```diff
@@ -48,8 +48,12 @@ const (
 	osPatchEnabledDefault     = false
 	debugEnabledDefault       = false
 
-	osPatchStateFileWindows = `C:\Program Files\Google\OSConfig\osconfig_patch.state`
-	osPatchStateFileLinux   = "/etc/osconfig/osconfig_patch.state"
+	configDirWindows        = `C:\Program Files\Google\OSConfig`
+	configDirLinux          = "/etc/osconfig"
+	osPatchStateFileWindows = configDirWindows + `\osconfig_patch.state`
+	osPatchStateFileLinux   = configDirLinux + "/osconfig_patch.state"
```
hopkiw: `path.Join` ?

config/config.go
```diff
@@ -387,3 +391,12 @@ func PatchStateFile() string {
 
 	return osPatchStateFileLinux
 }
+
+// RestartFile is the location of the restart required file.
```
hopkiw: vanity function, move this check into the body code

google-osconfig-agent.service
```diff
@@ -6,7 +6,7 @@ Wants=local-fs.target network-online.target network.target
 [Service]
 ExecStart=/usr/bin/google_osconfig_agent
 Restart=always
-RestartSec=120
+RestartSec=1
```
hopkiw: what's the background for this? are you sure you always want instant restarts? what's systemd behavior for infinite loops?

main.go
```diff
@@ -56,6 +56,16 @@ func run(ctx context.Context) {
 			logger.Errorf(err.Error())
 		}
 
+		if _, err := os.Stat(config.RestartFile()); err == nil {
+			logger.Infof("Restart signal recieved, waiting for tasks to complete.")
```
hopkiw: np `recieved` to `received`

hopkiw: think this should indicate more explicitly what's going on - no signal has actually been received, a marker file was detected.

packaging/debian/rules
```diff
@@ -15,6 +15,7 @@ override_dh_auto_install:
 	# Binary-only package.
 	dh_auto_install -- --no-source
 	mv debian/google-osconfig-agent/usr/bin/osconfig debian/google-osconfig-agent/usr/bin/google_osconfig_agent
+	dh_installdirs /etc/osconfig
```
hopkiw: if we're declaring that this directory should always exist when osconfig agent is installed, we should just ship a file into it and not have type-specific logic for creating empty directories. it can be a readme file.

packaging/google-osconfig-agent.spec
```diff
@@ -38,6 +38,7 @@ Contains the OSConfig agent binary and startup scripts
 GOPATH=%{_gopath} CGO_ENABLED=0 %{_go} build -ldflags="-s -w -X main.version=%{_version}" -mod=readonly -o google_osconfig_agent
 
 %install
+install -d %{buildroot}/etc/osconfig
```
hopkiw: see my comment on empty dirs above, but also i think you'll need a corresponding entry in %files for this to have any effect

config/config.go
```diff
@@ -48,8 +48,12 @@ const (
 	osPatchEnabledDefault     = false
 	debugEnabledDefault       = false
 
-	osPatchStateFileWindows = `C:\Program Files\Google\OSConfig\osconfig_patch.state`
-	osPatchStateFileLinux   = "/etc/osconfig/osconfig_patch.state"
+	configDirWindows        = `C:\Program Files\Google\OSConfig`
+	configDirLinux          = "/etc/osconfig"
```
hopkiw: if this is going to have 'config', ok, but if it's only going to have a marker file, we should be creating that in /, /var/tmp or similar.

config/config.go
adjackura: Constants can't have function calls.

config/config.go
adjackura: Not sure what you mean by this?

config/config.go
adjackura: Agreed, I think storing it in temp would be better

google-osconfig-agent.service
adjackura: Fixed with StartLimitInterval and StartLimitBurst

packaging/debian/rules
adjackura: removed

packaging/google-osconfig-agent.spec
adjackura: removed

config/config.go
adjackura: Just to make it clearer, I just moved the restart file to /tmp and c:\windows\temp, the config dir is still used for state like it was previously was , but that should probably moved to /var/local

config/config.go
hopkiw: this entire function that is switching between two constants based on platform. it's used in one place, and so it essentially replaces an if check with a function call. just do this if check there and remove this function

adjackura: Store store these constants outside of the config package, or make two constants public? Making both of them public expands the public surface of this package making it more complex.

adjackura: Spoke offline, configuration/constant management is probably something we can improve later on in the agent.

PR #26, opened 2019-06-04
Fix inventory respecting old metadata key

adjackura:


inventory/inventory.go
```diff
@@ -110,6 +108,8 @@ func Get() *InstanceInventory {
 		logger.Errorf("packages.GetPackageUpdates() error: %v", err)
 	}
 
+	hs.LastUpdated = time.Now().UTC().Format(time.RFC3339)
```
ryanwe: This ship has probably sailed but understanding this better, I think `LastInventoryCheckCompleted` would be more accurate.

PR #27, opened 2019-06-10
Don't close logger before initializing, fix check for yum-cron.service

adjackura:


ospatch/system_linux.go
```diff
@@ -35,7 +35,7 @@ func disableAutoUpdates() {
 		if err != nil {
 			if eerr, ok := err.(*exec.ExitError); ok {
 				// Error code of 1 indicates disabled.
```
ryanwe: Does the comment mean "was already disabled" or "was successfully disabled" or both?

adjackura: Is currently disabled, the above check is just running "is-enabled"

PR #28, opened 2019-06-11
Fix logging initialization

adjackura:


main.go
```diff
@@ -91,6 +91,11 @@ func main() {
 	flag.Parse()
 	ctx := context.Background()
 
+	// If this call to SetConfig fails (like a metadata error) we can't continue.
+	if err := config.SetConfig(); err != nil {
```
ryanwe: Fwiw, the need to call this before doing other things has lead to a few bugs already. We may want to consider moving this to an init call or using lazy initialization of the config values.

PR #30, opened 2019-06-11
Keep installing WUA updates until there are no updates left

adjackura:


ospatch/updates_windows.go
```diff
@@ -166,60 +166,61 @@ func (r *patchRun) installWUAUpdates() error {
...
-		return fmt.Errorf("GetWUAUpdateCollection error: %v", err)
-	}
-
-	countRaw, err := updts.GetProperty("Count")
-	if err != nil {
-		return err
-	}
-	count, _ := countRaw.Value().(int32)
+	// We keep searching for and installing updates until the count == 0 or there is an error.
+	for {
```
ryanwe: Should we put a max updates here just to be safe? It could even be a high number like 500 or 1000. Just to make sure that we never enter a weird state where we detect an update over and over again (maybe because it fails to install in some weird way).

ospatch/updates_windows.go
```diff
@@ -167,7 +167,7 @@ func (r *patchRun) installWUAUpdates() error {
 	defer session.Release()
 
 	// We keep searching for and installing updates until the count == 0 or there is an error.
-	for {
```
ryanwe: nit: can you use a const here instead of a magic number?

PR #31, opened 2019-06-19
Recipes

hopkiw:
First draft for recipe support

Using a file naming pattern based on name of major function
Stubs for artifact support - separate PR incoming from Trevor
Lots of things still in draft or development mode, but fine to merge as nothing references them.


ospackage/recipes/BuildCommand.go
```diff
@@ -0,0 +1,100 @@
...
+
+	res := []string{path}
+	res = append(res, step.FileExec.Args...)
+	return res, nil
+}
+
+func StepScriptRun(step *osconfigpb.SoftwareRecipe_Step_ScriptRun, artifacts map[string]string) ([]string, error) {
+	// TODO: should be putting this in stepN_type/ dir, but that needs me to know the dir way in advance..
+	// actually this is an artifact. we should have made them referenced as artifacts.. we'll have to repeat file creation logic here.
+	f, err := os.Create("/tmp/scriptrun")
```
adjackura: ioutil.TempDir()

hopkiw: this is a placeholder as mentioned in the todo..we don't want real 'temp' dirs.

ospackage/recipes/InstallRecipe.go
```diff
@@ -0,0 +1,62 @@
...
+		cmdObj := exec.Command(cmd[0], cmd[1:]...)
+		dirName := recipe.Name
+		if recipe.Version != "" {
+			dirName = fmt.Sprintf("%s_%s", dirName, recipe.Version)
+		}
+		cmdObj.Dir = path.Join(RECIPE_BASE_PATH, dirName, "runId", "stepName")
+		if err := os.MkdirAll(cmdObj.Dir, os.ModeDir|0755); err != nil {
+			return fmt.Errorf("Failed to create working dir for step %d: %s", idx, err)
+		}
+		cmdObj.Env = append(cmdObj.Env, fmt.Sprintf("RECIPE_NAME=%s", recipe.Name))
```
adjackura: Since you know all the elements why not just append them all at once to avoid this the reallocation?

ospackage/recipes/RecipeDB.go
```diff
@@ -0,0 +1,81 @@
...
+	// ${ROOT}/recipe[_ver]/runId/stepN_type/
+	if version == "" {
+		return []int{0}, nil
+	}
+	var ret []int
+	for idx, element := range strings.Split(version, ".") {
+		if idx > 3 {
+			return nil, fmt.Errorf("Invalid version string")
+		}
+		val, err := strconv.ParseUint(element, 10, 0)
```
adjackura: strconv.Atoi() ?

hopkiw: from go doc:
```
func Atoi(s string) (int, error)
    Atoi is equivalent to ParseInt(s, 10, 0), converted to type int.
```
but these specifically need to be unsigned ints; we would otherwise validate 1.-2.0 as a valid version.

ospackage/recipes/InstallRecipe.go
hopkiw: what's the performance penalty ? i was erring on the side of readability but if it's worth the optimization i could do that.

ospackage/recipes/RecipeDB.go
adjackura: Ack

ospackage/recipes/InstallRecipe.go
adjackura: Fairly minimal, that's why its not common practice to create slices with the required length before the append loop you have below. My suggestion to keep the readability (and less repitition) would be something like:
envs := []string{
fmt.Sprintf("RECIPE_NAME=%s", recipe.Name)
...
...
}

cmdObj.Env = append(cmdObj.Env, envs...)

adjackura:
File name convention is lower case with _ as the separator.

ospackage/recipes/BuildCommand.go
```diff
@@ -0,0 +1,108 @@
+package recipes
```
adjackura: Missing License at top of the files.

PR #35, opened 2019-07-10
 Move ospackage to guest policies and v1alpha2

adjackura:


main.go
```diff
@@ -134,8 +134,8 @@ func main() {
 		inventory.Run()
 		tasker.Close()
 		return
-	case "ospackage":
-		ospackage.Run(ctx, config.Instance())
+	case "policies":
```
ryanwe: since this will be in our docs, can we support "guestpolicies" too? And consider continuing to support "ospackage" for now so existing users don't need to re-enable the feature.

policies/policies.go
```diff
@@ -0,0 +1,301 @@
+//  Copyright 2018 Google Inc. All Rights Reserved.
```
ryanwe: 2019

policies/policies.go
```diff
@@ -0,0 +1,301 @@
...
+		if r := repo.GetPackageRepository().GetZypper(); r != nil {
+			zypperRepos = append(zypperRepos, r)
+			continue
+		}
+	}
+
+	var aptInstalled, aptRemoved, aptUpdated []*osconfigpb.Package
+	var yumInstalled, yumRemoved, yumUpdated []*osconfigpb.Package
+	var zypperInstalled, zypperRemoved, zypperUpdated []*osconfigpb.Package
+	var gooInstalled, gooRemoved, gooUpdated []*osconfigpb.Package
```
ryanwe: minor: Since this is desired state, consider names like "aptToInstall". This will help contrast these values with the maps of what is actually already installed or available for update.

policies/policies.go
```diff
@@ -0,0 +1,301 @@
...
+			case osconfigpb.DesiredState_REMOVED:
+				gooRemoved = append(gooRemoved, pkg.GetPackage())
+				aptRemoved = append(aptRemoved, pkg.GetPackage())
+				yumRemoved = append(yumRemoved, pkg.GetPackage())
+				zypperRemoved = append(zypperRemoved, pkg.GetPackage())
+			case osconfigpb.DesiredState_UPDATED:
+				gooUpdated = append(gooUpdated, pkg.GetPackage())
+				aptUpdated = append(aptUpdated, pkg.GetPackage())
+				yumUpdated = append(yumUpdated, pkg.GetPackage())
+				zypperUpdated = append(zypperUpdated, pkg.GetPackage())
```
ryanwe: consider DRYing this up with a function like `func addPackage(goo, apt, yum, zypper) (goo, apt, yum, zypper)`

policies/policies.go
```diff
@@ -0,0 +1,301 @@
...
+			}
+		case osconfigpb.Package_ZYPPER:
+			switch pkg.GetPackage().GetDesiredState() {
+			case osconfigpb.DesiredState_INSTALLED:
+				zypperInstalled = append(zypperInstalled, pkg.GetPackage())
+			case osconfigpb.DesiredState_REMOVED:
+				zypperRemoved = append(zypperRemoved, pkg.GetPackage())
+			case osconfigpb.DesiredState_UPDATED:
+				zypperUpdated = append(zypperUpdated, pkg.GetPackage())
+			}
```
ryanwe: here too I think you can factor out the block of these case statements into a single function call. Alternatively, you could create a new interface with some functions that work for each package manager type.

policies/policies.go
```diff
@@ -0,0 +1,301 @@
...
+				logger.Errorf("Error creating repo file: %v", err)
+				errs = append(errs, fmt.Sprintf("error creating zypper repo file: %v", err))
+			}
+		}
+		if err := zypperRepositories(zypperRepos, config.ZypperRepoFilePath()); err != nil {
+			logger.Errorf("Error writing zypper repo file: %v", err)
+			errs = append(errs, fmt.Sprintf("error writing zypper repo file: %v", err))
+		}
+		if err := zypperChanges(zypperInstalled, zypperRemoved, zypperUpdated); err != nil {
+			errs = append(errs, fmt.Sprintf("error performing zypper changes: %v", err))
```
ryanwe: Factoring this out into package manager based interface methods may help remove the duplication here too.

policies/policies.go
adjackura: I did some work factoring this out, I'll want to do some more and add better tests but that's probably best left for a future PR

policies/policies.go
adjackura: This file stayed largely the same, just name change so we want to keep the contents under the old copyright

PR #41, opened 2019-07-23
Install and remove each package by itself

ryanwe:
- This avoids problems where a single package installation failure prevents all packages from being installed. For example, without this change, apt-get install good, another-good, bad will not install any
packages.


adjackura:
It would probably be better to handle this in the the InstallPackages functions, try to install them all together like we have right now (way more efficient) and falling back to one at a time if there is an error and only returning an error if one or more individual package installs fail.

PR #42, opened 2019-07-24
adding recipedb code

TrevorBoone:


policies/recipes/recipedb.go
```diff
@@ -18,22 +18,60 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"encoding/json"
+	"io/ioutil"
+)
+
+var (
+	dbPath := "~/placeholder_location/swr.json"
```
TrevorBoone: Where should the recipedb file live on the different operating systems?

policies/recipes/recipedb.go
```diff
@@ -15,25 +15,69 @@
...
+		return RecipeDB{}, err
+	}
+	return db, nil
 }
 
 // GetRecipe returns the Recipe object for the given recipe name.
 func (db *RecipeDB) GetRecipe(name string) (Recipe, bool) {
-	return Recipe{}, false
+	r, ok := db.recipes[name]
+	return r, ok
```
TrevorBoone: I went for the strategy where it will always write state to disk but only read from disk once. What's the concurrency model for this, should it read from disk all the time and does it need any locking mechanisms?

policies/recipes/recipedb.go
hopkiw: a state file that persists is probably appropriate to put somewhere under `/var/cache`, but we would want to look at the linux FHS and maybe some similar products to be sure. should be somewhere that indicates it isn't available for editing (i.e. not `/etc`) and not somewhere temporary (`/tmp`, `/var/run`, etc.)

policies/recipes/recipedb.go
hopkiw: it should not be possible for the same recipe name to be present in multiple policies, so reading once is probably ok.

will consider the write locking scenario, but it at least doesn't need it -yet-.

policies/recipes/recipedb.go
```diff
@@ -15,25 +15,69 @@
...
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"os"
 	"strconv"
 	"strings"
 )
 
+var (
+	dbPath = "~/placeholder_location/swr.json"
```
hopkiw: FYI expansion of ~ is done by the shell and won't work here, you'll need to pick a real placeholder until we get the real real :)

policies/recipes/recipedb.go
TrevorBoone: So I'm curious for your thoughts are but so far the best option I've found is /var/lib

TrevorBoone: Well from talking to you and more research the best I've found is /var/lib/google/. This seems to make sense from an overall filesystem perspective and also is similar to where programs, such as apt, put similar data.

policies/recipes/installrecipe.go
```diff
@@ -15,24 +15,35 @@
...
+	"github.com/golang/protobuf/jsonpb"
 )
 
 const (
 	recipeBasePath = "/tmp/osconfig_software_recipes"
 	// TODO: move to constants, split linux and windows.
 )
 
+var (
+	marshaller = jsonpb.Marshaler{}
```
hopkiw: why is it global?

TrevorBoone: My impression is that it is just configuration for the marshaler, which shouldn't vary at all so the thought was just to initialize it once. While looking I can't find any safety guarantees for it so I'll change it to creating a new one each time as the empty struct allocation should be negligible cost.

hopkiw:
bc of go.mod/go.sum files changing, i can't approve. /assign adjackura

policies/recipes/recipedb.go
```diff
@@ -15,25 +15,69 @@
...
+	versionNum, err := convertVersion(version)
+	if err != nil {
+		return err
+	}
+	db.recipes[name] = Recipe{name: name, version: versionNum}
+	dbBytes, err := json.Marshal(db)
+	if err != nil {
+		return err
+	}
+	f, err := os.Create(dbPath)
```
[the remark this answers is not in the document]
TrevorBoone: I switched it to writing to a temp file. Is os.TempDir an okay way to get the temporary directory or should I specify something like /tmp exmplicitly?

policies/recipes/recipedb.go
```diff
@@ -15,25 +15,69 @@
...
+func newRecipeDB() (RecipeDB, error) {
+	db := RecipeDB{}
+	f, err := os.Open(dbPath)
+	if err != nil {
+		return RecipeDB{}, err
+	}
+	defer f.Close()
+	bytes, err := ioutil.ReadAll(f)
+	if err != nil {
+		if os.IsNotExist(err) {
```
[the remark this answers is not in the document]
TrevorBoone: Done.

policies/recipes/recipedb.go
```diff
@@ -15,25 +15,69 @@
...
+
 // RecipeDB represents local state of installed recipes.
-type RecipeDB struct{}
+type RecipeDB struct {
+	recipes map[string]Recipe
+}
 
-func newRecipeDB() RecipeDB {
-	return RecipeDB{}
+func newRecipeDB() (RecipeDB, error) {
```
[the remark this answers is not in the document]
TrevorBoone: Done.

policies/recipes/recipedb.go
```diff
@@ -15,25 +15,69 @@
...
+	db := RecipeDB{}
```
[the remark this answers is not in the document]
TrevorBoone: Done.

policies/recipes/recipedb.go
```diff
@@ -15,32 +15,85 @@
...
+	db.recipes[name] = Recipe{name: name, version: versionNum, installTime: time.Now().Unix()}
+	dbBytes, err := json.Marshal(db)
+	if err != nil {
+		return err
+	}
+	if err := os.MkdirAll(filepath.Dir(dbPath), 0755); err != nil {
+		return err
+	}
+
+	f, err := os.Create(tempPath)
```
adjackura: The tempPath should be the same as the dbPath with just a temp string appended. Moves across file systems are not atomic.

policies/recipes/recipedb.go
```diff
@@ -15,32 +15,85 @@
...
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	if _, err = f.Write(dbBytes); err != nil {
+		return err
+	}
+	os.Rename(tempPath, dbPath)
```
adjackura: return os.Rename(tempPath, dbPath)

policies/recipes/recipedb.go
```diff
@@ -15,32 +15,85 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
+)
+
+var (
+	dbPath   = "/var/lib/google/osconfig_recipedb"
```
adjackura: Windows path?

TrevorBoone: I added support for Windows. I put it under C:\ProgramData\Google but I am not positive whether that is the correct place or not.

policies/recipes/recipedb.go
TrevorBoone: Done.

policies/recipes/recipedb.go
TrevorBoone: Okay, I put it in the same directory and am using ioutil.TempFile to create the file and choose the exact name.

PR #43, opened 2019-07-25
Update shebang and arg handling

hopkiw:
Check exit codes, remove unnecessary quoting, standardize interpreter handling

### DIFF
--- a/policies/recipes/steps.go
+++ b/policies/recipes/steps.go
@@ -22,7 +22,6 @@ import (
 	"path/filepath"
 	"runtime"
 	"strconv"
-	"strings"
 
 	osconfigpb "github.com/GoogleCloudPlatform/osconfig/_internal/gapi-cloud-osconfig-go/google.golang.org/genproto/googleapis/cloud/osconfig/v1alpha2"
 )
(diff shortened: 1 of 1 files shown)


policies/recipes/steps.go
```diff
@@ -144,26 +143,41 @@ func StepScriptRun(step *osconfigpb.SoftwareRecipe_Step_ScriptRun, artifacts map
...
 
-	var qargs []string
-	//if step.ScriptRun.Interpreter == osconfigpb.SoftwareRecipe_Step_RunScript_POWERSHELL {
-	for _, arg := range step.ScriptRun.Args {
-		qargs = append(qargs, fmt.Sprintf("%q", arg))
+	var command string
+	var args []string
+	switch step.ScriptRun.Interpreter {
+	case osconfigpb.SoftwareRecipe_Step_RunScript_POWERSHELL:
+		command = "PowerShell.exe"
```
adjackura: Use the full path here (C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe), you might even want to use the SystemRoot environment variable to grab the correct Windows directory but that may be overkill

Also do we want to have a check for Windows or note here? Otherwise the error would be odd.

policies/recipes/steps.go
```diff
@@ -144,26 +143,41 @@ func StepScriptRun(step *osconfigpb.SoftwareRecipe_Step_ScriptRun, artifacts map
...
+		args = append(args, []string{"-File", scriptPath}...)
+	case osconfigpb.SoftwareRecipe_Step_RunScript_SHELL:
```
adjackura: Wasn't this supposed to be cmd.exe on Windows as well?

policies/recipes/steps.go
hopkiw: the error will be that powershell.exe isn't found, right?

policies/recipes/steps.go
```diff
@@ -144,26 +144,44 @@ func StepScriptRun(step *osconfigpb.SoftwareRecipe_Step_ScriptRun, artifacts map
...
+	var command string
+	var args []string
+	switch step.ScriptRun.Interpreter {
+	case osconfigpb.SoftwareRecipe_Step_RunScript_POWERSHELL:
+		command = "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\PowerShell.exe"
+		args = append(args, []string{"-File", scriptPath}...)
+	case osconfigpb.SoftwareRecipe_Step_RunScript_SHELL:
+		command = "/bin/sh"
+		if runtime.GOOS == "windows" {
+			command = "cmd.exe"
```
adjackura: C:\Windows\System32\cmd.exe

policies/recipes/steps.go
adjackura: That's an obtuse error for the user though, a simple check would allow us to produce a much clearer error "PowerShell interpreter not supported on non Windows systems" or something like that.

policies/recipes/steps.go
```diff
@@ -127,106 +126,79 @@ func StepFileExec(step *osconfigpb.SoftwareRecipe_Step_FileExec, artifacts map[s
...
+		}
 	case osconfigpb.SoftwareRecipe_Step_RunScript_SHELL:
 		if runtime.GOOS == "windows" {
-			return scriptRunCmd(step, runEnvs, stepDir)
+			args = append([]string{"/c", cmd}, args...)
+			cmd = "C:\\Windows\\System32\\cmd.exe"
+		}
+		qargs := cmd
+		for _, arg := range args {
+			qargs = fmt.Sprintf("%s %q", qargs, arg)
```
hopkiw: the end result here has to be ["/bin/sh", "-c", "/path/to/recipe_script_source \"quoted arg\" \"quoted arg\""]; args have to end up with script name in one string at the end. unsure if cmd needs something similar, will investigate

PR #44, opened 2019-07-30
adding windows support to script and file execution

TrevorBoone:


policies/recipes/steps.go
```diff
@@ -131,27 +132,100 @@ func StepFileExec(step *osconfigpb.SoftwareRecipe_Step_FileExec, artifacts map[s
...
+	if err != nil {
+		return err
+	}
+
+	args := append([]string{"-File", scriptPath}, step.ScriptRun.Args...)
+	return executeCommand("PowerShell", stepDir, runEnvs, args...)
+}
+
+func scriptRunCmd(step *osconfigpb.SoftwareRecipe_Step_ScriptRun, runEnvs []string, stepDir string) error {
+	scriptPath := filepath.Join(stepDir, "script.bat")
```
TrevorBoone: as far as I can tell on windows it needs the right file extension to be able to run correctly on windows, is that correct?

policies/recipes/steps.go
```diff
@@ -131,27 +132,100 @@ func StepFileExec(step *osconfigpb.SoftwareRecipe_Step_FileExec, artifacts map[s
 
 // StepScriptRun builds the command for a ScriptRun step
 func StepScriptRun(step *osconfigpb.SoftwareRecipe_Step_ScriptRun, artifacts map[string]string, runEnvs []string, stepDir string) error {
```
TrevorBoone: I don't have access to a physical windows machine so I couldn't actually run any of these code paths, what is the best way to actually run this code without one? My current plan is just to spin up a windows VM on GCE and run it there.

policies/recipes/artifacts.go
```diff
@@ -67,11 +68,11 @@ func FetchArtifacts(ctx context.Context, artifacts []*osconfigpb.SoftwareRecipe_
 }
 
 func fetchArtifact(ctx context.Context, artifact *osconfigpb.SoftwareRecipe_Artifact, directory string) (string, error) {
-	localPath := filepath.Join(directory, artifact.Id)
 	var reader io.ReadCloser
-	var checksum string
+	var checksum, extension string
```
TrevorBoone: I added code to keep the extension for artifacts since those are meaningful on windows.

TrevorBoone:
One weird oddity I found is that the ExecFile step doesn't seem to work with .ps1 files as it seems like the default behavior for those is to open them in notepad.

policies/recipes/steps.go
```diff
@@ -131,27 +132,100 @@ func StepFileExec(step *osconfigpb.SoftwareRecipe_Step_FileExec, artifacts map[s
...
+	if runtime.GOOS == "windows" {
```
adjackura: I think this would look cleaner if you had this check wihin each case that it mattered.

policies/recipes/steps.go
```diff
@@ -131,27 +132,100 @@ func StepFileExec(step *osconfigpb.SoftwareRecipe_Step_FileExec, artifacts map[s
...
+	default:
+		return fmt.Errorf("invalid interpreter %q", step.ScriptRun.Interpreter)
+	}
+}
+
+func scriptRunSh(step *osconfigpb.SoftwareRecipe_Step_ScriptRun, runEnvs []string, stepDir string) error {
 	scriptPath := filepath.Join(stepDir, "script")
 
-	f, err := os.Create(scriptPath)
+	err := writeScript(scriptPath, step.ScriptRun.Script)
```
adjackura: if err :=...; err !=nil {

adjackura: Here and below

policies/recipes/steps.go
TrevorBoone: Done.

policies/recipes/steps.go
TrevorBoone: Done.

PR #45, opened 2019-07-31
Add apt key management and rpm reboot required signal

adjackura:
Manage APT keys with Go's openpgp library
Check if a reboot is required on rpm based systems based on key packages (packages pulled from RHEL and SUSE existing commands/docs)
Bump version to 1.1.0


ospatch/updates.go
```diff
@@ -0,0 +1,116 @@
...
+	"os"
+	"os/exec"
+	"strconv"
+)
+
+const (
+	rpmQuery = "/usr/bin/rpmquery"
+)
+
+func exists(path string) (bool, error) {
```
ryanwe: Should we have a central place for `os` wrappers like this? I think SWR has some too?

ospatch/updates.go
```diff
@@ -0,0 +1,116 @@
...
+		if os.IsNotExist(err) {
+			return false, nil
+		}
+		return false, err
+	}
+
+	return true, nil
+}
+
+func getBtime(stat string) (int, error) {
```
ryanwe: Can you document what "btime" is? It would also be useful to document the whole approach, maybe with `rpmReboot` since that's the public function. For example, "On systems using rpm, the best way to check if a reboot is required is to look at...". We'll also want that wording in our API so users know when a reboot would be triggered.

inventory/packages/rpm.go
```diff
@@ -25,7 +25,7 @@ import (
 var (
 	rpmquery string
 
-	rpmqueryArgs = []string{"-a", "--queryformat", `%{NAME} %{ARCH} %{VERSION}-%{RELEASE}\n`}
+	rpmqueryArgs = []string{"-a", "--queryformat", "%{NAME} %{ARCH} %{VERSION}-%{RELEASE}\n"}
```
ryanwe: What's the difference between `rpmquery` and `rpmQuery`? Should they be the same variable or should one be renamed?

ospatch/updates_test.go
```diff
@@ -0,0 +1,86 @@
+//  Copyright 2018 Google Inc. All Rights Reserved.
```
ryanwe: 2019

ospatch/updates.go
adjackura: Maybe in the future if we have a lot of duplicate code, I prefer to keep simple functions like this close to the code that uses them.

inventory/packages/rpm.go
adjackura: command is named rpmquery so I'll stick with that

PR #46, opened 2019-08-01
Fix a few inventory bugs, add EL8 build

adjackura:
Fixed bugs:
Running inventory right after EL boot could cause a panic because of a yum output line matching enough checks and then an unchecked index
APT updates can have multiple repos they exist in, fixed parsing to allow for this
Tests coming for inventory functions in follow up PR

### DIFF
--- a/inventory/packages/apt_deb.go
+++ b/inventory/packages/apt_deb.go
@@ -59,7 +59,7 @@ func InstallAptPackages(pkgs []string) error {
 	for _, s := range strings.Split(string(out), "\n") {
 		msg += fmt.Sprintf(" %s\n", s)
 	}
-	DebugLogger.Printf("apt install output:\n%s\n", msg)
+	DebugLogger.Printf("apt install output:\n%s", msg)
 	return nil
 }
 
@@ -78,7 +78,7 @@ func RemoveAptPackages(pkgs []string) error {
 	for _, s := range strings.Split(string(out), "\n") {
 		msg += fmt.Sprintf(" %s\n", s)
 	}
-	DebugLogger.Printf("apt remove output:\n%s\n", msg)
+	DebugLogger.Printf("apt remove output:\n%s", msg)
 	return nil
 }
 
@@ -161,7 +161,7 @@ func InstalledDebPackages() ([]PkgInfo, error) {
 	for _, ln := range lines {
 		pkg := strings.Fields(ln)
 		if len(pkg) != 3 {
-			DebugLogger.Printf("%q does not represent a deb\n", ln)
+			DebugLogger.Printf("%q does not represent a deb", ln)
 			continue
 		}
 
--- a/inventory/packages/apt_deb.go
+++ b/inventory/packages/apt_deb.go
@@ -82,17 +82,7 @@ func RemoveAptPackages(pkgs []string) error {
 	return nil
 }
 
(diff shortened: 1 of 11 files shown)


inventory/packages/apt_deb.go
```diff
@@ -112,17 +122,17 @@ func AptUpdates() ([]PkgInfo, error) {
...
 			continue
 		}
-		if len(pkg) == 5 {
+		if strings.HasPrefix(pkg[2], "(") {
 			// We don't want to record new installs.
 			// Inst firmware-linux-free (3.4 Debian:9.9/stable [all])
 			continue
 		}
-		if len(pkg) != 6 {
+		if strings.HasPrefix(pkg[2], "[") && strings.HasPrefix(pkg[3], "(") && strings.HasSuffix(pkg[len(pkg)-1], "]") {
```
ryanwe: When doing this type of check, it may be helpful to include an example of what this is looking for in the comments. That way if there are false positives detected in the future, cases that are similar but need to be handled differently, it'll be clear what this check is looking for. (Like what you have for the last check. But maybe add "eg." to make it clearer.)

inventory/packages/yum.go
```diff
@@ -132,16 +99,19 @@ func YumUpdates() ([]PkgInfo, error) {
...
 		if len(pkg) == 2 && pkg[0] == "Obsoleting" && pkg[1] == "Packages" {
 			break
 		}
 		if len(pkg) != 3 {
-			DebugLogger.Printf("%s does not represent a yum update\n", ln)
 			continue
 		}
 		name := strings.Split(pkg[0], ".")
+		if len(name) != 2 {
+			DebugLogger.Printf("%s does not represent a yum update.", ln)
```
ryanwe: nit: consider wrapping `%s` with single quotes to make it easier to parse the log. (below too)

inventory/packages/apt_deb.go
adjackura: Nice catch, this actually doesn't do what I want (I had caught it in my tests but those aren't in this PR)

inventory/packages/apt_deb.go
```diff
@@ -104,58 +104,72 @@ func AptUpdates() ([]PkgInfo, error) {
...
+		if strings.HasPrefix(pkg[2], "(") {
 			// We don't want to record new installs.
 			// Inst firmware-linux-free (3.4 Debian:9.9/stable [all])
 			continue
 		}
-		if len(pkg) != 6 {
-			DebugLogger.Printf("%q does not represent an apt update\n", ln)
+		// Make sure this line matches expectations:
+		// Inst google-cloud-sdk [245.0.0-0] (246.0.0-0 cloud-sdk-stretch:cloud-sdk-stretch [all])
+		if !strings.HasPrefix(pkg[2], "[") || !strings.HasPrefix(pkg[3], "(") || !strings.HasSuffix(pkg[len(pkg)-1], ")") {
```
ryanwe: This seems fine but you may want to consider using a regex (or regexes) instead for clarity.

adjackura: The point here is to directly reflect the actions performed below, a regex would be less readable and prone to errors as it would need to allow for lots of variations. We really don't care about any fields other than those listed here (and the first one is a bit redundant with the previous check, so we could drop one of those)

PR #54, opened 2019-08-16
adding support for archives

TrevorBoone:


policies/recipes/steps.go
```diff
@@ -88,7 +95,290 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+		}
+		err = os.Chtimes(filen, time.Now(), f.Modified)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func decompress(reader io.ReadCloser, compression osconfigpb.SoftwareRecipe_Step_ExtractArchive_ArchiveType) (io.ReadCloser, error) {
```
TrevorBoone: Compression ended up being really weird with reading the file multiple times. I didn't want to read the file into memory so I settled for decompressing it every pass through.

policies/recipes/steps.go
```diff
@@ -88,7 +95,290 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+	switch compression {
+	case osconfigpb.SoftwareRecipe_Step_ExtractArchive_TAR:
+		return reader, nil
+	case osconfigpb.SoftwareRecipe_Step_ExtractArchive_TAR_GZIP:
+		return gzip.NewReader(reader)
+	case osconfigpb.SoftwareRecipe_Step_ExtractArchive_TAR_BZIP:
+		return ioutil.NopCloser(bzip2.NewReader(reader)), nil
+	case osconfigpb.SoftwareRecipe_Step_ExtractArchive_TAR_LZMA:
+		return nil, fmt.Errorf("lzma currently unsupported")
```
TrevorBoone: Also there don't seem to be standard go libraries for lzma and xz. No third party libraries jumped out at me as particularly stable but I also didn't spend a lot of time digging on it.

policies/recipes/steps.go
TrevorBoone: from a conversation with hopkiw I will decompress the whole file into a temporary directory first then just read from that already compressed file.

policies/recipes/steps.go
TrevorBoone: I added https://github.com/ulikunitz/xz as that seems like the most stable but I haven't dug into it too deeply

policies/recipes/steps.go
```diff
@@ -90,8 +100,335 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+		if err != nil {
+			return err
+		}
+		defer compressed.Close()
+		reader, err := gzip.NewReader(compressed)
+		if err != nil {
+			return err
+		}
+		defer reader.Close()
+		decompressed, err := ioutil.TempFile(stepDir, "archive-*.tar")
```
hopkiw: why this name? i don't think go is going to expand a shell glob. not sure what it would expand to if it did

policies/recipes/steps.go
```diff
@@ -90,8 +100,335 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+			return err
+		}
+		return extractTar(decompressed.Name(), step.ArchiveExtraction.Destination)
+	default:
+		return fmt.Errorf("Unrecognized archive type %q", step.ArchiveExtraction.GetType())
+	}
+}
+
+func zipIsDir(name string) bool {
+	return strings.HasSuffix(name, string(os.PathSeparator))
```
hopkiw: is this right? docs for zip.FileHeader.Name say 
```
        // Name is the name of the file.
        //
        // It must be a relative path, not start with a drive letter (such as "C:"),
        // and must use forward slashes instead of back slashes. A trailing slash
        // indicates that this file is a directory and should have no data.
```

which would seem to indicate this would always be a forward slash, not a platform dependent path separator.

policies/recipes/steps.go
```diff
@@ -90,8 +100,335 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+		if err != nil {
+			return err
+		}
+		fmt.Printf("os.Create %s\n", filen)
+		reader, err := f.Open()
+		if err != nil {
+			return err
+		}
+
+		dst, err := os.OpenFile(filen, os.O_RDWR|os.O_CREATE, 0755)
```
hopkiw: why always 0755? we don't want to respect the file modes from the zip header?

policies/recipes/steps.go
```diff
@@ -90,8 +100,335 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+			return err
+		}
+		if header.Typeflag == tar.TypeDir && stat.IsDir() {
+			// it's ok if directories already exist
+			continue
+		}
+		return fmt.Errorf("file exists: %s", filen)
+	}
+
+	file.Seek(0, 0)
```
hopkiw: btw this was just the testing mechanism i was using, i haven't determined if there's a better way to rewind a tar

policies/recipes/steps.go
```diff
@@ -90,8 +100,335 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+				_, err = io.Copy(dst, tr)
+			}
+		case tar.TypeLink:
+			err = os.Link(header.Linkname, filen)
+			continue
+		case tar.TypeSymlink:
+			err = os.Symlink(header.Linkname, filen)
+			continue
+		case tar.TypeChar:
+			err = unix.Mknod(filen, uint32(unix.S_IFCHR), int(unix.Mkdev(uint32(header.Devmajor), uint32(header.Devminor))))
```
hopkiw: this too was just where my working code was at the time-it seems particularly difficult to reimplement `mknod` in go, we should do our due diligence and look for others who have solved this problem. i'm not very confident in this being the 'right' way to do it

policies/recipes/steps.go
TrevorBoone: It's a format particular to ioutil.Tempfile. It replaces the * with a random number, making sure that the number doesn't collide with existing files. This should be the only file in the directory so I'll switch to the os.create method.

policies/recipes/steps.go
TrevorBoone: I'm under the impression that permissions are not a default part of the zip header. If it's an important feature I could look into how most zip utilities store them.

policies/recipes/steps.go
TrevorBoone: Okay, I didn't find a way around reading the file multiple times and the tar reader doesn't seem to provide a way to go backward at all so I'll keep this implementation for now.

policies/recipes/steps.go
TrevorBoone: mknod seems to map fairly directly to a syscall so calling that (either through the unix or syscall packages) seems to be correct.

policies/recipes/steps.go
TrevorBoone: You're correct, I changed it.

policies/recipes/steps.go
hopkiw: i'm basing this entirely on just reading the go docs for the zip library. if it's populated i would think we'd want to respect it, but idk if it's usually populated.

policies/recipes/steps.go
hopkiw: it's more about the way the arguments are constructed. i recall having a hard time getting that to work and not finding good go docs on it.

policies/recipes/steps.go
hopkiw: I had no idea about that tempfile feature, thx for explaining

policies/recipes/steps.go
TrevorBoone: After looking at the docs in more detail you are correct, I was focused on the struct fields I didn't look too much at the methods apparently. I'll add the support for it in.

policies/recipes/steps.go
TrevorBoone: Yeah, I'm in the same boat. As far as I can tell this works but I don't have enough expertise in unix block/char devices to know. Is there anyone else in particular who would know this area?

policies/recipes/steps.go
```diff
@@ -90,8 +99,344 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+	if !ok {
+		return fmt.Errorf("%q not found in artifact map", step.ArchiveExtraction.GetArtifactId())
+	}
+	switch step.ArchiveExtraction.GetType() {
+	case osconfigpb.SoftwareRecipe_Step_ExtractArchive_ZIP:
+		return extractZip(filename, step.ArchiveExtraction.Destination)
+	case osconfigpb.SoftwareRecipe_Step_ExtractArchive_TAR:
+		return extractTar(filename, step.ArchiveExtraction.Destination)
+	case osconfigpb.SoftwareRecipe_Step_ExtractArchive_TAR_GZIP:
+		compressed, err := os.Open(filename)
```
adjackura: Don't decompress to a file, just pass the reader along to tar. Also all these opens should happen before the switch, then extractZip and extractTar should just take a io.Reader. something like:

file, err := os.Open(filename)
defer file.Close()
step.ArchiveExtraction.GetType() {
case osconfigpb.SoftwareRecipe_Step_ExtractArchive_ZIP:
return extractZip(file, step.ArchiveExtraction.Destination)
case osconfigpb.SoftwareRecipe_Step_ExtractArchive_TAR:
return extractTar(file, step.ArchiveExtraction.Destination)
case osconfigpb.SoftwareRecipe_Step_ExtractArchive_TAR_GZIP:
gz, err := gzip.NewReader(file)
defer reader.Close()
return extractTar(gz, step.ArchiveExtraction.Destination)

policies/recipes/steps.go
```diff
@@ -90,8 +99,344 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+			continue
+		case tar.TypeReg, tar.TypeRegA:
+			fmt.Printf("os.Create %s (owner %s/%d group %s/%d)\n", filen, header.Uname, header.Uid, header.Gname, header.Gid)
+			dst, err := os.Create(filen)
+			if err == nil {
+				_, err = io.Copy(dst, tr)
+			}
+		case tar.TypeLink:
+			err = os.Link(header.Linkname, filen)
+			continue
```
adjackura: Error check?

policies/recipes/steps.go
```diff
@@ -90,8 +99,344 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+		case tar.TypeFifo:
+			err = mkFifo(filen, uint32(header.Mode))
+		default:
+			fmt.Printf("unknown type for %s\n", filen)
+			continue
+		}
+		if err != nil {
+			return err
+		}
+		err = os.Chmod(filen, os.FileMode(header.Mode))
```
adjackura: Why chmod here instead of in os.OpenFile?

policies/recipes/steps.go
```diff
@@ -90,8 +99,344 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+		if err != nil {
+			return err
+		}
+		err = os.Chown(filen, header.Uid, header.Gid)
```
adjackura: Add check for windows here, you don't want to run os.Chown on windows.
Also do this using the same file handle

policies/recipes/steps.go
```diff
@@ -90,8 +99,344 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+		}
+		if err != nil {
+			return err
+		}
+		switch header.Typeflag {
+		case tar.TypeDir:
+			continue
+		case tar.TypeReg, tar.TypeRegA:
+			fmt.Printf("os.Create %s (owner %s/%d group %s/%d)\n", filen, header.Uname, header.Uid, header.Gname, header.Gid)
+			dst, err := os.Create(filen)
```
adjackura: You need to close

policies/recipes/steps.go
```diff
@@ -90,8 +99,344 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+			return err
+		}
+		filen, err := common.NormPath(filepath.Join(dst, header.Name))
+		if err != nil {
+			return err
+		}
+		filedir := filepath.Dir(filen)
+		_, err = os.Stat(filedir)
+		if err != nil {
+			err = os.MkdirAll(filedir, 0755)
```
adjackura: Skip the stat, MkdirAll does that for you just:
if err:= os.MkdirAll(filepath.Dir(filen), 0755); err != nil {
return err
}

policies/recipes/steps.go
```diff
@@ -90,8 +99,344 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+			continue
+		}
+		return fmt.Errorf("file exists: %s", filen)
+	}
+
+	file.Seek(0, 0)
+	tr = tar.NewReader(file)
+
+	// Create dirs
+	for {
```
adjackura: You shouldn't need to loop through twice, just create directories and files in the same loop.

policies/recipes/steps.go
```diff
@@ -90,8 +99,344 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+		if err != nil {
+			return err
+		}
+		if zipIsDir(f.Name) {
+			continue
+		}
+		filedir := filepath.Dir(filen)
+		_, err = os.Stat(filedir)
+		if os.IsNotExist(err) {
+			err = os.MkdirAll(filedir, 0755)
```
adjackura: Same as below

policies/recipes/steps.go
```diff
@@ -90,8 +99,344 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+		}
+		if zipIsDir(f.Name) && stat.IsDir() {
+			// it's ok if directories already exist
+			continue
+		}
+		return fmt.Errorf("file exists: %s", filen)
+	}
+
+	// Create dirs
+	for _, f := range zr.File {
```
adjackura: Same as below, just do this in one loop

policies/recipes/steps.go
```diff
@@ -90,8 +99,344 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+			return err
+		}
+		return extractTar(decompressed.Name(), step.ArchiveExtraction.Destination)
+	default:
+		return fmt.Errorf("Unrecognized archive type %q", step.ArchiveExtraction.GetType())
+	}
+}
+
+func zipIsDir(name string) bool {
+	return strings.HasSuffix(name, "/")
```
adjackura: You may need to also check os.PathSeparator

TrevorBoone: The zip format seems to specify forward slashes specifically, are files not to spec common? From looking into it historically they seem to be reasonably common. I quickly added code in to deal with different slash directions but I'm curious if that's necessary.

policies/recipes/steps.go
TrevorBoone: Done

policies/recipes/steps.go
TrevorBoone: Done.

policies/recipes/steps.go
TrevorBoone: done.

policies/recipes/steps.go
TrevorBoone: I could have the openfile too but the chmod here also works on block and char devices. I'll move it so it's handled in each file type.

policies/recipes/steps.go
TrevorBoone: So the issue here is that we go through the file twice so it seemed like we either need to decompress it temporarily or decompress it twice (I'm not sure which of those is practically better).

Also we could switch it to only one loop and not check for conflicts early and the we would need to only use the reader once if that would make sense but I'm unsure whether this is better or not.

policies/recipes/steps.go
TrevorBoone: Done

policies/recipes/steps.go
adjackura: Test it on Windows to be sure, if it's still '/' you can leave it

policies/recipes/steps.go
adjackura: I'm not sure I understand, a file allows for seeking and I would argue CPU cycles are cheaper (and faster) than disk. Also we aren't extracting twice, the seek is faster than the full extract, also efficiency is already really poor as we run an additional stat for every file to check for conflicts.

policies/recipes/steps.go
hopkiw: the zip spec strictly indicates forward slash, i don't think it's a platform variance so much as checking whether or not all zip producing tools follow the spec. if these zip files are likely to occur then yeah we should have supporting logic

policies/recipes/steps.go
hopkiw: I don't understand how we would seek a gzip reader, in the tar.gz case

policies/recipes/steps.go
TrevorBoone: Essentially what I did was convert '/' to os.PathSeparator then for directories just checked if the path ends with os.PathSeparator. This means on Windows either '/' or '\' will work as path separators but on Linux only '/' will work. This means that zips that use '\' will create directories on windows but on Linux will extract according to the spec.

policies/recipes/steps.go
TrevorBoone: I switched it to not extracting to a file. I had to reinitialize the gzip and tar readers after seeking the files

policies/recipes/steps.go
```diff
@@ -15,17 +15,27 @@
...
 	"path/filepath"
 	"runtime"
 	"strconv"
+	"strings"
+	"time"
 
 	"github.com/GoogleCloudPlatform/osconfig/common"
 
 	osconfigpb "github.com/GoogleCloudPlatform/osconfig/_internal/gapi-cloud-osconfig-go/google.golang.org/genproto/googleapis/cloud/osconfig/v1alpha2"
+
```
adjackura: you have some extra lines between imports, have all std imports together, then one line then all thrid_party imports

policies/recipes/steps.go
hopkiw: i think your comment had some misformatting. but why would we want this differentiated outcome? why not use the simpler 'any trailing slash means a dir' logic? we're adding complex edge cases, and i don't see the motivation

policies/recipes/steps.go
TrevorBoone: Done.

policies/recipes/steps.go
TrevorBoone: It would take more than just checking for dirs like that to actually support '\' as a path separator on Linux. Essentially I would prefer to be consistent about whether '\' was treated as a path separator on each distro.

For Windows some historical tools have created zips with '\' as a separator (and my impression is that modern tools will often handle unzipping them correctly) and also go's os library itself will treat '\' as a path separator so not having '\' be a path separator at all would be fairly difficult.

Linux doesn't have that historical baggage (and as far as I'm aware the current Linux unzip utility doesn't treat '\' as a path separator) and treating '\' as a separator would break handling of zips where '\' is used in a file name and isn't a separator.

I don't have great information on whether finding zips not to spec or filenames with '\' in them is more likely but I erred on the side of sticking to the spec and working the same as the Linux unzip. If this explanation isn't clear we can talk about it more in chat.

policies/recipes/steps.go
```diff
@@ -90,10 +97,274 @@ func parsePermissions(s string) (os.FileMode, error) {
...
 func StepArchiveExtraction(step *osconfigpb.SoftwareRecipe_Step_ArchiveExtraction, artifacts map[string]string) error {
-	fmt.Println("StepArchiveExtraction")
+	filename, ok := artifacts[step.ArchiveExtraction.GetArtifactId()]
+	if !ok {
+		return fmt.Errorf("%q not found in artifact map", step.ArchiveExtraction.GetArtifactId())
+	}
+	switch step.ArchiveExtraction.GetType() {
+	case osconfigpb.SoftwareRecipe_Step_ExtractArchive_ZIP:
+		return extractZip(filename, step.ArchiveExtraction.Destination)
+	case osconfigpb.SoftwareRecipe_Step_ExtractArchive_TAR_GZIP:
```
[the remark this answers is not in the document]
TrevorBoone: I'll switch to this, I had forgotten that you can combine cases with commas

policies/recipes/steps.go
```diff
@@ -90,10 +97,274 @@ func parsePermissions(s string) (os.FileMode, error) {
...
+		if err != nil {
+			return err
+		}
+		filedir := filepath.Dir(filen)
+		err = os.MkdirAll(filedir, 0755)
+		if err != nil {
+			return err
+		}
+		switch header.Typeflag {
+		case tar.TypeDir:
```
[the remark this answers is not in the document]
TrevorBoone: The ordering of directories isn't set either. My current plan is to just create directories as needed (I set the permissions on them to 700 initially to prevent other people from reading them) and then change the permissions and owner of them to be correct when I get to the correct record in the tar.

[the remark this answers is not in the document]
hopkiw: ok, let's see if this works out for us

policies/recipes/steps.go
hopkiw: you seem to be commenting `" as a path separator`

PR #55, opened 2019-08-21
Add ZypperPatches to inventory

adjackura:
Bump version,
Fix systemd service file to work with older versions of systemd

### DIFF
--- a/inventory/packages/zypper.go
+++ b/inventory/packages/zypper.go
@@ -29,8 +29,8 @@ var (
 
 	zypperInstallArgs     = []string{"install", "--no-confirm"}
 	zypperRemoveArgs      = []string{"remove", "--no-confirm"}
-	zypperListArgs        = []string{"packages", "--installed-only"}
 	zypperListUpdatesArgs = []string{"-q", "list-updates"}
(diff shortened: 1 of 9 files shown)


inventory/packages/zypper.go
```diff
@@ -96,3 +97,64 @@ func ZypperUpdates() ([]PkgInfo, error) {
...
+		}
+
+		name := string(bytes.TrimSpace(patch[1]))
+		cat := string(bytes.TrimSpace(patch[2]))
+		sev := string(bytes.TrimSpace(patch[3]))
+		status := string(bytes.TrimSpace(patch[5]))
+		sum := string(bytes.TrimSpace(patch[6]))
+		switch status {
+		case "needed":
+			avail = append(avail, ZypperPatch{Name: name, Category: cat, Severity: sev, Summary: sum})
```
ryanwe: Why is "not needed" not an available patch?

adjackura: Not sure it would make sense to display these, they will not be applied, they are available in the sense that they exist but are not a candidate for install. We could them if we added the status field, but I think that would just be confusing.

ryanwe: SGTM. We can always extend and/or categorize these in the future.

PR #57, opened 2019-09-03
Improve code coverage

iamsubratp:
- Add unit test for osinfo and inventory package
-  Small refactors to make code testable

-----------------------------------------------
coverage before
-----------------------------------------------
ok      github.com/GoogleCloudPlatform/osconfig/config  0.075s  coverage: 70.0% of statements
ok      github.com/GoogleCloudPlatform/osconfig/inventory       0.115s  coverage: 30.8% of statements
ok      github.com/GoogleCloudPlatform/osconfig/inventory/osinfo        0.122s  coverage: 83.7% of statements
ok      github.com/GoogleCloudPlatform/osconfig/inventory/packages      0.078s  coverage: 59.7% of statements
ok      github.com/GoogleCloudPlatform/osconfig/ospatch 0.109s  coverage: 19.6% of statements

-----------------------------------------------
coverage after
-----------------------------------------------
ok      github.com/GoogleCloudPlatform/osconfig/config  0.076s  coverage: 70.0% of statements
ok      github.com/GoogleCloudPlatform/osconfig/inventory       0.152s  coverage: 84.6% of statements
ok      github.com/GoogleCloudPlatform/osconfig/inventory/osinfo        0.085s  coverage: 85.4% of statements
ok      github.com/GoogleCloudPlatform/osconfig/inventory/packages      0.096s  coverage: 60.1% of statements
ok      github.com/GoogleCloudPlatform/osconfig/ospatch 0.073s  coverage: 20.0% of statements


inventory/inventory_test.go
```diff
@@ -137,3 +146,705 @@ func TestWrite(t *testing.T) {
...
+
+	config.SetVersion("1")
+
+	ii := Get()
+
+	assertion.Equal("test-hostname", ii.Hostname, "hostname does not match")
+	assertion.Equal("Debian buster", ii.LongName, "long name does not match")
+	assertion.Equal("test-kernel", ii.KernelVersion, "kernel version does not match")
+	assertion.Equal("foo", ii.InstalledPackages.Rpm[0].Name, "package name does not match")
+	assertion.Equal("1.2.3-4", ii.InstalledPackages.Rpm[0].Version, "package version does not match")
```
ryanwe: Should there be some tests that do match?

inventory/osinfo/osinfo_linux.go
```diff
@@ -29,12 +28,11 @@ import (
 
 var (
 	entRelVerRgx = regexp.MustCompile(`\d+(\.\d+)?(\.\d+)?`)
-	getUname     = func() ([]byte, error) {
+
+	// Getuname returns the kernel details
```
ryanwe: minor: `GetUname`

inventory/inventory_test.go
iamsubratp: the tests are for matching. the messages are error messages if the assertion fails.

inventory/osinfo/osinfo_linux.go
iamsubratp: done!

common/common.go
```diff
@@ -141,10 +143,30 @@ func newHTTPClient() *http.Client {
...
 	return true
 }
+
+// OsHostname is a wrapper to get os hostname
+var OsHostname = func() (name string, err error) {
+	return os.Hostname()
+}
+
+// ReadFile is a wrapper to read file
+var ReadFile = func(file string) ([]byte, error) {
```
adjackura: I'd prefer to not stub out things we don't need to (readfile, exists maybe), also these shouldn't be common if no other code uses them, just keep it private next to the code that uses it.

inventory/osinfo/osinfo.go
```diff
@@ -28,7 +28,7 @@ type DistributionInfo struct {
 }
 
 // Architecture attempts to standardize architecture naming.
-func Architecture(arch string) string {
+var Architecture = func(arch string) string {
```
adjackura: Why us this a var? We should use the actual implementation in tests as it does nothing fancy

inventory/osinfo/osinfo_linux.go
```diff
@@ -29,12 +28,11 @@ import (
 
 var (
 	entRelVerRgx = regexp.MustCompile(`\d+(\.\d+)?(\.\d+)?`)
-	getUname     = func() ([]byte, error) {
+
+	// GetUname returns the kernel details
+	GetUname = func() ([]byte, error) {
```
adjackura: Why is this public?

common/common.go
iamsubratp: i introduced stubs for these small utility functions because of a series of reasons.
- to test presence of package manager binaries, i could have either created fake binaries on the native file system or a fake binary on a test file system or i could stub the exists method itself. i chose to stub because creating fake binaries would not work system where the actual binary is already present(though there are ways to get around it), for simplicity and less maintenance i stubbed the "exists" method. with this approach i had this concern that we are not testing existence of package manager binary; but i ignored it since we can safely assume that apt would be present on an apt based system and if it not then the agent must crash anyways. it is on the customer to fix it(correct me if i am wrong here).

- for readfile; i do feel i could have just used the native file system to create the files, but since i was already using the fake filesystem for package binaries; i did not want to mix things up.




> these shouldn't be common if no other code uses them, just keep it private next to the code that uses it.

most of it like anything related to fetcher is already moved to a different package. i would still like to keep these utility functions public since these are accessed by different package. will change this to "utility"?

inventory/osinfo/osinfo_linux.go
iamsubratp: i am stubbing the getuname function in inventory package

inventory/osinfo/osinfo.go
iamsubratp: yeah... i did this initially thinking i might need to get mock output from architecture function; but it seems i dont need to mock the output of this method.

inventory/osinfo/osinfo_linux.go
iamsubratp: made this private

PR #59, opened 2019-09-05
Add fallback logic for installing/removing apt packages when applying…

manjunathganga:
When there are invalid packages that are configured to be installed/updated/deleted, the apt package manager fails to install the valid packages as well.

For example: `apt-get install dos2unix xxxxxx` fails to install the valid `dos2unix` package because of the presence of the invalid package `xxxxxx`.

As a result, this PR adds fallback logic to install the packages individually when the combined install fails. We add this as a fallback and not as default logic because the combined install is relatively faster than individual installs.

The same rationale applies when removing the packages as well.

Note: This PR only applies to the `apt` package manager. Once we are in agreement with this approach, we can apply this to the other package managers as well

### DIFF
--- a/inventory/packages/apt_deb.go
+++ b/inventory/packages/apt_deb.go
@@ -16,7 +16,6 @@ package packages
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -62,20 +61,6 @@ func InstallAptPackages(pkgs []string) error {
 	return err
 }
 
-// InstallAptPackagesIndividually installs apt packages individually.
-func InstallAptPackagesIndividually(pkgs []string) error {
-	var errs []string
-	for _, pkg := range pkgs {
-		if err := InstallAptPackages([]string{pkg}); err != nil {
-			errs = append(errs, fmt.Sprintf("Error installing package: %v. Error details: %v", pkg, err))
-		}
-	}
-	if len(errs) > 0 {
-		return errors.New(strings.Join(errs, ",\n"))
-	}
-	return nil
-}
-
 // RemoveAptPackages removes apt packages.
 func RemoveAptPackages(pkgs []string) error {
 	args := append(aptGetRemoveArgs, pkgs...)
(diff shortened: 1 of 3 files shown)


inventory/packages/apt_deb.go
```diff
@@ -61,6 +62,20 @@ func InstallAptPackages(pkgs []string) error {
 	return err
 }
 
+// InstallAptPackagesIndividually installs apt packages individually.
+func InstallAptPackagesIndividually(pkgs []string) error {
```
iamsubratp: is it possible to not introduce this new method here in packge manager and instead just introduce this fallback in "policies" where we call "InstallAptPackages". 
I am asking this because this is supposed to be just go implementation of package managers(or atleast that is how i see this package)

policies/apt.go
```diff
@@ -145,7 +145,13 @@ func aptChanges(aptInstalled, aptRemoved, aptUpdated []*osconfigpb.Package) erro
 		logger.Infof("Installing packages %s", changes.packagesToInstall)
 		if err := packages.InstallAptPackages(changes.packagesToInstall); err != nil {
 			logger.Errorf("Error installing apt packages: %v", err)
-			errs = append(errs, fmt.Sprintf("error installing apt packages: %v", err))
+
```
iamsubratp: i would have loved to the some unit test testing this fallback logic. I am actually working on adding more coverage to this package.

manjunathganga: Thanks. Now the fallback logic is moved here so that we can test this part.

inventory/packages/apt_deb.go
manjunathganga: Done

inventory/packages/apt_deb_test.go
```diff
@@ -35,15 +35,15 @@ func TestInstallAptPackagesReturnsError(t *testing.T) {
 	}
 }
 
-func TestRemoveApt(t *testing.T) {
-	run = getMockRun([]byte("TestRemoveApt"), nil)
+func TestRemoveAptPackages(t *testing.T) {
```
iamsubratp: thanks for renaming this. makes more sense now

PR #67, opened 2019-09-11
Fix recipeDB init

hopkiw:
Stacked PR, view latest commit only

Corrects 'assignment to nil map' RTE
Resulting on-disk JSON is e.g. `[{Name: "recipe_name", Version: "0", ...}, { ... } ]`

This isn't strictly valid JSON as top level isn't an object, but the marshaling functions don't mind.

### DIFF
--- a/policies/recipes/recipedb.go
+++ b/policies/recipes/recipedb.go
@@ -16,7 +16,6 @@ package recipes
 
 import (
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


policies/recipes/recipedb.go
```diff
@@ -32,16 +32,37 @@ var (
...
+	Version     recipeVersion
+	InstallTime int64
+	Success     bool
+}
+
+// RecipeDB represents local state of installed recipes.
+type RecipeDB map[string]Recipe
+
+// NewRecipeDB instantiates a recipeDB.
+func NewRecipeDB() (RecipeDB, error) {
```
adjackura: Is there a reason for this being public? Or really anything in this package other than the entry point, does something in guestpolicies actually need to call other functions/structs?

hopkiw: Nothing needs to be public besides InstallRecipe and struct /struct attributes for marshaling.

PR #68, opened 2019-09-12
SWR remove working directory after run

hopkiw:
Stacked PR - only first commit

Set filecopy default permissions
Remove working directory after recipe install
Remove fmt.Printf calls


adjackura:
If you drop the other commits and we can merge the cleanup function now

PR #74, opened 2019-09-23
Use Uname to get Linux OSInfo

adjackura:


inventory/inventory.go
```diff
@@ -78,34 +77,30 @@ func Get() *InstanceInventory {
...
 	if err != nil {
-		logger.Errorf("packages.GetInstalledPackages() error: %v", err)
+		logger.Errorf("osinfo.Get() error: %v", err)
 	}
 
-	hs.PackageUpdates, err = packages.GetPackageUpdates()
-	if err != nil {
-		logger.Errorf("packages.GetPackageUpdates() error: %v", err)
-	}
+	hs.Hostname = oi.Hostname
```
hopkiw: if you include all the members, why not embed osinfo

inventory/osinfo/osinfo_linux.go
```diff
@@ -104,24 +103,22 @@ func GetDistributionInfo() (*DistributionInfo, error) {
...
 	b, err := ioutil.ReadFile(releaseFile)
 	if err != nil {
-		di = &DistributionInfo{ShortName: Linux}
+		oi = &OSInfo{ShortName: Linux}
 	} else {
-		di = parseReleaseFunc(string(b))
+		oi = parseReleaseFunc(string(b))
 	}
 
-	out, err := getUname()
```
hopkiw: does getUname still need to exist, then?

inventory/osinfo/osinfo_linux.go
```diff
@@ -104,24 +103,22 @@ func GetDistributionInfo() (*DistributionInfo, error) {
...
+	var uts unix.Utsname
+	if err := unix.Uname(&uts); err != nil {
+		return nil, fmt.Errorf("unix.Uname error: %v", err)
 	}
-	di.Kernel = strings.TrimSpace(string(out))
-	// No need to get fancy here, assume the binary architecture
-	// is the same as the system.
-	di.Architecture = Architecture(runtime.GOARCH)
-	return di, nil
+	oi.Hostname = string(uts.Nodename[:])
```
hopkiw: i don't quite understand the purpose of this line; what do you get out of unix.Uname that needs this treatment

inventory/inventory.go
adjackura: Mainly because of the way the write works. We could change that but that would be a bigger PR

inventory/osinfo/osinfo_linux.go
adjackura: This is a byte array, [65]byte. Better than what it is in syscall [65]int8

inventory/osinfo/osinfo_linux.go
adjackura: Nope, removed

PR #77, opened 2019-09-25
small bugfixes

TrevorBoone:
two small bugs I found while working on the end to end tests.

### DIFF
--- a/external/fetcher.go
+++ b/external/fetcher.go
@@ -64,7 +64,7 @@ func DownloadStream(r io.ReadCloser, checksum, localPath string) error {
 	if err != nil {
 		return err
 	}
-	file, err := os.OpenFile(localPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0700)
+	file, err := os.Create(localPath)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


external/fetcher.go
```diff
@@ -64,7 +64,7 @@ func DownloadStream(r io.ReadCloser, checksum, localPath string) error {
 	if err != nil {
 		return err
 	}
-	file, err := os.Create(localPath)
+	file, err := os.OpenFile(localPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0700)
```
TrevorBoone: Create doesn't by default give execute permissions (even to owner)

policies/recipes/artifacts.go
```diff
@@ -70,7 +70,7 @@ func fetchArtifact(ctx context.Context, artifact *osconfigpb.SoftwareRecipe_Arti
 		extension = path.Ext(uri.Path)
 		checksum = remote.Checksum
 		cl := &http.Client{}
-		reader, err := getHTTPArtifact(cl, *uri)
```
TrevorBoone: reader shadowed previous variable

external/fetcher.go
adjackura: We should pass in the permissions or leave it up to the caller to set them after, 0700 does not seem like a normal permission set for a function called simply called DownloadStream

TrevorBoone: so after some thought I added a chmod in the execution step for artifacts, but not local files. I figured changing permissions on local files was not something we wanted to do in this api.

PR #79, opened 2019-09-25
Move ospatch to ospatchog, cleanup WUA library

adjackura:
WUA library updated to be easier to understand
ospatchog will be replaced by new agentendpoint ospatch soon


inventory/packages/wua_windows.go
```diff
@@ -15,15 +15,139 @@ package packages
...
+// IUpdateSession is a an IUpdateSession.
+type IUpdateSession struct {
+	obj *ole.IUnknown
+	ses *ole.IDispatch
+}
+
+func NewUpdateSession() (*IUpdateSession, error) {
+	wuaSession.Lock()
+	if err := ole.CoInitializeEx(0, ole.COINIT_MULTITHREADED); err != nil {
+		return nil, err
```
ryanwe: If `NewUpdateSession` returns an error, will `wuaSession` ever be unlocked? It doesn't look like that will happen.

inventory/packages/wua_windows.go
```diff
@@ -15,15 +15,139 @@ package packages
...
+
+func (c *IUpdateCollection) Add(updt *IUpdate) error {
+	if _, err := c.CallMethod("Add", updt); err != nil {
+		return fmt.Errorf(`updateColl.CallMethod("Add", updt): %v`, err)
+	}
+	return nil
+}
+
+func (c *IUpdateCollection) RemoveAt(i int) error {
+	if c == nil {
```
ryanwe: here and below, is this an expected state? When would this happen and should it have comments for when/why it's needed?

inventory/packages/wua_windows.go
```diff
@@ -15,15 +15,139 @@ package packages
...
+	return count, nil
+}
+
+func (c *IUpdateCollection) Item(i int) (*IUpdate, error) {
+	updtRaw, err := c.GetProperty("Item", i)
+	if err != nil {
+		return nil, err
+	}
+	return &IUpdate{updtRaw.ToIDispatch()}, nil
+}
```
ryanwe: Is there any test coverage for this beyond the e2d tests?

adjackura: Nope, many of these calls can't really be unit tested as they are all interactions with COM, and even the ones that can can only be tested on Windows and we aren't setup for that. This is where an integration test that is not a full e2e test with the API would be really useful.

inventory/packages/wua_windows.go
adjackura: This is not expected and was here for my testing

inventory/packages/wua_windows.go
adjackura: fixed

adjackura: Also FYI this library still has some sharp edges that you need to be aware of to use, I tried to make it simpler to understand but the effort to try to make this really universally safe to use is not really worth it as it is meant for the osconfig agent and we manage our action serialization well in there.

PR #84, opened 2019-10-02
Add basic agentendpoint logic

adjackura:
This does not contain any patch or exec system code yet, just the basic api interactions and basic logic.

### DIFF
--- a/agentendpoint/agentendpoint.go
+++ b/agentendpoint/agentendpoint.go
@@ -31,10 +31,11 @@ import (
 	agentendpointpb "github.com/GoogleCloudPlatform/osconfig/_internal/gapi-cloud-osconfig-go/google.golang.org/genproto/googleapis/cloud/osconfig/agentendpoint/v1alpha1"
 )
 
+const apiRetrySec = 600
+
 var (
 	errServerCancel = errors.New("task canceled by server")
 	taskStateFile   = config.TaskStateFile()
-	patchEnabled    = config.OSPatchEnabled()
 )
 
 // Client is a an agentendpoint client.
@@ -62,7 +63,7 @@ func (c *Client) ReportTaskProgress(ctx context.Context, req *agentendpointpb.Re
 		return nil, err
 	}
 	req.InstanceIdToken = token
-	if err := retryAPICall(2100*time.Second, "ReportTaskProgress", func() error {
+	if err := retryAPICall(apiRetrySec*time.Second, "ReportTaskProgress", func() error {
 		res, err = c.raw.ReportTaskProgress(ctx, req)
 		if err != nil {
 			return err
@@ -84,7 +85,7 @@ func (c *Client) ReportTaskComplete(ctx context.Context, req *agentendpointpb.Re
 		return err
 	}
 	req.InstanceIdToken = token
-	if err := retryAPICall(2100*time.Second, "ReportTaskComplete", func() error {
+	if err := retryAPICall(apiRetrySec*time.Second, "ReportTaskComplete", func() error {
 		res, err := c.raw.ReportTaskComplete(ctx, req)
 		if err != nil {
 			return err
@@ -122,11 +123,6 @@ func (c *Client) runTask(ctx context.Context) {
 
 		switch task.GetTaskType() {
 		case agentendpointpb.TaskType_APPLY_PATCHES:
-			if !patchEnabled {
-				logger.Infof("Recieved TaskType_APPLY_PATCHES but OSPatch is not enabled")
-				continue
-			}
-
 			if err := c.RunApplyPatches(ctx, task.GetApplyPatchesTask()); err != nil {
 				logger.Errorf(err.Error())
 			}
--- a/agentendpoint/agentendpoint.go
+++ b/agentendpoint/agentendpoint.go
@@ -1,0 +1,221 @@
+//  Copyright 2019 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+// Package agentendpoint connects to the osconfig agentendpoint api.
+package agentendpoint
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"time"
+
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+	agentendpoint "github.com/GoogleCloudPlatform/osconfig/_internal/gapi-cloud-osconfig-go/cloud.google.com/go/osconfig/agentendpoint/apiv1alpha1"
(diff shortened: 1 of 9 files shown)


agentendpoint/agentendpoint.go
```diff
@@ -0,0 +1,221 @@
...
+
+		task := startRes.GetTask()
+		if task == nil {
+			return
+		}
+
+		switch task.GetTaskType() {
+		case agentendpointpb.TaskType_APPLY_PATCHES:
+			if !patchEnabled {
+				logger.Infof("Recieved TaskType_APPLY_PATCHES but OSPatch is not enabled")
```
jaiminsh: Should we call ReportTaskComplete with status FAILED because OS Patch is not enabled?

agentendpoint/agentendpoint.go
```diff
@@ -0,0 +1,221 @@
...
+				continue
+			}
+
+			if err := c.RunApplyPatches(ctx, task.GetApplyPatchesTask()); err != nil {
+				logger.Errorf(err.Error())
+			}
+		case agentendpointpb.TaskType_EXEC_STEP_TASK:
+			if err := c.RunExecStep(ctx, task.GetExecStepTask()); err != nil {
```
jaiminsh: If patch is not enabled, we shouldn't be allowing exec step as well, right?

agentendpoint/agentendpoint.go
```diff
@@ -0,0 +1,221 @@
...
+
+// ReportTaskProgress calls the agentendpoint service ReportTaskProgress call injecting the InstanceIdToken.
+func (c *Client) ReportTaskProgress(ctx context.Context, req *agentendpointpb.ReportTaskProgressRequest) (res *agentendpointpb.ReportTaskProgressResponse, err error) {
+	logger.Debugf("ReportTaskProgress request:\n%s", util.PrettyFmt(req))
+	token, err := config.IDToken()
+	if err != nil {
+		return nil, err
+	}
+	req.InstanceIdToken = token
+	if err := retryAPICall(2100*time.Second, "ReportTaskProgress", func() error {
```
jaiminsh: Make 2100 a constant, do we want to retry for 35 mins before failing?

agentendpoint/agentendpoint.go
```diff
@@ -0,0 +1,221 @@
...
+		tasker.Enqueue("PatchRun", func() {
+			st.PatchTask.run(ctx)
+		})
+	}
+
+	return nil
+}
+
+// WaitForTaskNotification waits for and acts on any task notification indefinitely.
+func (c *Client) WaitForTaskNotification(ctx context.Context) error {
```
jaiminsh: WaitForTaskNotification is the entry point to watch for tasks, who is the caller?

agentendpoint/exec_task.go
```diff
@@ -0,0 +1,106 @@
...
+		ErrorMessage: errMsg,
+		Output:       output,
+	}
+	if err := e.client.ReportTaskComplete(ctx, req); err != nil {
+		return fmt.Errorf("error reporting completed state: %v", err)
+	}
+	return nil
+}
+
+func (e *execTask) run(ctx context.Context) error {
```
jaiminsh: What happens when the agent reportTaskProgress and then the instance is restarted. Would the agent not pick up the exec step again?

agentendpoint/exec_task.go
```diff
@@ -0,0 +1,106 @@
...
+		},
+	}
+	res, err := e.client.ReportTaskProgress(ctx, req)
+	if err != nil {
+		return fmt.Errorf("error reporting state %s: %v", agentendpointpb.ExecStepTaskProgress_STARTED, err)
+	}
+
+	if res.GetTaskDirective() == agentendpointpb.TaskDirective_STOP {
+		return e.reportCompletedState(ctx, errServerCancel.Error(), &agentendpointpb.ReportTaskCompleteRequest_ExecStepTaskOutput{
+			// TODO: Maybe there should be a canceled state instead?
```
wj-chen: Seems good to have a CANCELED state. Does reporting COMPLETED here result in an overall patch job CANCEL state?

agentendpoint/patch_task.go
```diff
@@ -0,0 +1,211 @@
...
+
+func (r *patchTask) debugf(format string, v ...interface{}) {
+	logger.Log(logger.LogEntry{Message: fmt.Sprintf(format, v...), Severity: logger.Debug, Labels: r.LogLabels})
+}
+
+func (r *patchTask) infof(format string, v ...interface{}) {
+	logger.Log(logger.LogEntry{Message: fmt.Sprintf(format, v...), Severity: logger.Info, Labels: r.LogLabels})
+}
+
+func (r *patchTask) warningf(format string, v ...interface{}) {
```
wj-chen: Apparently unused

agentendpoint/exec_task.go
```diff
@@ -0,0 +1,106 @@
...
+	sh      = "/bin/sh"
+
+	winPowershell string
+	winCmd        string
+
+	goos = runtime.GOOS
+
+	errLinuxPowerShell = errors.New("interpreter POWERSHELL cannot be used on non-Windows system")
+	errWinNoInt        = fmt.Errorf("interpreter must be specified for a Windows system")
+)
```
wj-chen: nit: add these in the next PR with the actual execution logic?

agentendpoint/exec_task.go
adjackura: Exec doesnt have state we wish to save, we don't allow for recovery from a crash or reboot. We are leaving the logic up to the server though, so if we reportTaskProgress then crash and restart I would expect the server to record it as a failure as we lost our place for some unknown reason.

agentendpoint/agentendpoint.go
adjackura: It will be in osconfig main.go, right now this is unused

agentendpoint/agentendpoint.go
adjackura: Actually this didn't make much sense, I removed the check from here and will add it in to the caller. If patch isn't enabled we don't even want the endpoint setup right now. Ill need to rethink this so just dropped from this PR

agentendpoint/agentendpoint.go
adjackura: See other comment, dropped this from this PR for now.

agentendpoint/exec_task.go
adjackura: The service handles all that, all the agent knows is that it was told to stop before the task was finished. It is most likely because of a timeout or a user cancellation but only the service knows the exact reason. My comment here is mainly to differentiate to the service that the task on this agent as canceled before completion and not actually complete

agentendpoint/exec_task.go
jaiminsh: Before starting the execution and after calling ReportTaskProgress, if the instance restarts - the server expects that the agent will finish the execution, so we should be saving the task to execute. While the agent was running the task, if the instance restarts then yes that should be marked as failure. What do you think?

adjackura: There is no reason the agent should restart at that point, or any reason there will be a gap of any serious amount of time between ReportTaskProgress and the actual execution (we're talking ms here, it reports started, then checks if it should continue, then starts execution). Since we don't know why the agent restarted or where we may be in the execution process and we do not know how to recover so we should just treat it as a failure (the server will, the agent will be in an odd state). We could save state so the agent could come back up and know it is in a failed state and then send the failure report, but that's adding additional logic to the agent that probably doesn't need to be there (unless you tell me that will be simpler than the service doing it).

See this PR for the rest of the exec logic: https://github.com/GoogleCloudPlatform/osconfig/pull/86/

PR #85, opened 2019-10-03
bugfixes for windows

TrevorBoone:
here is a collection of small bugfixes for the agent on windows platforms


policies/recipes/recipedb.go
```diff
@@ -89,8 +89,10 @@ func (db RecipeDB) addRecipe(name, version string, success bool) error {
 	if err != nil {
 		return err
 	}
-	defer f.Close()
-	if _, err = f.Write(dbBytes); err != nil {
+
+	_, err = f.Write(dbBytes)
+	f.Close()
```
TrevorBoone: the file was still open when the os.rename was being called, which was causing errors on windows

policies/recipes/steps.go
```diff
@@ -499,5 +500,14 @@ func executeCommand(cmd string, args []string, workDir string, runEnvs []string,
 			}
 		}
 	}
+	logger.Infof("Combined output for %q command:\n%s", cmd, o)
 	return err
 }
+
+func chown(file string, uid, gid int) error {
```
TrevorBoone: is there some other way to chown on windows? I'm not sure it really makes much sense to change the owner anyway since I think it's just used for the tar command right now.

policies/recipes/steps.go
```diff
@@ -499,5 +500,14 @@ func executeCommand(cmd string, args []string, workDir string, runEnvs []string,
 			}
 		}
 	}
+	logger.Infof("Combined output for %q command:\n%s", cmd, o)
```
TrevorBoone: I added some debugging here because I found it incredibly useful while trying to debug some of these issues, let me know if this shouldn't be here.

policies/recipes/steps.go
```diff
@@ -484,13 +484,14 @@ func writeScript(path, contents string) error {
 func executeCommand(cmd string, args []string, workDir string, runEnvs []string, allowedExitCodes []int32) error {
 	cmdObj := exec.Command(cmd, args...)
 	cmdObj.Dir = workDir
+	cmdObj.Env = append(cmdObj.Env, os.Environ()...)
```
TrevorBoone: for some reason starting powershell was failing when it was run as a service. Adding in os.Environ seemed to fix it (I think that Powershell is expecting certain variables to be setup when it starts)

hopkiw: in general, we want a clean environment without any carried-over environment variables for working space. can we figure out what specifically powershell needs to be set?

policies/recipes/recipedb.go
adjackura: You need to check the return on the close:

if _, err = f.Write(dbBytes); err != nil {
f.Close()
return err
}

if err := f.Close() {
return err
}

policies/recipes/steps.go
adjackura: This would be very complicated, the environment powershell expects is a combination of variables that you should not try to create by hand. The varaibles you need a a combination of explicit system variables (HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment), and implicit variables like ProgramFiles,COMPUTERNAME,HOMEDRIVE,SystemDrive,SystemRoot...

I think using CreateEnvironmentBlock would be the 'correct' way if we wanted a clean environment, but here in our case I think just inheriting the parents makes sense and shouldn't cause us issues:
https://docs.microsoft.com/en-us/windows/win32/api/userenv/nf-userenv-createenvironmentblock?redirectedfrom=MSDN
https://godoc.org/golang.org/x/sys/windows#CreateEnvironmentBlock

some more reading:
https://docs.microsoft.com/en-us/windows/win32/procthread/environment-variables

policies/recipes/recipedb.go
TrevorBoone: done

policies/recipes/recipedb.go
```diff
@@ -89,8 +89,14 @@ func (db RecipeDB) addRecipe(name, version string, success bool) error {
 	if err != nil {
 		return err
 	}
-	defer f.Close()
-	if _, err = f.Write(dbBytes); err != nil {
+
+	_, err = f.Write(dbBytes)
```
adjackura: Avoid shadowing where you can:

if err := ...

policies/recipes/steps.go
TrevorBoone: So the documents that Andrew posted matches up with what I found. @hopkiw how important is not carrying over variables for this? Generally our service will be running as a service on the machine and I expect the environment there should be essentially the same as if we created a new block. Also would it be worth it to just add these environment variables for powershell execution? My worry is that powershell won't be the only thing that requires an initialized environment.

policies/recipes/recipedb.go
TrevorBoone: is it desirable to shadow variables? My instinct is to not shadow variables but I'll do whatever is most idiomatic in go

policies/recipes/steps.go
TrevorBoone: So I did some more digging and changed my mind slightly on this. I added methods to create a 'default' environment, which is created from the current process token on windows and just returns an empty slice on Linux.
I think creating the environment differently per OS should be fine and to me this seems like it gets a fairly clean environment without jumping through too many hoopy.

https://godoc.org/golang.org/x/sys/windows#GetCurrentProcessToken
https://godoc.org/golang.org/x/sys/windows#Token.Environ

policies/recipes/recipedb.go
adjackura: For most things no, but for err re instantiating for the if block scope is desirable as you wont leak it outside of the scope. This is all because error checks are always if err != nil, so keeping it scoped where possible is preferable.

TrevorBoone: That makes a lot of sense. I changed the error handling style to be more consistent

policies/recipes/steps.go
```diff
@@ -484,13 +487,18 @@ func writeScript(path, contents string) error {
 func executeCommand(cmd string, args []string, workDir string, runEnvs []string, allowedExitCodes []int32) error {
 	cmdObj := exec.Command(cmd, args...)
 	cmdObj.Dir = workDir
+	defaultEnv, err := createDefaultEnvironment()
+	if err != nil {
+		return fmt.Errorf("Error creating default environment: %v", err)
```
adjackura: Don't capitalized errors:
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

PR #86, opened 2019-10-08
Add exec task to agent endpoint

adjackura:


agentendpoint/exec_task.go
```diff
@@ -71,9 +148,63 @@ func (e *execTask) run(ctx context.Context) error {
 		})
 	}
 
+	stepConfig := e.Task.GetExecStep().GetLinuxExecStepConfig()
+	if goos == "windows" {
+		stepConfig = e.Task.GetExecStep().GetLinuxExecStepConfig()
```
jaiminsh: This should be GetWindowsExecStepConfig

agentendpoint/exec_task.go
```diff
@@ -16,14 +16,91 @@ package agentendpoint
...
+		return "", errors.New("gcsObject cannot be nil")
+	}
+
+	var reader io.ReadCloser
+	cl, err := storage.NewClient(ctx)
+	if err != nil {
+		return "", fmt.Errorf("error creating gcs client: %v", err)
+	}
+	external.FetchGCSObject(ctx, cl, gcsObject.Object, gcsObject.Bucket, gcsObject.GenerationNumber)
+	defer reader.Close()
```
wj-chen: reader needs to be assigned, see https://github.com/GoogleCloudPlatform/osconfig/commit/10ef0dbda51d24ea5f2d927d9ea65182f7d0851a, it was dropped after the package common -> package external refactor

agentendpoint/exec_task.go
```diff
@@ -16,14 +16,91 @@ package agentendpoint
...
+	logger.Debugf("Fetched GCS object bucket %s object %s generation number %d", gcsObject.GetBucket(), gcsObject.GetObject(), gcsObject.GetGenerationNumber())
```
wj-chen: Was thinking about this when preparing for cohort demo the other day...is it ok for us to debug-log the bucket and object names? Are they PII?

Same with some other logging in this file

agentendpoint/exec_task.go
```diff
@@ -16,14 +16,91 @@ package agentendpoint
...
+func executeCommand(path string, exitCodes []int32, args []string, loggingLables map[string]string) (int32, error) {
+	logger.Debugf("Running command %s with args %s", path, args)
+
+	cmd := exec.Command(path, args...)
+	out, err := run(cmd)
+	var exitCode int32
+	if cmd.ProcessState != nil {
+		exitCode = int32(cmd.ProcessState.ExitCode())
+		logger.Infof("Command exit code: %d, stdout:\n%s", exitCode, out)
+	}
```
wj-chen: I think `cmd.CombinedOutput()` includes stderr from the command too, maybe just update the log message to mention that?

wj-chen: And will cmd.ProcessState but nil somehow? Will err be guaranteed non-nil in that case (so we go into line 95)?

agentendpoint/exec_task.go
```diff
@@ -16,14 +16,91 @@ package agentendpoint
...
+	localPath := filepath.Join(os.TempDir(), path.Base(gcsObject.GetObject()))
+	if err := external.DownloadStream(reader, "", localPath, 0755); err != nil {
+		return "", fmt.Errorf("error downloading GCS object: %s", err)
+	}
+
+	logger.Debugf("Downloaded to local path %s", localPath)
+	return localPath, nil
+}
+
+func executeCommand(path string, exitCodes []int32, args []string, loggingLables map[string]string) (int32, error) {
```
wj-chen: exitCodes are not used yet?

agentendpoint/exec_task.go
```diff
@@ -71,9 +148,63 @@ func (e *execTask) run(ctx context.Context) error {
...
+			exitCode, err = executeCommand(localPath, codes, nil, e.LogLabels)
+		}
+	case agentendpointpb.ExecStepConfig_SHELL:
+		if goos == "windows" {
+			exitCode, err = executeCommand(winCmd, codes, []string{"/c", localPath}, e.LogLabels)
+		} else {
+			exitCode, err = executeCommand(sh, codes, []string{localPath}, e.LogLabels)
+		}
+	case agentendpointpb.ExecStepConfig_POWERSHELL:
+		if goos != "windows" {
```
wj-chen: nit: the other cases check for `if goos == "windows"`

agentendpoint/exec_task.go
adjackura: "ProcessState contains information about an exited process", so I assume we can trust it to be there. 

From the code the only way it wouldn't is if we got an error with the setup of the exec call.

agentendpoint/exec_task.go
adjackura: Logs are all stored in the users project, so logs themselves are PII, so I would say this is fine.

agentendpoint/exec_task.go
adjackura: Dropped them from the agent code as this should be handled in the service

PR #87, opened 2019-10-08
Fork ospatch to work with agentendpoint

adjackura:
This is essentially a straight copy/paste with some bits pulled out. Those will be added back in in a subsequent PR connecting the task execution code to patch

### DIFF
--- a/ospatch/runner.go
+++ b/ospatch/runner.go
@@ -15,10 +15,7 @@
 package ospatch
 
 import (
-	"math"
-	"math/rand"
 	"os/exec"
-	"time"
 
 	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
 )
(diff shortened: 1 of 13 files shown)


ospatch/runner.go
```diff
@@ -0,0 +1,57 @@
...
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		logger.Debugf("error running %q with args %q: %v, stdout: %s", cmd.Path, cmd.Args, err, out)
+		return nil, err
+	}
+	return out, nil
+}
+
+// retry tries to retry f for no more than maxRetryTime.
+func retry2(maxRetryTime time.Duration, desc string, logF func(string, ...interface{}), f func() error) error {
```
ryanwe: Why "retry2" and not "retry"? If there's a name collision, it might be worth combining the two functions.

ospatch/updates.go
```diff
@@ -0,0 +1,110 @@
...
+		if err != nil {
+			continue
+		}
+		if itime > btime {
+			return true
+		}
+	}
+
+	return false
+
```
ryanwe: nit: remove extra <br>

ospatch/runner.go
adjackura: This was supposed to be removed

PR #88, opened 2019-10-21
Add beginnings of includes/excludes for package management

adjackura:
Includes name may/probably will change in the future.
All package management functions for GooGet and APT moved to the 'packages' package from ospatch
Packages to be updated are now evaluated, logged, then installed
Dryrun option does everything up to actually install new packages.
APT logic updated to support both dist/full and normal upgrades
Logging package has been updated so logging initialization changed slightly

### DIFF
--- a/inventory/packages/apt_deb.go
+++ b/inventory/packages/apt_deb.go
@@ -31,11 +31,15 @@ var (
 	dpkgquery string
 	aptGet    string
 
-	dpkgQueryArgs        = []string{"-W", "-f", "${Package} ${Architecture} ${Version}\n"}
-	aptGetInstallArgs    = []string{"install", "-y"}
-	aptGetRemoveArgs     = []string{"remove", "-y"}
-	aptGetUpdateArgs     = []string{"update"}
-	aptGetUpgradableArgs = []string{"full-upgrade", "--just-print", "-V"}
+	dpkgQueryArgs     = []string{"-W", "-f", "${Package} ${Architecture} ${Version}\n"}
+	aptGetInstallArgs = []string{"install", "-y"}
+	aptGetRemoveArgs  = []string{"remove", "-y"}
+	aptGetUpdateArgs  = []string{"update"}
+
+	aptGetUpgradeCmd     = "upgrade"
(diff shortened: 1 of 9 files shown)


inventory/packages/apt_deb.go
```diff
@@ -31,11 +31,15 @@ var (
...
-	aptGetUpgradableArgs = []string{"full-upgrade", "--just-print", "-V"}
+	dpkgQueryArgs     = []string{"-W", "-f", "${Package} ${Architecture} ${Version}\n"}
+	aptGetInstallArgs = []string{"install", "-y"}
+	aptGetRemoveArgs  = []string{"remove", "-y"}
+	aptGetUpdateArgs  = []string{"update"}
+
+	aptGetUpgradeCmd     = "upgrade"
+	aptGetFullUpgradeCmd = "full-upgrade"
+	aptGetDistUpgradeCmd = "dist-upgrade"
+	aptGetUpgradableArgs = []string{"--just-print", "-qq"}
```
iamsubratp: does it make sense to have a separate package for "package-managers"? currently if patch package has to use anything from package managers, it has to import inventory which does not sense.

ospatch/apt_upgrade.go
```diff
@@ -15,32 +15,15 @@
...
-	// AptGetFullUpgrade specifies apt-get full-upgrade should be run.
-	AptGetFullUpgrade
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+	"github.com/GoogleCloudPlatform/osconfig/inventory/packages"
 )
 
 type aptGetUpgradeOpts struct {
-	upgradeType aptGetUpgradeType
+	upgradeType packages.AptUpgradeType
+	includes    []string
```
iamsubratp: we finalized on the api field names and decided to use "exclusive-(packages|patches)". exclusive_patches for windows and zypper and exclusive_packages for apt and yum

iamsubratp:
/ approve

inventory/packages/apt_deb.go
adjackura: Yeah, I plan on moving the directory eventually. Right now I'm still migration all package calls to this library.

PR #89, opened 2019-10-21
Connect agentendpoint and ospatch

adjackura:
Much of this is simply a copy/paste from the existng ospatchog code tweaked to run via agentendpoint tasks.


agentendpoint/patch_linux.go
```diff
@@ -0,0 +1,70 @@
+//  Copyright 2018 Google Inc. All Rights Reserved.
```
ryanwe: 2019. Or are these files copied?

PR #90, opened 2019-10-21
set a PowerShell execution policy when running an OS Config pre-/post-patch task

wj-chen:
Agent e2e test case "[OSPatch] [PatchJob runs pre-step and post-step] [windows-cloud/windows-2008-r2]" has been failing, and step execution shows the following output:

"File C:\Windows\TEMP\windows_gcs_pre_patch_script.ps1 cannot be loaded because 
running scripts is disabled on this system. For more information, see 
about_Execution_Policies at http://go.microsoft.com/fwlink/?LinkID=135170.
    + CategoryInfo          : SecurityError: (:) [], ParentContainsErrorRecord 
   Exception
    + FullyQualifiedErrorId : UnauthorizedAccess"

Now setting execution policy to RemoteSigned (checked that it's the default of windows-1803-core, windows-2012-r2-core, windows-2019-core, probably others too) for the PowerShell session when executing a step, according to "Set a different policy for one session" at https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_execution_policies?view=powershell-6#managing-the-execution-policy-with-powershell .

/woof

/assign @adjackura


agentendpoint/exec_task.go
```diff
@@ -184,7 +184,7 @@ func (e *execTask) run(ctx context.Context) error {
 		}
 	case agentendpointpb.ExecStepConfig_POWERSHELL:
 		if goos == "windows" {
-			exitCode, err = executeCommand(winPowershell, []string{"-File", localPath}, e.LogLabels)
+			exitCode, err = executeCommand(winPowershell, []string{"-ExecutionPolicy", "RemoteSigned", "-File", localPath}, e.LogLabels)
```
adjackura: Just set this to Bypass, 
also now that I look at this use these flags as well:
"-NonInteractive", "-NoProfile"

wj-chen: Done. I'll update API comment and gcloud help text to mention these accordingly

PR #91, opened 2019-10-22
Update logging, add agentendpoint command to the binary

adjackura:


main.go
```diff
@@ -143,6 +176,15 @@ func main() {
 	case "ospatch":
 		ospatchog.Run(ctx, make(chan struct{}))
 		return
+	case "agentendpoint":
```
ryanwe: Shouldn't this be in addition to "ospatch" or "policies"? I thought using the agent endpoint was an implementation detail of those features even if it's a global switch.

main.go
```diff
@@ -49,6 +54,21 @@ func (l *logWriter) Write(b []byte) (int, error) {
...
+func (s *serialPort) Write(b []byte) (int, error) {
+	c := &serial.Config{Name: s.port, Baud: 115200}
+	p, err := serial.OpenPort(c)
+	if err != nil {
+		return 0, err
+	}
+	defer p.Close()
+
+	return p.Write(b)
+}
```
ryanwe: Should this logic be in `main.go` or would a new `logger.go` module make more sense?

adjackura: Separate files in the same package just act as logical separation, this was such a small addition and main.go is not overloaded I thought an extra file was overkill (boilerplate would be longer than actual code)

main.go
adjackura: We will need to discuss how we want to gate this (or enabled/disable certain features) as it is very different than in the past. This argument (and all these args) are meant for debug running of the agent, the way the endpoint works is that the agent just connects and listens for tasks, there is no additional flag needed to run a patch task or an exec task. Maybe the name is misleading as this doesn't so much have to do with the agentendpoint as it does with WaitForTaskNotification. The args are not meant to be a stable API so we can rethink this in the future.
Because all this is alpha right now I will block the endpoint connection using our current patch alpha flag and not have one for exec separately as we have that tied to patch in the API. In the future if we are to add additional tasks we can decide if we want to block them via a flag on the agent while they are in alpha, or just use the fact that they are alpha and the agent needs to be told to use the alpha API and so all alpha features will be enabled if that is set.

Also the "ospatch" arg here will be removed as that only corresponds to the current patch workflow and not the new one. "guestpolicies" are agent driven so those will stay and we will just move to using the endpoint calls.

PR #93, opened 2019-10-31
Make tasker asynchronous

dowgird:
* Make tasker.Enqueue() non-blocking (unless Close() is called first).

* Make calls to tasker.Enqueue() synchronous

* Replace sleep with synchronous call(Fixes the TODO)

### DIFF
--- a/ospatchog/patch_run.go
+++ b/ospatchog/patch_run.go
@@ -52,11 +52,7 @@ var (
 func initPatch(ctx context.Context) {
 	cancelC = make(chan struct{})
 	disableAutoUpdates()
-	go Run(ctx, cancelC)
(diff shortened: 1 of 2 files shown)


adjackura:
I'll have to think about this for a bit. Right now its important that Enqueue will block. The code in ospatchog is the exception, but all that is getting removed anyway.

dowgird:
Right, I see how blocking is in general necessary here. This PR was more a test case for my tool chain, so I'm closing it now.

Setting up a notification channel sounds like what the Go book endorses: https://golang.org/doc/effective_go.html#chan_of_chan, so I may come back with something similar to that. 

Thanks for looking!

PR #94, opened 2019-11-05
Use the agentendpoint for patch and guestpolicies

adjackura:
All old calls to osconfig replaced with agentendpoint
Log around starting and stopping the TaskNotification aspect of the agent rethought and reimplemented
In the future (post beta) this logic will most likely change but for an alpha feature set this works as its has previously withe the os-config-enabled-prerelease-features metadata key


config/config.go
```diff
@@ -46,50 +46,46 @@ const (
...
-		f = strings.TrimSpace(f)
+		f = strings.ToLower(strings.TrimSpace(f))
 		switch f {
-		case "ospatch":
-			c.osPatchEnabled = true
-		case "ospackage":
-			c.osPackageEnabled = true
+		case "tasknotification", "ospatch": // ospatch is the legacy flag
+			c.taskNotificationEnabled = true
+		case "ospackage", "guestpolicies":
```
manjunathganga: nit: add a similar comment for ospackage to identify it as a legacy flag similar to the comment above for ospatch?

main.go
```diff
@@ -83,11 +87,24 @@ func run(ctx context.Context) {
 			return
 		}
 
-		// This sets up the patching system to run in the background.
-		ospatchog.Configure(ctx)
+		if config.TaskNotificationEnabled() && (taskNotificationClient == nil || taskNotificationClient.Closed()) {
+			// Start WaitForTaskNotification if we need to.
+			taskNotificationClient, err = agentendpoint.NewClient(ctx)
+			if err != nil {
+				logger.Errorf(err.Error())
```
manjunathganga: Under what conditions would this return nil and is it ok to continue after this error is received? Will we retry to get the client?

adjackura: We don't want the entire service to fail if client dialing does (that should be very rare as the dialing happens async, otherwise its a configuration error on our part), we will simple retry in 10m

If this ends up being an error we actually see we can add addition shorter retry logic.

PR #97, opened 2019-11-11
Sync in latest API changes, allow for WaitForTaskNotification to be canceled 

adjackura:
First 4 files are auogenerated and can be ignored.
ReportTaskStart -> StartNextTask
Use ServiceLables provided in Task response


main.go
```diff
@@ -181,10 +181,10 @@ func main() {
 		if err != nil {
 			logger.Fatalf(err.Error())
 		}
-		if err := client.WaitForTaskNotification(ctx); err != nil {
-			logger.Fatalf(err.Error())
+		for {
+			client.WaitForTaskNotification(ctx)
+			time.Sleep(10 * time.Minute)
```
ryanwe: Under what conditions can we end up in this state?

adjackura: This is just for running waitfortasknotification on the command line, and it mirrors what we have with the service. WaitForTaskNotification is reconfigured every 10 min. For the most part this is a noop. The expectation is that if you want to quit you just ctrl-c.

adjackura: Changed to a select so this is clearer

agentendpoint/agentendpoint.go
```diff
@@ -269,30 +289,73 @@ func (c *Client) loadTaskFromState(ctx context.Context) error {
...
+				return
+			default:
+			}
 
-		if err := c.receiveTaskNotification(ctx, noti, token); err != nil {
-			if s, ok := status.FromError(err); ok && s.Code() == codes.PermissionDenied {
-				// Service is not enabled for this project.
-				time.Sleep(config.SvcPollInterval())
-				continue
+			if err := c.waitForTask(ctx); err != nil {
```
jaiminsh: Currently the error handling is divided between two methods for stream failures, is that on purpose?
You should ideally handle all stream errors in WaitForTask and based on the logic of retrying/swallowing up error should pass a known exception to the WaitForTaskNotification method. Does that sound reasonable?

adjackura: Yes, handling it all in waitForTask makes the testing much simpler too. Really all i need is an error for the long sleep right now (PermissionDenied)

PR #103, opened 2019-11-14
Implement Windows service directly, handle systemd stop signal

adjackura:


google-osconfig-agent.service
```diff
@@ -9,6 +9,8 @@ Restart=always
 RestartSec=1
 StartLimitInterval=120
 StartLimitBurst=3
+KillMode=mixed
+KillSignal=SIGTERM
```
hopkiw: this is the default for `KillMode=mixed`, right?

google-osconfig-agent.service
```diff
@@ -9,6 +9,8 @@ Restart=always
 RestartSec=1
 StartLimitInterval=120
 StartLimitBurst=3
+KillMode=mixed
```
hopkiw: Why change this?

main_windows.go
```diff
@@ -86,3 +90,43 @@ func obtainLock() {
...
+				status <- c.CurrentStatus
+			case svc.Stop, svc.Shutdown:
+				cncl()
+			default:
+			}
+		}
+	}
+}
+
+func runService(ctx context.Context, run func(context.Context)) {
```
hopkiw: same here. we own this and are not creating a general-purpose library. don't take functions as arguments.

google-osconfig-agent.service
adjackura: SIGTERM was getting sent to all processes, which was causing issues with things like dpkg and left them in a messy state. If i read the docs correctly this should only send SIGTERM to osconfig, then SIGKILL to all after the timeout

google-osconfig-agent.service
adjackura: Yes, I put it in here to make it clear what signal is getting sent without having to dig through docs first.

PR #107, opened 2019-11-20
Implement patch whitelisting for zypper

iamsubratp:
--with-update is a tricky case to handle with patches. All the other fields/flags in zyppersettings work on patches except for --with-update, which runs on packages.
readmore: https://www.mankier.com/8/zypper#--with-update

Here is the algorithm to prepare patch and package list for zypper:

- Fetch the list of patches with the filters on severity, category
- If user specifies --with-update, fetch the list of packages that will be updated.
- Filter out any package that will be updated as a part of a patch install from the list of packages to be updated;
- to know packages are going to be updated as a part of a patch, we have used "zypper info -t " command
-  for the list of excluded patches specified by user, filter out the patches from the previously generated patch list.

### DIFF
--- a/ospatch/zypper_patch.go
+++ b/ospatch/zypper_patch.go
@@ -17,6 +17,7 @@ package ospatch
 import (
 	"fmt"
 
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
 	"github.com/GoogleCloudPlatform/osconfig/inventory/packages"
 )
 
@@ -122,6 +123,17 @@ func RunZypperPatch(opts ...ZypperPatchOption) error {
 	}
 
 	fPatches, fpkgs, err := runFilter(patches, zOpts.exclusivePatches, zOpts.excludes, zOpts.withUpdate)
+
+	logger.Infof("Updating %d patches.", len(fPatches))
+	logger.Debugf("Patches to be installed: %s", fPatches)
+	logger.Infof("Updating %d packages.", len(fpkgs))
+	logger.Debugf("Packages to be installed: %s", fpkgs)
+
+	if zOpts.dryrun {
+		logger.Infof("Running in dryrun mode, not updating packages.")
+		return nil
+	}
+
 	return packages.ZypperInstall(fPatches, fpkgs)
 }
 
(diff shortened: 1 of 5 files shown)


ospatch/zypper_patch.go
```diff
@@ -67,39 +73,124 @@ func ZypperUpdateWithUpdate(withUpdate bool) ZypperPatchOption {
...
+	for _, patch := range fPatches {
+		args = append(args, "patch:"+patch.Name)
 	}
-	for _, s := range zOpts.severities {
-		args = append(args, "--severity="+s)
+	for _, pkg := range fpkgs {
+		args = append(args, "package:"+pkg.Name)
 	}
 
 	if _, err := zOpts.runner(exec.Command(zypper, args...)); err != nil {
```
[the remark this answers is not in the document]
iamsubratp: makes sense. will change

ospatch/zypper_patch.go
```diff
@@ -25,11 +28,14 @@ var (
...
-	withOptional bool
-	withUpdate   bool
-	runner       func(cmd *exec.Cmd) ([]byte, error)
+	categories       []string
+	severities       []string
+	withOptional     bool
+	withUpdate       bool
+	excludes         []string
+	exclusivePatches []string
+	runner           func(cmd *exec.Cmd) ([]byte, error)
```
adjackura: You can drop the runner now as we don't use it anymore

iamsubratp: done

ospatch/zypper_patch.go
```diff
@@ -67,39 +71,100 @@ func ZypperUpdateWithUpdate(withUpdate bool) ZypperPatchOption {
...
-		args = append(args, "--with-update")
+	patches, err := packages.ZypperPatches(zListOpts...)
+	if err != nil {
+		return err
 	}
-	for _, c := range zOpts.categories {
-		args = append(args, "--category="+c)
+
+	fPatches, fpkgs, err := runFilter(patches, zOpts.exclusivePatches, zOpts.excludes, zOpts.withUpdate)
+	return packages.ZypperInstall(fPatches, fpkgs)
```
adjackura: Add logging and dryrun return like we have with the other systems:
https://github.com/GoogleCloudPlatform/osconfig/blob/master/ospatch/apt_upgrade.go#L89

iamsubratp: done

PR #108, opened 2019-11-20
Fix YUM output parsing

adjackura:
Yum has an annoying feature that formats the output based on terminal size, if no terminal is detected it defaults to 80 columns. This causes issues when we try to parse the output of Yum, we need to figure out what packages are available for update and there is no other builtin command to get us this info so I had two options. The first was to implement a much more advanced parser with some sort of look ahead for potential split lines and attempt to reform them into what we think a package line is. The other was to just attach a terminal of sufficient size to yum so that we could just parse the output line by line like we already are.


inventory/packages/yum.go
```diff
@@ -31,7 +31,8 @@ var (
 	yumInstallArgs       = []string{"install", "--assumeyes"}
 	yumRemoveArgs        = []string{"remove", "--assumeyes"}
 	yumCheckUpdateArgs   = []string{"check-update", "--assumeyes"}
-	yumUpdateArgs        = []string{"update", "--assumeno", "--cacheonly"}
+	yumUpdateArgs        = []string{"update", "--assumeyes"}
+	yumUpdatesArgs       = []string{"update", "--assumeno", "--cacheonly"}
 	yumUpdateMinimalArgs = []string{"update-minimal", "--assumeno", "--cacheonly"}
```
iamsubratp: my understanding is we are using --assumeno as an alternative to list command. consider renaming the variables to reflect the listing behavior. yumlistupdatesargs?

PR #112, opened 2019-11-22
Update Windows update install logic

adjackura:


agentendpoint/patch_windows.go
```diff
@@ -52,53 +52,65 @@ func classFilter(cs []agentendpointpb.WindowsUpdateSettings_Classification) ([]s
...
+
+	if count == 0 {
+		return 0, nil
+	}
+
+	r.infof("%d Windows updates to install", count)
+
+	for i := 0; i < int(count); i++ {
+		if err := r.reportContinuingState(ctx, agentendpointpb.ApplyPatchesTaskProgress_APPLYING_PATCHES); err != nil {
+			return 0, err
```
jaiminsh: Should this i, err as you successfully patched i updates?

agentendpoint/patch_windows.go
```diff
@@ -52,53 +52,65 @@ func classFilter(cs []agentendpointpb.WindowsUpdateSettings_Classification) ([]s
...
+		}
+		updt, err := updts.Item(i)
+		if err != nil {
+			return 0, err
```
jaiminsh: Ditto for this

agentendpoint/patch_windows.go
```diff
@@ -52,53 +52,65 @@ func classFilter(cs []agentendpointpb.WindowsUpdateSettings_Classification) ([]s
...
+
+func (r *patchTask) wuaUpdates(ctx context.Context) error {
 	cf, err := classFilter(r.Task.GetPatchConfig().GetWindowsUpdate().GetClassifications())
 	if err != nil {
 		return err
 	}
 
 	// We keep searching for and installing updates until the count == 0 or there is an error.
-	retries := 50
+	retries := 20
```
iamsubratp: what's the reason that we have to retry; is it that we do not get all updates in one run or is it like the windows update installation are flaky? asking because retries would increase the time to patch an instance

adjackura: Updates are flaky :)

iamsubratp: ok

PR #118, opened 2019-12-12
Bump version, a few fixes for beta candidate 

adjackura:


config/config.go
```diff
@@ -78,14 +78,28 @@ type config struct {
...
 	for _, f := range strings.Split(features, ",") {
 		f = strings.ToLower(strings.TrimSpace(f))
 		switch f {
 		case "tasknotification", "ospatch": // ospatch is the legacy flag
-			c.taskNotificationEnabled = true
+			if enabled {
+				c.taskNotificationEnabled = true
+			} else {
+				c.taskNotificationEnabled = false
+			}
```
iamsubratp: c.taskNotificationEnabled = enabled
?

config/config.go
```diff
@@ -78,14 +78,28 @@ type config struct {
...
+				c.guestPoliciesEnabled = true
+			} else {
+				c.guestPoliciesEnabled = false
+			}
+		case "osinventory":
+			if enabled {
+				c.osInventoryEnabled = true
+			} else {
+				c.osInventoryEnabled = false
+			}
```
iamsubratp: same as above?

e2e_tests/config/config.go
```diff
@@ -41,11 +41,11 @@ var (
 	zones                  map[string]int
 	testZone               = flag.String("test_zone", "", "test zone")
 	testZones              = flag.String("test_zones", "{}", "test zones")
+	projects               []string
+	testProjectIDs         = flag.String("test_project_ids", "", "test project ids")
```
iamsubratp: did you plan to introduce multiple project support because of quota issues?

iamsubratp: btw, i just noticed that we have unstable and staging tests as well that run against the same project which also accounts towards the same quota.
https://github.com/GoogleCloudPlatform/compute-image-tools/blob/ac478ec597b4966d1d67282b5f59432286c5b012/test-infra/prow/config.yaml#L285

do you think, instead of introducing ability of run in multiple projects we should just move unstable and staging tests to different projects.
a concern that i have with current project is that the test vms will spin up in any one of the projects and in case we are tracing some issue, we will have to jump across projects(wouldn't be easy debugging)

adjackura: Yes, its not so much a resource issue as it is a api quota issue, starting so many instances and interacting with them quickly puts the tests up against the standard compute request quota

PR #122, opened 2019-12-16
Fix a few bugs, bump version

adjackura:
Correct IDToken expiry check
Check zypper install exit code
Correctly handle dry run patch option
Don't try to run zypper update when there are no updates available 
Add backoff to receiveTaskNotification for ResourceExhausted


iamsubratp:
looks good to me! will wait for Paul/Jaimin for agent endpoint chagnes

PR #124, opened 2019-12-18
Update e2e_tests to only create 10 instances at a time to limit qps

adjackura:


e2e_tests/compute/instance.go
```diff
@@ -47,50 +45,6 @@ func (i *Instance) Cleanup() {
 	}
 }
 
-// WaitForSerialOutput waits to a string match on a serial port.
-func (i *Instance) WaitForSerialOutput(match string, port int64, interval, timeout time.Duration) error {
```
iamsubratp: is this not being used anymore? iirc, this is what we use for test assertions

adjackura: Yeah, we dropped it in favor of using guest attributes

PR #129, opened 2020-01-03
Update software recipes

adjackura:
Add DpkgInstallation and RpmInstallation step types to software recipes 
Add test for DpkgInstallation and RpmInstallation
Clean up some command line args to be more verbose
Fix zip file extraction on Windows

### DIFF
--- a/inventory/packages/apt_deb.go
+++ b/inventory/packages/apt_deb.go
@@ -232,10 +232,6 @@ func InstalledDebPackages() ([]PkgInfo, error) {
 func DpkgInstall(path string) error {
 	args := append(dpkgInstallArgs, path)
 	out, err := run(exec.Command(dpkg, args...))
-	var msg string
-	for _, s := range strings.Split(string(out), "\n") {
-		msg += fmt.Sprintf(" %s\n", s)
-	}
-	DebugLogger.Printf("dpkg output:\n%s", msg)
+	DebugLogger.Printf("dpkg output:\n%s", strings.ReplaceAll(string(out), "\n", "\n "))
 	return err
 }
--- a/inventory/packages/apt_deb.go
+++ b/inventory/packages/apt_deb.go
@@ -27,10 +27,11 @@ import (
 )
 
 var (
-	// dpkg-query
+	dpkg      string
 	dpkgquery string
 	aptGet    string
 
+	dpkgInstallArgs   = []string{"--install"}
 	dpkgQueryArgs     = []string{"-W", "-f", "${Package} ${Architecture} ${Version}\n"}
 	aptGetInstallArgs = []string{"install", "-y"}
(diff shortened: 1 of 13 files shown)


inventory/packages/apt_deb.go
```diff
@@ -224,3 +227,15 @@ func InstalledDebPackages() ([]PkgInfo, error) {
...
 	return parseInstalledDebpackages(out), nil
 }
+
+// DpkgInstall installs a deb package.
+func DpkgInstall(path string) error {
+	args := append(dpkgInstallArgs, path)
+	out, err := run(exec.Command(dpkg, args...))
+	var msg string
+	for _, s := range strings.Split(string(out), "\n") {
+		msg += fmt.Sprintf(" %s\n", s)
```
hopkiw: you're just.. adding a space to the beginning of every line? i don't know why but if so why not replace `'\n'` with `'\n '`

policies/recipes/artifacts.go
```diff
@@ -34,7 +35,8 @@ import (
 func fetchArtifacts(ctx context.Context, artifacts []*agentendpointpb.SoftwareRecipe_Artifact, directory string) (map[string]string, error) {
 	localNames := make(map[string]string)
 
-	for _, a := range artifacts {
+	for i, a := range artifacts {
+		logger.Debugf("Downloading artifact %d: %q", i, a)
```
hopkiw: np: i don't think the index of the artifact matters at all, since they're all named

policies/recipes/steps.go
```diff
@@ -396,12 +397,32 @@ func stepInstallMsi(step *agentendpointpb.SoftwareRecipe_Step_InstallMsi, artifa
...
+	}
+
+	return packages.DpkgInstall(path)
 }
 
-func stepInstallRpm(step *agentendpointpb.SoftwareRecipe_Step_InstallRpm, artifacts map[string]string, runEnvs []string, stepDir string) error {
-	return fmt.Errorf("InstallRpm not yet supported")
+func stepInstallRpm(step *agentendpointpb.SoftwareRecipe_Step_InstallRpm, artifacts map[string]string) error {
+	if !packages.RPMExists {
+		return fmt.Errorf("rpm does not exist on system")
```
hopkiw: is this error wrapped by step runner? i think the full error message needs to be more like 'InstallRPM step but `rpm` does not exist on system'

inventory/packages/apt_deb.go
adjackura: This is just for debug log readability. I want to go back and fix all these functions debug logging in the future and maybe get rid of this (right now this is just what all the other functions do).

policies/recipes/steps.go
adjackura: Added better error logging at the caller.

PR #130, opened 2020-01-03
Add more checks for Windows reboot required

adjackura:


ospatch/updates_windows.go
```diff
@@ -28,21 +28,46 @@ import (
...
+		`SOFTWARE\Microsoft\Windows\CurrentVersion\WindowsUpdate\Auto Update\RebootRequired`,
+		`SOFTWARE\Microsoft\Windows\CurrentVersion\Component Based Servicing\RebootPending`,
+	}
+	for _, key := range regKeys {
+		logger.Debugf("Checking if reboot required by testing the existance of %s", key)
+		k, err := registry.OpenKey(registry.LOCAL_MACHINE, key, registry.QUERY_VALUE)
+		if err == nil {
+			k.Close()
+			logger.Debugf("%s exists indicating a reboot is required.", key)
+			rebootRequired = true
```
ryanwe: Is it possible for these keys to exist but the value is `0` or `false`?

adjackura: No, they are created in a temporary state, the existence means that a reboot is required (WUA api also checks for this)

ospatch/updates_windows.go
```diff
@@ -28,21 +28,46 @@ import (
...
-	logger.Debugf("Checking if reboot required by looking at %s", regPath)
-	k, err := registry.OpenKey(registry.LOCAL_MACHINE, reg, registry.QUERY_VALUE)
+	// https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefileexw#remarks
+	logger.Debugf("Checking for PendingFileRenameOperations")
+	k, err := registry.OpenKey(registry.LOCAL_MACHINE, `SYSTEM\CurrentControlSet\Control\Session Manager`, registry.QUERY_VALUE)
 	if err != nil {
-		if err == registry.ErrNotExist {
-			logger.Debugf("%s does not exist, indicating no reboot is required.", regPath)
-			return false, nil
+		if err != registry.ErrNotExist {
```
ryanwe: minor: for readability, you may want to mirror the `if err == nil ... else if err != ErrNotExist`. It would remove the extra nesting and add consistency. You could also refactor `if err != nil && err != registry.ErrNotExist` since you duplicate it 3 times in this function.

ospatch/updates_windows.go
```diff
@@ -28,21 +28,46 @@ import (
...
+			return false, err
+		}
+	} else {
+		val, _, err := k.GetStringsValue("PendingFileRenameOperations")
+		if err != nil && err != registry.ErrNotExist {
+			return false, err
+		}
+		k.Close()
```
ryanwe: below you only do this when `err == nil`. Is it ok to call this if there was an error that didn't trigger an early return?

ospatch/updates_windows.go
```diff
@@ -28,21 +28,46 @@ import (
...
+		`SOFTWARE\Microsoft\Windows\CurrentVersion\Component Based Servicing\RebootPending`,
+	}
+	for _, key := range regKeys {
+		logger.Debugf("Checking if reboot required by testing the existance of %s", key)
+		k, err := registry.OpenKey(registry.LOCAL_MACHINE, key, registry.QUERY_VALUE)
+		if err == nil {
+			k.Close()
+			logger.Debugf("%s exists indicating a reboot is required.", key)
+			rebootRequired = true
+			break
```
ryanwe: minor: It would be simpler to `return true, nil` instead of having the mutable state, break, and return at the end.

ospatch/updates_windows.go
adjackura: this would probably be an issue, changed to match other sections

PR #131, opened 2020-01-08
Update metadata configuration

adjackura:
Also minor fix to zypper package install args


config/config.go
```diff
@@ -175,47 +179,92 @@ func createConfigFromMetadata(md metadataJSON) *config {
...
 		}
 	}
+	if md.Instance.Attributes.PollInterval != nil {
+		if val, err := md.Instance.Attributes.PollInterval.Int64(); err == nil {
+			c.osConfigPollInterval = int(val)
+		}
+	}
+
+	switch {
+	case md.Instance.Attributes.DebugEnabledOld != "":
```
wj-chen: Not project first then instance here and for the endpoint setting?

adjackura: Yep, missed, I updated and cleaned up the code more

PR #132, opened 2020-01-10
Update e2e tests

adjackura:
Fix a few bugs with centos/rhel tests

### DIFF
--- a/e2e_tests/test_suites/guestpolicies/guest_policies_utils.go
+++ b/e2e_tests/test_suites/guestpolicies/guest_policies_utils.go
@@ -67,8 +67,9 @@ func getStartupScript(image, pkgManager, packageName string) *computeApi.Metadat
 	switch pkgManager {
 	case "apt":
 		ss = `systemctl stop google-osconfig-agent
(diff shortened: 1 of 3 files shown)


e2e_tests/test_suites/guestpolicies/guest_policies_utils.go
```diff
@@ -243,21 +258,21 @@ baseurl=https://packages.cloud.google.com/yum/repos/osconfig-agent-test-reposito
 enabled=1
 gpgcheck=0
 EOM
-zypper -n remove %[2]s
-zypper -n --no-gpg-checks install %[2]s-3.03-2.fc7
+zypper -n remove ed
```
ryanwe: Is there a chance we could get a false signal if inventory including `ed` is already reported before we remove it?

adjackura: It's not actually installed on anything but centos6, i put these here in case a future image has it. Also we don't look for it until after we remove it and restart the agent

PR #136, opened 2020-01-15
Fix data race in e2e_tests

adjackura:
Clean imports


e2e_tests/test_config/project_config.go
```diff
@@ -46,11 +47,11 @@ func GetProject() *Project {
...
 	}
 
-	testZones := config.Zones()
+	testZones := map[string]int{}
 	var zoneIndices []string
-
-	for z := range testZones {
-		zoneIndices = append(zoneIndices, z)
+	for k,v :=range config.Zones() {
+		testZones[k] = v
```
ryanwe: Would `copy()` work here and make your intention clearer?

ryanwe: But more importantly doesn't this break things. Isn't the idea that all calls to `AquireZone()` take from the same global map `p.testZones` which is an alias for `config.Zones`? By copying `testZones` you're saying that anyone who calls `GetProject` gets their own allotment of zonal resources.

adjackura: copy() is for slices: https://golang.org/pkg/builtin/#copy
This isn't super clean but the way this works now is GetProject will only ever create one Project per project-id that you provide, previously we were incorrectly sharing zones across tests, no we are not

ryanwe: Ah. I think I understand. So the bug was that in addition to the zonal capacity configuration being reused across projects (which we still do), the actual `map` tracking the capacity consumed by the tests was too, right? If so, does that mean that we might have more capacity than we've been allowing ourselves to use because of this and this change will speed up our test runs? It looks like we limit ourselves to 10 instances per zone, I thought we had more capacity than that. Is this some sort of "safe" limit?

Also, +1 for cleaning this up. I'm not sure if the lazy initialization helps more than it hurts readability. We also might want to consider per-project zonal quota though it seems like the quota locks and spreading the load across projects is the most helpful. Either way, that's stuff for another PR.

e2e_tests/compute/instance.go
```diff
@@ -113,19 +112,19 @@ func (i *Instance) RecordSerialOutput(ctx context.Context, storageClient *storag
 		// Instance is stopped or stopping.
 		status, _ := i.client.InstanceStatus(path.Base(i.Project), path.Base(i.Zone), i.Name)
 		if !isTerminal(status) {
-			logger.Errorf("Instance %q: error getting serial port: %s", i.Name, err)
+			fmt.Printf("Instance %q: error getting serial port: %s", i.Name, err)
```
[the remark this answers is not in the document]
adjackura: It is not passed into these functions, and really dont a huge need to complicate the code. If we ever get time we could work on cleaning this up but for this rarer errors just printing will be enough

PR #140, opened 2020-01-22
Fix typo and systemd service file

### LINKED ISSUE #138: Typo in metadata server DNS name in error message
Spotted this by chance while reading code:

https://github.com/GoogleCloudPlatform/osconfig/blob/7b05fdf14f5dfa320cbd6b81418dc46198a83163/config/config.go#L283

should be `metadata.google.internal`

### LINKED ISSUE #139: systemd provider/consumer idiom violation in the .service file, race condition
**Short:** `Wants=network.target` is a bug potentially leading to a race condition, because the target is passive, and the osconfig service is its consumer, not provider.

https://github.com/GoogleCloudPlatform/osconfig/blob/7b05fdf14f5dfa320cbd6b81418dc46198a83163/google-osconfig-agent.service#L3-L4

**Long explanation:** systemd documentation is not written in the most clear way there is; it uses a very formal register of speech but peppered with jargon terms that are never defined. For example, "pulling" a target is used throughout the documentation but is never defined anywhere. No wonder many people get the meaning of special targets wrong. In these two lines the service *pulls* a *passive* target while being a consumer, not producer of the target state, and this is a dependency protocol violation w. r. t. `network.target`.

This is described in systemd.special(7) under the second level heading "Special Passive System Units." 

> [The passive] targets are generally not part of the initial boot transaction, unless they are explicitly pulled in by one of the implementing services. Note specifically that these passive target units are generally not pulled in by the consumer of a service, but by the provider of the service. This means: a consuming service should order itself after these targets (as appropriate), but not pull it in. A providing service should order itself before these targets (as appropriate) and pull it in (via a Wants= type dependency).

The terminology is confusing, so I hope my explanation would be helpful.  *Pulling* a target simply means mentioning it in a `Wants=`, `Requires=` or other directives that make the target considered for the loading set, over which systemd then establishes a load order graph. Ordering directives, like `After` and `Before`, are not even looked at unless a unit is *pulled*.

Conventionally, there are two types of targets, *active* and *passive* (this terminology selection is also rather unfortunate, IMO). *Passive* targets are pulled by *providers* of their functionality, which also add a `Before` dependency on it. *Active* targets are pulled by the *consumers* of the functionality, and they arrange themselves in the order graph `After=` it. It's no coincidence that the set of systemd's special passive targets closely mirrors the [list of LSB init "facilities."](https://refspecs.linuxfoundation.org/LSB_4.1.0/LSB-Core-generic/LSB-Core-generic.pdf)[Sec. 20.6, page 644].

In other words, the `network.target` may or may not run at all, depending on whether or not the networking system available (which makes no sense in GCE; but we're at a general concept). A consumer pulling such a target would only confuse other consumers: it won't *cause* the facility to become magically available. They other consumers will start, but there would in fact be no networking, because the passive target neither Wants or Requires anything by itself. In systemd, systemd-networkd.service(8) is supposed to (and does) pull this target, and has a `Before` dependency on it. It's the only legit service to say "ok, go guys, network services are enabled."

In a typical "passive" target use scenario, a service that wants to use the network to provide network services only (think good old telnetd or ftpd) in theory may declare `WantedBy=network.target` in the `[Install]` section and `After=network.target` in the `[Unit]` section. W.r.t. the `network.target` specifically, this is rarely done in practice: All systemd-provided services are rather triggered on demand by their corresponding `.socket` unit, a mechanism similar to xinetd.

If a service wants a "full blown" network availability immediately on its startup, with IP addresses obtained, DNS client working and routes established, they claim `Wants=network-online.target` and arrange themselves after it. This target is *active*: it Wants a service that blocks the systemd's internal job until network in fact becomes fully-functional. A single `After` without a `Wants` won't do: the provider must be requested to be started.

It is a much more correct approach to do `Requires=network-online.target` if the service is a brick without network, or `Wants=` the same if it can still work without it (again, no practical difference in GCE, where everything is a brick without networking). But w.r.t. the `network.target`, the consuming service is only allowed to order itself `After=` it (for `Before`, there is a similar passive `network-pre.target`);  to Want or Require it is a violation of the consumer/producer conventions of the systemd ecosystem.

In my image build based on the recent GCE Debian 10 image, the `google-osconfig-agent.service` is the only offender (note that some dependencies below may be indirect, as `systemctl show` shows internal, already precomputed closures of both load and dependency sets). Look at the `WantedBy=` set specifically:
```
~$ systemctl show network.target | egrep '^(Want|Requ|Bind|Bound|Before|After|Confl)'
WantedBy=google-osconfig-agent.service systemd-networkd.service
Conflicts=shutdown.target
Before=apt-daily.service systemd-user-sessions.service ssh.service google-startup-scripts.service rc-local.service google-network-daemon.service shutdown.target nfs-server.service mill.mount network-online.target google-accounts-daemon.service chrony.service google-clock-skew-daemon.service google-shutdown-scripts.service google-osconfig-agent.service clusterid-from-metadata.service nfs-mountd.service google-instance-setup.service
After=systemd-networkd.service network-pre.target systemd-resolved.service
```
`systemd-networkd.service` is a provider, so it pulls the target, but other guys only run After it (thus all becoming a Before dependency of the network.target itself).

So, if `systemd-networkd.service` happens to run first, everything is OK; but if `google-osconfig-agent.service` would happen to run first, all the guys listed above in the `Before=` set may get very confused!

And since `network-online.target` implies `network.target`, it's ok to drop the network.target from the After clause too. This would look less redundant, but entirely up to you. :) But `Wants=network.target` is really a bug.

adjackura:
Fixes #138 
Fixes #139

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -277,10 +277,10 @@ func createConfigFromMetadata(md metadataJSON) *config {
 	return c
 }
 
-func formatError(err error) error {
+func formatMetadataError(err error) error {
 	if urlErr, ok := err.(*url.Error); ok {
 		if _, ok := urlErr.Err.(*net.DNSError); ok {
-			return fmt.Errorf("DNS error when requesting metadata, check DNS settings and ensure metadata.internal.google is setup in your hosts file")
+			return fmt.Errorf("DNS error when requesting metadata, check DNS settings and ensure metadata.google.internal is setup in your hosts file")
 		}
 		if _, ok := urlErr.Err.(*net.OpError); ok {
 			return fmt.Errorf("network error when requesting metadata, make sure your instance has an active network and can reach the metadata server")
@@ -302,7 +302,7 @@ func SetConfig() error {
 		// Try up to 3 times to wait for slow network initialization, after
 		// that resort to using defaults and returning the error.
 		if webErrorCount == 2 {
-			webError = formatError(webError)
+			webError = formatMetadataError(webError)
 			break
 		}
 		webErrorCount++
--- a/google-osconfig-agent.service
+++ b/google-osconfig-agent.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Google OSConfig Agent
-After=local-fs.target network-online.target network.target
-Wants=local-fs.target network-online.target network.target
+After=local-fs.target network-online.target
+Wants=local-fs.target network-online.target
 
 [Service]
 ExecStart=/usr/bin/google_osconfig_agent


kkm000:
LGTM, thanks for the quick fix! :)

PR #141, opened 2020-01-23
Update startup logic to prevent hanging

adjackura:
Windows expects a quick signal that the service is started, waiting for metadata could slow that.
Waiting for metadata could prevent the agent from shutting down as quickly as it should (~15s), updated to wait on the canceled context.
Overall there should not be any major changes to startup, just code moved around.

### DIFF
--- a/main.go
+++ b/main.go
@@ -74,12 +74,82 @@ func (s *serialPort) Write(b []byte) (int, error) {
 	return p.Write(b)
 }
 
+var deferredFuncs []func()
+
 func run(ctx context.Context) {
+	// Remove any existing restart file.
+	if err := os.Remove(config.RestartFile()); err != nil && !os.IsNotExist(err) {
+		logger.Errorf("Error removing restart signal file: %v", err)
+	}
+
+	// Setup logging.
+	opts := logger.LogOpts{LoggerName: "OSConfigAgent"}
+	if config.Stdout() {
+		opts.Writers = []io.Writer{os.Stdout}
+	}
+	if runtime.GOOS == "windows" {
+		opts.Writers = append(opts.Writers, &serialPort{"COM1"})
+	}
+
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -74,12 +74,82 @@ func (s *serialPort) Write(b []byte) (int, error) {
 	return p.Write(b)
 }
 
+var deferredFuncs []func()
+
 func run(ctx context.Context) {
+	// Remove any existing restart file.
+	if err := os.Remove(config.RestartFile()); err != nil && !os.IsNotExist(err) {
```
ryanwe: Why do we need to do this and not the OS itself?

main.go
```diff
@@ -74,12 +74,82 @@ func (s *serialPort) Write(b []byte) (int, error) {
...
+	if config.Stdout() {
+		opts.Writers = []io.Writer{os.Stdout}
+	}
+	if runtime.GOOS == "windows" {
+		opts.Writers = append(opts.Writers, &serialPort{"COM1"})
+	}
+
+	// If this call to SetConfig fails (like a metadata error) we can't continue.
+	if err := config.SetConfig(ctx); err != nil {
+		logger.Init(ctx, opts)
```
ryanwe: If `SetConfig` fails, won't `logger.Init()` probably fail too? And if not, would it make more sense to call `logger.Init()` before this and include the logic to exit if it fails?

main.go
adjackura: Not sure what you mean, the restart file is a file only known to the agent, getting each OS service manager to do this would be annoying and probably not even possible on some systems

main.go
adjackura: Logger.Init wont fail unless there is something very wrong, we need SetConfig to get project id and settings for logger, we only do this here so that we will log somewhere in the event of an issue

PR #151, opened 2020-02-07
e2e_tests: Record serial output in artifacts

adjackura:
Instead of uploading directly to GCS, just save serial console output to the artifacts directory that our test wrapper uploads

### DIFF
--- a/e2e_tests/config/config.go
+++ b/e2e_tests/config/config.go
@@ -45,7 +45,7 @@ var (
 	testProjectIDs         = flag.String("test_project_ids", "", "test project ids")
 
 	// OutDir is the out directory to use.
-	OutDir = flag.String("out_dir", "/tmp", "junit xml directory")
+	OutDir = flag.String("out_dir", "/tmp", "artifact directory")
 )
 
 func init() {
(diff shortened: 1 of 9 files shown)


e2e_tests/config/config.go
```diff
@@ -45,7 +45,7 @@ var (
 	testProjectIDs         = flag.String("test_project_ids", "", "test project ids")
 
 	// OutDir is the out directory to use.
-	OutDir = flag.String("out_dir", "/tmp", "junit xml directory")
+	OutDir = flag.String("out_dir", "/tmp", "artifact directory")
```
iamsubratp: is temp the artifact directory? may be use the env variable set by prow

e2e_tests/utils/utils.go
```diff
@@ -29,6 +29,7 @@ import (
 
 var (
 	yumInstallAgent = `
+sleep 10
```
iamsubratp: why this sleep?

e2e_tests/config/config.go
adjackura: That is how it works, prow passes in the flag

e2e_tests/utils/utils.go
adjackura: To attempt to deflake what may be a race condition with upgrading the agent on centos/rhel 7

PR #164, opened 2020-04-13
poll agent config more frequently

iamsubratp:
currently, the agent config is polled at the same frequency as the inventory or guestpolicy run. The current default is 10 mins. So, when a customer enables our service customers have to wait for 10mins before they can see any change.
Decoupled the servicepoll interval from metadapoll interval. Also, currently not allowing customers to change the metadata poll interval since this is agent's property. We can decide if we want customers to be able to specify their own poll interval.

log enabled feature. sample log
Apr 16 10:06:14 debian-tmux-setup OSConfigAgent[12234]: 2020-04-16T10:06:14.1932Z OSConfigAgent Info: OSConfig enabled features status:{GuestPolicies: false, OSInventory: true, PatchManagement: true}

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -20,8 +20,11 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"net"
+	"net/http"
 	"net/url"
+	"os"
 	"runtime"
 	"strconv"
 	"strings"
@@ -29,11 +32,17 @@ import (
 	"time"
 
 	"cloud.google.com/go/compute/metadata"
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
 	"golang.org/x/oauth2/jws"
 )
 
 const (
-	// InstanceMetadata is the instance metadata URL.
+	// metadataIP is the documented metadata server IP address.
+	metadataIP = "169.254.169.254"
+	// metadataHostEnv is the environment variable specifying the
+	// GCE metadata hostname.
+	metadataHostEnv = "GCE_METADATA_HOST"
+	// InstanceMetadata is the compute metadata URL.
 	InstanceMetadata = "http://metadata.google.internal/computeMetadata/v1/instance"
 	// IdentityTokenPath is the instance identity token path.
 	IdentityTokenPath = "instance/service-accounts/default/identity?audience=osconfig.googleapis.com&format=full"
@@ -60,6 +69,7 @@ const (
 	restartFileLinux     = configDirLinux + "/osconfig_agent_restart_required"
 
 	osConfigPollIntervalDefault = 10
+	osConfigMetadataPollTimeout = 300
 )
 
 var (
@@ -70,6 +80,7 @@ var (
 	agentConfig   = &config{}
 	agentConfigMx sync.RWMutex
 	version       string
+	lEtag         = &lastEtag{Etag: "0"}
 )
 
 type config struct {
@@ -97,6 +108,23 @@ func getAgentConfig() config {
 	agentConfigMx.RLock()
 	defer agentConfigMx.RUnlock()
 	return *agentConfig
+}
+
+type lastEtag struct {
+	mu   sync.RWMutex
+	Etag string
+}
+
+func (e *lastEtag) set(etag string) {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	e.Etag = etag
+}
+
+func (e *lastEtag) get() string {
+	e.mu.RLock()
+	defer e.mu.RUnlock()
+	return e.Etag
 }
 
 func parseBool(s string) bool {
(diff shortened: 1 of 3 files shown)


iamsubratp:
testing details: tested custom agent on a vm. here is the agent debug logs

Apr 13 22:25:04 debian-tmux-setup OSConfigAgent[31546]: 2020-04-13T22:25:04.8332Z OSConfigAgent Debug: postAttributeCompressed http://metadata.google.internal/computeMetadata/v1/instance/guest-attributes/guestInventory/PackageUpdates: {Yum:[] Rpm:[] Apt:[] Deb:[] Zypper:[] ZypperPatches:[] Gem:[] Pip:[] GooGet:[] WUA:[] QFE:[]}
Apr 13 22:25:04 debian-tmux-setup OSConfigAgent[31546]: 2020-04-13T22:25:04.8422Z OSConfigAgent Debug: postAttribute http://metadata.google.internal/computeMetadata/v1/instance/guest-attributes/guestInventory/LastUpdated: 2020-04-13T22:25:04Z
Apr 13 22:25:04 debian-tmux-setup OSConfigAgent[31546]: 2020-04-13T22:25:04.8518Z OSConfigAgent Debug: Finished task "Run OSInventory".
Apr 13 22:25:04 debian-tmux-setup OSConfigAgent[31546]: 2020-04-13T22:25:04.8519Z OSConfigAgent Debug: Waiting for tasks to run.


Apr 13 22:26:02 debian-tmux-setup OSConfigAgent[31546]: 2020-04-13T22:26:02.6355Z OSConfigAgent Debug: Polling metadata.
Apr 13 22:27:02 debian-tmux-setup OSConfigAgent[31546]: 2020-04-13T22:27:02.6355Z OSConfigAgent Debug: Polling metadata.
Apr 13 22:28:02 debian-tmux-setup OSConfigAgent[31546]: 2020-04-13T22:28:02.6355Z OSConfigAgent Debug: Polling metadata.
Apr 13 22:29:02 debian-tmux-setup OSConfigAgent[31546]: 2020-04-13T22:29:02.6355Z OSConfigAgent Debug: Polling metadata.
Apr 13 22:30:02 debian-tmux-setup OSConfigAgent[31546]: 2020-04-13T22:30:02.6355Z OSConfigAgent Debug: Polling metadata.
Apr 13 22:31:02 debian-tmux-setup OSConfigAgent[31546]: 2020-04-13T22:31:02.6355Z OSConfigAgent Debug: Polling metadata.
Apr 13 22:32:02 debian-tmux-setup OSConfigAgent[31546]: 2020-04-13T22:32:02.6355Z OSConfigAgent Debug: Polling metadata.
Apr 13 22:33:02 debian-tmux-setup OSConfigAgent[31546]: 2020-04-13T22:33:02.6356Z OSConfigAgent Debug: Polling metadata.
Apr 13 22:34:02 debian-tmux-setup OSConfigAgent[31546]: 2020-04-13T22:34:02.6355Z OSConfigAgent Debug: Polling metadata.
Apr 13 22:35:02 debian-tmux-setup OSConfigAgent[31546]: 2020-04-13T22:35:02.6354Z OSConfigAgent Debug: Polling metadata.

main.go
```diff
@@ -152,14 +152,26 @@ func run(ctx context.Context) {
 }
 
 func runLoop(ctx context.Context) {
+	go func() {
+		metadataPollTicker := time.NewTicker(config.MetadataPollInterval())
+		for {
+			logger.Debugf("Polling metadata.")
```
adjackura, line 158: Don't debug log this as that will just completly fill logs. We need to have a few more log levels so we don't just jump from info to debug logs then we can add more verbose debug logging. I do want to add the Notice level for significant events and potentially only write those logs by default. Then we can use info for more logs and debug can be for really verbose logs. But we may also want a debug level.

main.go
```diff
@@ -152,14 +152,26 @@ func run(ctx context.Context) {
...
+
 	var taskNotificationClient *agentendpoint.Client
 	var err error
 	ticker := time.NewTicker(config.SvcPollInterval())
 	for {
-		if err := config.SetConfig(ctx); err != nil {
-			logger.Errorf(err.Error())
-		}
-
 		if _, err := os.Stat(config.RestartFile()); err == nil {
```
adjackura, line 175: This restart check as well as the following config.TaskNotificationEnabled() checks should be able to be moved as well.

main.go
iamsubratp: yeah..., i definitely think we need more  log level to make debugging easier and at the same time not filling up customers logs

main.go
iamsubratp: i think its better to keep the agent restart tied to pollinterval since that can be controlled through metadata.

config/config.go
```diff
@@ -290,60 +318,120 @@ func formatMetadataError(err error) error {
...
 	}
 
-	new := createConfigFromMetadata(metadata)
+	newAgentConfig := createConfigFromMetadata(metadataConfig)
 	agentConfigMx.Lock()
-	agentConfig = new
+	agentConfig = newAgentConfig
 	agentConfigMx.Unlock()
 
+	logFeatures()
```
adjackura, line 417: Don't do that here, the caller should probably handle this.
We should also only log if there is a change, so for now just skip this.

config/config.go
```diff
@@ -290,60 +318,120 @@ func formatMetadataError(err error) error {
...
 		case <-ctx.Done():
 			return nil
+		default:
+			continue
 		}
 	}
 
-	var metadata metadataJSON
-	if err := json.Unmarshal([]byte(md), &metadata); err != nil {
+	logger.Debugf("Metadata changed.")
```
adjackura, line 404: Don't log this as this ins't necessarily true, we will always get this after the timeout and nothing will change 99% of the time

config/config.go
```diff
@@ -290,60 +318,120 @@ func formatMetadataError(err error) error {
...
-	new := createConfigFromMetadata(metadata)
+	newAgentConfig := createConfigFromMetadata(metadataConfig)
 	agentConfigMx.Lock()
-	agentConfig = new
+	agentConfig = newAgentConfig
 	agentConfigMx.Unlock()
 
+	logFeatures()
+
+	if fn != nil {
```
adjackura, line 419: No need to do this since this call is not an endless loop. Just leave this function in the caller, all WatchConfig should do is wait for a change and then return.

config/config.go
```diff
@@ -290,60 +318,120 @@ func formatMetadataError(err error) error {
...
+		if webError == nil && eTag != lEtag.get() {
 			break
 		}
+
 		// Try up to 3 times to wait for slow network initialization, after
 		// that resort to using defaults and returning the error.
-		if webErrorCount == 2 {
-			webError = formatMetadataError(webError)
-			break
+		if webError != nil {
```
adjackura, line 388: This should probably be in getMetadata if we keep the separate InitConfig function, if we drop that and this is the only way to get metadata, leave this here.

config/config.go
```diff
@@ -290,60 +318,120 @@ func formatMetadataError(err error) error {
...
+	}
+
+	return string(all), resp.Header.Get("Etag"), nil
+}
+
+// InitConfig initializes AgentConfig and resets the lastEtag to "0".
+// InitConfig is only used during the application startup to
+// initialize the config once.
+// This is a stateless API to fetch current config from metadata server.
+func InitConfig(ctx context.Context) error {
```
adjackura, line 361: I don't see a need for this function, just run WatchConfig for our setup, it will just return immediately if it's the first time we run it.

main.go
```diff
@@ -154,12 +154,39 @@ func run(ctx context.Context) {
...
 	var taskNotificationClient *agentendpoint.Client
 	var err error
-	ticker := time.NewTicker(config.SvcPollInterval())
-	for {
-		if err := config.SetConfig(ctx); err != nil {
-			logger.Errorf(err.Error())
+
+	go func() {
+		for {
+			if err := config.WatchConfig(ctx, func() {
```
adjackura, line 160: Remove the func as WatchConfig returns, just have the logic run before WatchConfig here as we already have config setup at this point, then it will just run every time WatchConfig returns

config/config.go
iamsubratp: it wouldn't. after the timeout, if we still get back the lastetag, then we reconnect. this log line will only be reached if there is change in top level metadata

iamsubratp: ofcourse, there might not be changes in keys that are relevant to osconfig

config/config.go
iamsubratp: the problem is, the first call immediately returns with an etag. now i update our global lastetag. in the next call, since there is no change, the etag matches and we never initliaze the agentendpointclient

i could have reset the lastetag explicity in the first call, but that felt like a hack to me. so thought of adding a standard api to config library.

the other option is to use a global agentendpoint client and initialize it on the first watchconfig run. this might not be too bad,

open for better suggestion.

config/config.go
iamsubratp: i had the same feelings that we should only log if there is change in osconfig features, since diff based actions might be a little complicated. since this is an explicit ask from customer, i think its better to start with logging current state of the world, and moving forward we can change this to diff based.

also, since the feature list belonged to config library, i thought keeping the logging in config library would be best.

iamsubratp: ok.. i think it makes sense to move the call to called of watchconfig. sorry for misunderstanding.

config/config.go
iamsubratp: done

main.go
iamsubratp: done

config/config.go
adjackura: For debug level that's fine

config/config.go
adjackura: See my later comment, the caller should handle the actions, and the agentendpoint initialization should hapen at the top of the loop, then the call to watch

config/config.go
iamsubratp: initconfig uses watchconfig anyways. it does not use getmetadata directly.

i dont have any hard opinion on where to place the retry strategy, but my thought was, since we have used retry on metadata api, i did not want to change that.
moving the retry strategy down into getmetadata around the http calls is also fine, but i wouldn't like initconfig to directly call getmetadata config, because then we will have to overload the the initconfig method with preparation of suffix for getmetadata call and we will end up replicating things.

thoughts?

config/config.go
iamsubratp: sorry i marked that conversation done. pulling your comments from there

>>>Remove the func as WatchConfig returns, just have the logic run before WatchConfig here as we already have config setup at this point, then it will just run every time WatchConfig returns

so, iiuc, this is what you are suggesting,
`
go func() {
    for {
        initializesomething()
        fetchconfig()
    }
}
`
this will be functionally correct, however, the for loop reads a little odd. it's not intuitive that we are initializingsomething based on a fetch. it wouldn't read like a normal for loop contruct. 
may be it's ok to add this exploitation since this is an infinite for loop.
however, if you think the readability aspect for this loop is fine with this, happy to make it this way. let me know what you think

adjackura: I don't think the loop reads odd at all, and it makes the logic way simpler. in the loop we configure things, then wait for metadata changes, then configure things and so on.

config/config.go
```diff
@@ -290,60 +318,114 @@ func formatMetadataError(err error) error {
...
+	}
+	all, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", "", err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return "", "", err
+	}
+
+	return string(all), resp.Header.Get("Etag"), nil
```
adjackura, line 354: Just return []byte for metadata, it's way more efficient.

config/config.go
```diff
@@ -290,60 +318,114 @@ func formatMetadataError(err error) error {
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusNotFound {
+		return "", "", err
+	}
+	all, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", "", err
+	}
+	if resp.StatusCode != http.StatusOK {
```
adjackura, line 350: Do this before the ReadAll

iamsubratp: done

config/config.go
```diff
@@ -290,60 +318,98 @@ func formatMetadataError(err error) error {
...
 		}
+
 		// Try up to 3 times to wait for slow network initialization, after
 		// that resort to using defaults and returning the error.
-		if webErrorCount == 2 {
-			webError = formatMetadataError(webError)
-			break
+		if webError != nil {
+			if webErrorCount == 2 {
+				webError = formatMetadataError(webError)
```
adjackura, line 374: just return err here, no need to go through the rest of the function

main.go
```diff
@@ -154,12 +154,39 @@ func run(ctx context.Context) {
...
+				// Cancel WaitForTaskNotification if we need to, this will block if there is
+				// an existing current task running.
+				if err := taskNotificationClient.Close(); err != nil {
+					logger.Errorf(err.Error())
+				}
+			}
+			if err := config.WatchConfig(ctx); err != nil {
+				logger.Errorf(err.Error())
+			}
+			config.LogFeatures()
```
adjackura, line 178: Do this at the top of the loop

iamsubratp: done

config/config.go
iamsubratp: done

PR #169, opened 2020-04-20
Retry installing WUA updates until canceled or retries exceeded

adjackura:
Don't return on first error, keep retrying unless we get a stop signal. Reduce retry number down to a more reasonable 10.

### DIFF
--- a/agentendpoint/patch_beta_windows.go
+++ b/agentendpoint/patch_beta_windows.go
@@ -93,7 +93,7 @@ func (r *patchTaskBeta) installWUAUpdates(ctx context.Context, cf []string) (int
 		defer updt.Release()
 
 		if err := session.InstallWUAUpdate(updt); err != nil {
-			return i, fmt.Errorf(`installUpdate(class, excludes, updt): %v`, err)
+			return i, fmt.Errorf(`installUpdate(updt): %v`, err)
 		}
 	}
 
@@ -106,12 +106,16 @@ func (r *patchTaskBeta) wuaUpdates(ctx context.Context) error {
 		return err
 	}
 
-	// We keep searching for and installing updates until the count == 0 or there is an error.
-	retries := 20
-	for i := 0; i < retries; i++ {
+	// We keep searching for and installing updates until the count == 0,
+	// we get a stop signal, or retries exceed 10.
+	retries := 10
+	for i := 1; i <= retries; i++ {
+		if err := r.reportContinuingState(ctx, agentendpointpb.ApplyPatchesTaskProgress_APPLYING_PATCHES); err != nil {
+			return err
+		}
 		count, err := r.installWUAUpdates(ctx, cf)
 		if err != nil {
-			return err
+			log.Errorf("Error installing Windows updates (attempt %d): %v", i, err)
 		}
 		if count == 0 {
 			return nil
@@ -122,7 +126,8 @@ func (r *patchTaskBeta) wuaUpdates(ctx context.Context) error {
 }
 
 func (r *patchTaskBeta) runUpdates(ctx context.Context) error {
-	if err := retryFunc(30*time.Minute, "installing Windows updates", func() error { return r.wuaUpdates(ctx) }); err != nil {
+	// Don't use retry function as wuaUpdates handles it's own retries.
+	if err := r.wuaUpdates(ctx); err != nil {
 		return err
 	}
 
--- a/agentendpoint/patch_beta_windows.go
+++ b/agentendpoint/patch_beta_windows.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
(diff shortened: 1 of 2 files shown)


agentendpoint/patch_beta_windows.go
```diff
@@ -106,12 +107,16 @@ func (r *patchTaskBeta) wuaUpdates(ctx context.Context) error {
...
 
-	// We keep searching for and installing updates until the count == 0 or there is an error.
-	retries := 20
-	for i := 0; i < retries; i++ {
+	// We keep searching for and installing updates until the count == 0,
+	// we get a stop signal, or retries exceed 10.
+	retries := 10
+	for i := 1; i <= retries; i++ {
+		if err := r.reportContinuingState(ctx, agentendpointpb.ApplyPatchesTaskProgress_APPLYING_PATCHES); err != nil {
+			return err
```
manjunathganga, line 115: Wrap error to provide stacktrace?

agentendpoint/patch_beta_windows.go
```diff
@@ -106,12 +107,16 @@ func (r *patchTaskBeta) wuaUpdates(ctx context.Context) error {
...
 	}
 
-	// We keep searching for and installing updates until the count == 0 or there is an error.
-	retries := 20
-	for i := 0; i < retries; i++ {
+	// We keep searching for and installing updates until the count == 0,
+	// we get a stop signal, or retries exceed 10.
+	retries := 10
+	for i := 1; i <= retries; i++ {
+		if err := r.reportContinuingState(ctx, agentendpointpb.ApplyPatchesTaskProgress_APPLYING_PATCHES); err != nil {
```
jaiminsh, line 114: If AgentEndpoint returns an error while ReportTaskProgress this will quit the loop, should we retry then as well? 
As mentioned in the comment, we should stop only when we get a stop signal.

agentendpoint/patch_beta_windows.go
```diff
@@ -106,12 +107,16 @@ func (r *patchTaskBeta) wuaUpdates(ctx context.Context) error {
...
+			return err
+		}
 		count, err := r.installWUAUpdates(ctx, cf)
 		if err != nil {
-			return err
+			logger.Errorf("Error installing Windows updates (attempt %d): %v", i, err)
```
manjunathganga, line 119: continue?

agentendpoint/patch_beta_windows.go
```diff
@@ -122,7 +127,8 @@ func (r *patchTaskBeta) wuaUpdates(ctx context.Context) error {
 }
 
 func (r *patchTaskBeta) runUpdates(ctx context.Context) error {
-	if err := retryFunc(30*time.Minute, "installing Windows updates", func() error { return r.wuaUpdates(ctx) }); err != nil {
+	// Don't use retry function as wuaUpdates handles it's own retries.
+	if err := r.wuaUpdates(ctx); err != nil {
 		return err
```
manjunathganga, line 132: Wrap error to provide stacktrace?

agentendpoint/patch_windows.go
```diff
@@ -107,12 +108,16 @@ func (r *patchTask) wuaUpdates(ctx context.Context) error {
...
 
-	// We keep searching for and installing updates until the count == 0 or there is an error.
-	retries := 20
-	for i := 0; i < retries; i++ {
+	// We keep searching for and installing updates until the count == 0,
+	// we get a stop signal, or retries exceed 10.
+	retries := 10
+	for i := 1; i <= retries; i++ {
+		if err := r.reportContinuingState(ctx, agentendpointpb.ApplyPatchesTaskProgress_APPLYING_PATCHES); err != nil {
+			return err
```
manjunathganga, line 116: Do we need to add/update tests foe this change?

agentendpoint/patch_beta_windows.go
adjackura: reportContinuingState provides the context already, this error path is most likely because of a cancelled or timed out job and the error logs make that clear (doesn't really matter that it came from exactly here).

I am looking into structured error messages though to allow for clear text logs but also almost stacktrace like debuggability from structured logs.

agentendpoint/patch_beta_windows.go
adjackura: reportContinuingState will return an error because of a cancelled or timed out job, it has it's own retries on api errors and we don't need an additional layer.

agentendpoint/patch_beta_windows.go
adjackura: The only code path for the underlying functions is through here so tracing is not a huge concern, and this can return error for a cancelled task and we don't want to wrap that as it can get confusing.

agentendpoint/patch_windows.go
adjackura: We currently can't unit test the system calls (WUA api), our e2e test handle the full flow but we don't exactly have a test for task cancelation right now. This change is strictly logging and moving the retry loop though.

PR #173, opened 2020-06-12
Handle transient failures in guestpolicy operations

iamsubratp:
- refactored out retry utility functions, so that they can be used in other part of the applications
- Added retries to guestpolicy operations

### DIFF
--- a/policies/policies.go
+++ b/policies/policies.go
@@ -23,12 +23,14 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
 	"github.com/GoogleCloudPlatform/osconfig/agentendpoint"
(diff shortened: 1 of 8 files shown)


policies/policies.go
```diff
@@ -175,7 +177,9 @@ func setConfig(egp *agentendpointpb.EffectiveGuestPolicy) {
 		if err := googetRepositories(gooRepos, config.GooGetRepoFilePath()); err != nil {
 			logger.Errorf("Error writing googet repo file: %v", err)
 		}
-		if err := googetChanges(gooInstallPkgs, gooRemovePkgs, gooUpdatePkgs); err != nil {
+		if err := retryutil.RetryFunc(2*time.Minute, "Applying googet changes", func() error {
```
ryanwe, line 180: Since these all use the same time, you may want to use a constant instead.

PR #174, opened 2020-06-16
Run 'dpkg --configure -a' when needed.

adjackura:


packages/apt_deb.go
```diff
@@ -88,6 +89,19 @@ func AptGetUpgradeShowNew(showNew bool) AptGetUpgradeOption {
...
 }
 
+func dpkgRepair(out []byte) bool {
+	// Error code 100 may occur for non repairable errors, just check the output.
+	if !bytes.Contains(out, []byte("dpkg --configure -a")) {
+		return false
+	}
+	DebugLogger.Printf("apt-get error, attempting dpkg repair.")
+	// Ignore error here, just log and rerun apt-get.
+	out, _ = run(exec.Command(dpkg, dpkgRepairArgs...))
```
ryanwe, line 99: Should we log the error too instead of completely ignoring it?

adjackura: We log out, thats really all that will be needed here, the error would just confuse things and is only helpful if there is no such path or some fundamental exec issue (unlikely)

PR #177, opened 2020-06-19
watch on osconfig specific metadata changes

iamsubratp:
without this, watchmetadata would return on any change in for this instance which this makes the agent logs a little spammy.

from now, the agent keeps a hash of the current agentconfig and only returns from watchconfig if there is anything that changes in the agentconfig.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -17,6 +17,7 @@ package config
 
 import (
 	"context"
+	"crypto/sha256"
 	"encoding/json"
 	"flag"
 	"fmt"
@@ -102,6 +103,13 @@ func (c *config) parseFeatures(features string, enabled bool) {
 			c.osInventoryEnabled = enabled
 		}
 	}
+}
+
+func (c *config) asSha256() string {
+	h := sha256.New()
+	h.Write([]byte(fmt.Sprintf("%v", c)))
+
+	return fmt.Sprintf("%x", h.Sum(nil))
 }
 
 func getAgentConfig() config {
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -77,10 +78,11 @@ var (
 	debug    = flag.Bool("debug", false, "set debug log verbosity")
 	stdout   = flag.Bool("stdout", false, "log to stdout")
 
-	agentConfig   = &config{}
-	agentConfigMx sync.RWMutex
-	version       string
-	lEtag         = &lastEtag{Etag: "0"}
+	agentConfig      = &config{}
+	lAgentConfigHash = ""
```
ryanwe, line 82: minor: why `l`? Is it for `last`? Isn't this the hash of the `agentConfig`?

config/config.go
```diff
@@ -364,7 +373,22 @@ func WatchConfig(ctx context.Context) error {
...
+			var metadataConfig metadataJSON
+			if err := json.Unmarshal(md, &metadataConfig); err != nil {
+				return err
+			}
+
+			newAgentConfig := createConfigFromMetadata(metadataConfig)
+			newAgentConfigHash := newAgentConfig.asSha256()
+
+			if lAgentConfigHash != newAgentConfigHash {
+				agentConfigMx.Lock()
```
ryanwe, line 386: Since `agentConfig` and `lAgentConfigHash` represent the same data, should they both be protected by the lock? Or would it be better to calculate `lAgentConfigHash` on the fly instead of caching it and dealing with redundant state?

config/config.go
iamsubratp: right. `l` is for last. the prefix is consistent with other variables that has been used to cache previous values, for example `lEtag`

config/config.go
iamsubratp: i thought it would be unnecessary to use lock for `lAgentConfigHash` because it is not a shared variable.
Also, `agentConfig` is written only by the current thread. So, the operations, calculating hash of `agentConfig` and updating `agentConfig`, are sequential.

config/config.go
iamsubratp: ok.. i see your concern around not being consistent with `agentConfig`. `lAgentConfigHash` does not sound like it is the hash `agentConfig`.

I will go ahead and change the variable name to `agentConfigHash`

config/config.go
adjackura: We don't need to worry about trying to be super efficient with the lock, just do this for simplicity and to avoid more global vars:

```
agentConfigMx.Lock()
if agentConfig.asSha256() != newAgentConfig.asSha256() {
  agentConfig = newAgentConfig
  agentConfigMx.Unlock()
  break
}
agentConfigMx.Unlock()
```

iamsubratp: sure

iamsubratp: done

PR #179, opened 2020-06-24
make agent talk to zonal agentendpoint

iamsubratp:
with zonal agentendpoints, agents will now be talking to zonal agentendpoints.

the zonal endpoint is crafted by prefixing `osconfig-endpoint` metadata with the instancezone name.
The assumption here is that, zone value is always present in instancemetadata
referrred this
https://cloud.google.com/compute/docs/storing-retrieving-metadata#default

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -298,6 +298,12 @@ func createConfigFromMetadata(md metadataJSON) *config {
 		c.debugEnabled = true
 	}
 
+	setSVCEndpoint(md, c)
+
+	return c
+}
+
+func setSVCEndpoint(md metadataJSON, c *config) {
 	switch {
 	case *endpoint != prodEndpoint:
 		c.svcEndpoint = *endpoint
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -311,7 +317,9 @@ func createConfigFromMetadata(md metadataJSON) *config {
 		c.svcEndpoint = md.Project.Attributes.OSConfigEndpointOld
 	}
 
-	return c
+	// Example instanceZone: projects/1071514831017/zones/us-west1-b
+	parts := strings.Split(c.instanceZone, "/")
+	c.svcEndpoint = fmt.Sprintf("%s-%s", parts[len(parts)-1], c.svcEndpoint)
```
ryanwe, line 322: I like that this makes it easy to add the zone when someone overrides the endpoint, but what if someone wants to test against a non-zonal endpoint (eg. the existing global endpoint or a dev environment)?

adjackura: I think if the flag is set we should honor it, and only prepend the zone if the default is being used

iamsubratp: adding relevant offline dicussion for others


Ryan: I think respecting the override completely might also make things like testing staging difficult.
Maybe we could allow for a wild card (eg. `{zone}-staging.osconfig.sandbox.googleapis.com` vs `staging.osconfig.sandbox.googleapis.com`)

Andrew Jackura, 16 mins
yeah, hat would work
better than my idea of anther option to set
we should just do a string replace
so if {zone} is not set its just a noop

PR #181, opened 2020-06-26
Add source field in googet spec.

gaohannk:


packaging/googet/google-osconfig-agent.goospec
```diff
@@ -5,6 +5,7 @@
   "authors": "Google Inc.",
   "license": "http://www.apache.org/licenses/LICENSE-2.0",
   "description": "Google OSConfig agent",
+  "source": "https://github.com/GoogleCloudPlatform/osconfig"
```
adjackura, line 8: need a trailing comma

PR #185, opened 2020-07-08
add timeout to watchconfig

iamsubratp:
Add an overall timeout for watchconfig and reduced the metadata watch timeout. this is to handle the case where agent is enabled and but the service is not enabled.
If there is no change in the metadata, the agent never gets a chance to establish a connection with the service.
the current timeout is 10 mins.
In the situtation mentioned above, it will take a total of 11 mins

also, removed the feature logging for now, since it will spam the serial console of vm incurring extra cost to customers. we will come up with a better plan for logging.

### DIFF
--- a/main.go
+++ b/main.go
@@ -157,7 +157,6 @@ func runLoop(ctx context.Context) {
 
 	go func() {
 		for {
-			config.LogFeatures()
 			if config.TaskNotificationEnabled() && (taskNotificationClient == nil || taskNotificationClient.Closed()) {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -157,7 +157,7 @@ func runLoop(ctx context.Context) {
 
 	go func() {
 		for {
-			config.LogFeatures()
+			//config.LogFeatures()
```
ryanwe, line 160: minor: can you delete this instead of leaving it commented out?

iamsubratp: done

PR #187, opened 2020-07-09
Add RegisterAgent rpcs, update libraries, update error handling

adjackura:
RegisterAgent currently unused
agentendpoint tests had to be updated because of new client library retries (really wasn't testing something that we cared about)
Moved a few error handling to new go 1.13 errors.Is and errors.As


config/config.go
```diff
@@ -83,6 +84,8 @@ var (
 	agentConfigMx sync.RWMutex
 	version       string
 	lEtag         = &lastEtag{Etag: "0"}
+
+	capabilities = []string{"PATCH_GA", "GUEST_POLICY_BETA"}
```
jaiminsh, line 88: Maybe have a comment on how these values are decided?

adjackura: Made a comment for how they are used, i don't really want to publish how we decide them beyond this.

PR #191, opened 2020-07-22
Add context logging, make sure any actions the agent takes are logged at info level

adjackura:
Added context logging library and moved almost all logging calls to it.
Any labels added get added as labels to any log call that uses the ctx or child context
My plan is to move this up to the core logging library once it supports all the features we want.


agentendpoint/patch_windows.go
```diff
@@ -128,24 +128,25 @@ func (r *patchTask) wuaUpdates(ctx context.Context) error {
 }
 
 func (r *patchTask) runUpdates(ctx context.Context) error {
-	// Don't use retry function as wuaUpdates handles it's own retries.
-	if err := r.wuaUpdates(ctx); err != nil {
-		return err
-	}
-
+	// Install GooGet updates first as this will allow us to update the agent prior to any potential WUA bugs/errors.
```
ryanwe, line 131: Should this be a separate PR?

main.go
```diff
@@ -110,9 +97,14 @@ func run(ctx context.Context) {
 		fmt.Printf("Error initializing logger: %v", err)
 		os.Exit(1)
 	}
-	packages.DebugLogger = log.New(&logWriter{}, "", 0)
+	ctx = clog.WithLabels(ctx, map[string]string{"instance_name": config.Name()})
```
ryanwe, line 100: Isn't this already included in the log event since it comes from the instance?

ospatch/apt_upgrade.go
```diff
@@ -89,12 +91,12 @@ func RunAptGetUpgrade(opts ...AptGetUpgradeOption) error {
 		pkgNames = append(pkgNames, pkg.Name)
 	}
 	logger.Infof("Updating %d packages.", len(pkgNames))
-	logger.Debugf("Packages to be installed: %s", fPkgs)
+	logger.Infof("Packages to be installed: %s", fPkgs)
```
ryanwe, line 94: It may also be clearer to combine this with the above log line since they're both `Info` now. Maybe` "Updating %d packages: %s", len(pkgNames), fPkgs `. Or you could go further and combine it with the "dryrun" message too. So for dryruns, log `Dryrun, not updating %d packages: %s", len(pkgNames), fPkgs `.

And fyi that people may code against this. So even if we add structured logging in the future, we may need to keep this log message around.

ospatch/yum_update.go
```diff
@@ -106,12 +108,12 @@ func RunYumUpdate(opts ...YumUpdateOption) error {
 		pkgNames = append(pkgNames, pkg.Name)
 	}
 	logger.Infof("Updating %d packages.", len(pkgNames))
-	logger.Debugf("Packages to be installed: %s", fPkgs)
+	logger.Infof("Packages to be installed: %s", fPkgs)
```
ryanwe, line 111: Same comment as above.

ospatch/zypper_patch.go
```diff
@@ -146,22 +148,22 @@ func RunZypperPatch(opts ...ZypperPatchOption) error {
 		logger.Infof("No patches to install.")
 	} else {
 		logger.Infof("Installing %d patches.", len(fPatches))
-		logger.Debugf("Patches to be installed: %s", fPatches)
+		logger.Infof("Patches to be installed: %s", fPatches)
```
ryanwe, line 151: Same comment as above.

ospatch/zypper_patch.go
```diff
@@ -146,22 +148,22 @@ func RunZypperPatch(opts ...ZypperPatchOption) error {
...
 	}
 
 	if len(fpkgs) == 0 {
 		logger.Infof("No non-patch packages to update.")
 	} else {
 		logger.Infof("Updating %d packages.", len(fpkgs))
-		logger.Debugf("Packages to be installed: %s", fpkgs)
+		logger.Infof("Packages to be installed: %s", fpkgs)
```
ryanwe, line 158: here too. This should also allow you to remove the `if len() == 0` check.

retryutil/retry.go
```diff
@@ -94,7 +95,7 @@ func RetryAPICall(maxRetryTime time.Duration, name string, f func() error) error
 			return err
 		}
 
-		logger.Errorf("Error calling %s, attempt %d, retrying in %s: %v", name, i, ns, err)
+		clog.Errorf(ctx, "Error calling %s, attempt %d, retrying in %s: %v", name, i, ns, err)
```
ryanwe, line 98: Should this be a Warning since we're still retrying?

main.go
adjackura: Thats just instance id unfortunately.

ospatch/apt_upgrade.go
adjackura: Ack on the structured logging. What I think i may end up doing is keeping the message field as the human readable form, and then adding more structure around that. So still one log entry just with both this message and also more structured fields.

clog/clog.go
```diff
@@ -0,0 +1,118 @@
...
+	labels map[string]string
+	ctx    context.Context
+}
+
+func (l *log) log(msg string, sev logger.Severity) {
+	logger.Log(logger.LogEntry{Message: msg, Severity: sev, CallDepth: 4, Labels: l.labels})
+}
+
+// Debugf simulates logger.Debugf and adds context labels.
+func Debugf(ctx context.Context, format string, args ...interface{}) {
```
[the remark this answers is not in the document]
adjackura: We can't make it invisible to customers, if you set the log level that level will be used, we can add more debug levels, so level 1 is less verbose, 2 is more, 3 is even more and so on. If we want to add this I can do it in another PR.

PR #200, opened 2020-07-30
Inventory agent call prototype

wj-chen:
Prototype-only, not ready for official review

### DIFF
--- a/agentendpoint/agentendpoint.go
+++ b/agentendpoint/agentendpoint.go
@@ -16,7 +16,10 @@
 package agentendpoint
 
 import (
+	"bytes"
 	"context"
+	"crypto/sha256"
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -145,6 +145,13 @@ func run(ctx context.Context) {
 		runLoop(ctx)
 	case "inventory", "osinventory":
 		inventory.Run()
```
adjackura, line 147: I think we should remove this function and move all the inventory reporting to ReportInventory, its important that we dont query local inventory twice as it is a heavyweight process. So ReportInventory can gather inventory, write to guest attributes, then send it to osconfig. This also lets us easily migrate to a task based system in the future but keep the guest attributes reporting aspect.

PR #203, opened 2020-08-03
report inventory to agent endpoint

wj-chen:
Also moved content in inventory/inventory_test.go to agentendpoint/inventory_test.go.


agentendpoint/inventory.go
```diff
@@ -0,0 +1,272 @@
...
+func formatWUAPackage(pkg packages.WUAPackage) *agentendpointpb.Inventory_SoftwarePackage_WuaPackage {
+	var categories []*agentendpointpb.Inventory_WindowsUpdatePackage_WindowsUpdateCategory
+	for idx, category := range pkg.Categories {
+		categories = append(categories, &agentendpointpb.Inventory_WindowsUpdatePackage_WindowsUpdateCategory{
+			Id:   pkg.CategoryIDs[idx],
+			Name: category,
+		})
+	}
+
+	supportUrls := []string{pkg.SupportURL}
```
wj-chen, line 232: @adjackura @paulseawa let me know if the supportUrls var is supposed to be more than this

agentendpoint/inventory.go
```diff
@@ -0,0 +1,272 @@
...
+	"time"
+
+	"github.com/GoogleCloudPlatform/guest-logging-go/logger"
+	"github.com/GoogleCloudPlatform/osconfig/attributes"
+	"github.com/GoogleCloudPlatform/osconfig/config"
+	agentendpointpb "github.com/GoogleCloudPlatform/osconfig/internal/google.golang.org/genproto/googleapis/cloud/osconfig/agentendpoint/v1alpha1"
+	"github.com/GoogleCloudPlatform/osconfig/inventory"
+	"github.com/GoogleCloudPlatform/osconfig/packages"
+	"github.com/golang/protobuf/ptypes"
+	timestamp "github.com/golang/protobuf/ptypes/timestamp"
```
adjackura, line 17: timestamppb

Also don't use any github.com/golang/protobuf libraries (recently deprecated), use google.golang.org/protobuf instead

agentendpoint/inventory.go
```diff
@@ -0,0 +1,272 @@
...
+
+// ReportInventory reports inventory to agent endpoint and writes it to guest attributes.
+func (c *Client) ReportInventory(ctx context.Context) {
+	state := inventory.Get()
+	write(state, inventoryURL)
+	c.report(ctx, state)
+}
+
+func write(state *inventory.InstanceInventory, url string) {
+	logger.Debugf("Writing instance inventory to guest attributes.")
```
adjackura, line 33: Sorry to do this, but I just pushed new logging system, use clog now for all logging calls (will need to bring context into all your functions).

agentendpoint/inventory.go
adjackura: Actually it should just be SupportURL, and we need to add [MoreInfoUrls](https://docs.microsoft.com/en-us/windows/win32/api/wuapi/nf-wuapi-iupdate-get_moreinfourls) to the agent and the api.

We can replace SupportUrls with MoreInfoUrls in the api as SupportUrl is not populated much of the time.

main.go
```diff
@@ -216,7 +221,13 @@ func runLoop(ctx context.Context) {
 
 		if config.OSInventoryEnabled() {
 			// This should always run after ospackage.SetConfig.
-			inventory.Run()
+			client, err := agentendpoint.NewClient(ctx)
+			if err != nil {
+				logger.Fatalf(err.Error())
```
adjackura, line 226: Don't log fatal here, just error. Actually for this implementation you should create the client inside tasker.Enqueue (dont need to change the "inventory" case above as that is only for the command line)

agentendpoint/inventory.go
wj-chen: Done

wj-chen: Done

main.go
wj-chen: Updated. Anything special happens with fatal logging?

agentendpoint/inventory.go
wj-chen: If you merge master into development I'll pull and update the logging system

agentendpoint/inventory.go
```diff
@@ -0,0 +1,260 @@
...
+	var categories []*agentendpointpb.Inventory_WindowsUpdatePackage_WindowsUpdateCategory
+	for idx, category := range pkg.Categories {
+		categories = append(categories, &agentendpointpb.Inventory_WindowsUpdatePackage_WindowsUpdateCategory{
+			Id:   pkg.CategoryIDs[idx],
+			Name: category,
+		})
+	}
+
+	// TODO: Populate supportUrls with MoreInfoUrls.
+	supportUrls := []string{}
```
wj-chen, line 232: @adjackura Added a TODO here and setting support URLs to be empty list in the meantime

main.go
adjackura: Fatal is the same as Error but with and os.Exit(1)  :)

agentendpoint/inventory.go
wj-chen: Done

main.go
wj-chen: I see thanks

PR #204, opened 2020-08-05
fix missing yum flags passed during yum update

iamsubratp:
* currently we do not pass update-minimal and security flag when customer sets these options.
* introduced commandRunner to make flags testing easier.
* refactored YumUpdates method to workaround exitcode 100 from yum check-update


packages/packages.go
```diff
@@ -95,6 +102,46 @@ type QFEPackage struct {
...
+// CommandRunner will execute the commands and return the results of that
+// execution.
+type CommandRunner interface {
+
+	// RunCommand takes precreated exec.Cmd and returns the results of execution.
+	RunCommand(command *exec.Cmd) ([]byte, error)
+
+	// Run takes string arguments of command to be executed
+	// and returns the results of execution.
+	Run(command string, args ...string) ([]byte, error)
```
ryanwe, line 114: It doesn't look like this function is ever called. Would it be possible to use this function instead of RunCommand? It seems redundant for callers to write `Runner.RunCommand(exec.Command(...))` instead of `Runner.RunCommand(...)`. Is it possible, to remove `RunCommand`, change `RunWithPty` to take `command string, args ...string` and have the CommandRunner wrap `exec.Command` too?

packages/yum.go
```diff
@@ -185,15 +157,37 @@ func YumUpdates(opts ...YumUpdateOption) ([]PkgInfo, error) {
...
+}
+
+func parseAndUpdateYumPackages(opts ...YumUpdateOption) ([]PkgInfo, error) {
+	yumOpts := &yumUpdateOpts{
+		security: false,
+		minimal:  false,
+	}
+
+	for _, opt := range opts {
+		opt(yumOpts)
```
ryanwe, line 176: This seems like a roundabout way to pass two booleans. Would it be clearer if this function, and `YumUpdates` take two params `security: bool, minimal: bool` instead?

packages/yum_test.go
```diff
@@ -16,47 +16,136 @@ package packages
...
+	Obsoleting:
+	  baz                                       noarch                         2.0.0-1                                              repo                                      10 M
+`)
+
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+
+	mockCommandRunner := mocks.NewMockCommandRunner(mockCtrl)
+	Runner = mockCommandRunner
+	mockCommandRunner.EXPECT().RunWithPty(exec.Command("/usr/bin/yum", []string{"update-minimal", "--assumeno", "--cacheonly", "--security"}...)).Return(data, nil).Times(1)
```
ryanwe, line 112: Nice tests!

packages/packages.go
```diff
@@ -95,6 +102,46 @@ type QFEPackage struct {
 	Caption, Description, HotFixID, InstalledOn string
 }
 
+// CommandRunner will execute the commands and return the results of that
+// execution.
+type CommandRunner interface {
```
ryanwe, line 107: Should this be in a more common location than `packages/packages.go`? Maybe `utils`?

packages/packages.go
iamsubratp: i added ```Run(...)``` to provide a simpler variation of ```RunCommand()```, that i was using in tests(for test setup), but ultimately didn't end up using it, but i left it with the idea that it can be used if a client of this CommandRunner does not want to create exec.Cmd on their own.
if i were to remove, i would prefer removing ```Run(...)``` and keep the signatures of ```RunWithPty() && RunCommand()``` to accept ```exec.Cmd```, since that is more intuitive(clients will read this as ```create exec.Cmd with exec.Command(...) and send it to commandRunner to execute the Cmd```)

packages/yum.go
iamsubratp: it is a general pattern that i have seen being used in golang code to pass around options. the same pattern in used for package managers as well.

packages/yum_test.go
iamsubratp: 😄

packages/packages.go
iamsubratp: moved to utils

packages/packages.go
```diff
@@ -50,8 +51,15 @@ var (
 
 	// DebugLogger is the debug logger to use.
 	DebugLogger = log.New(ioutil.Discard, "", 0)
+
+	// Runner is the CommandRunner used for running exec commands.
+	Runner util.CommandRunner
```
adjackura, line 56: Does this have to be public? Can the mocks not just be generated privately for this library?

util/util.go
```diff
@@ -70,3 +71,19 @@ func Exists(name string) bool {
 	}
 	return true
 }
+
+// CommandRunner will execute the commands and return the results of that
+// execution.
+type CommandRunner interface {
```
adjackura, line 77: This shouldn't be in util, it's not common. just keep in in packages. For the other libraries that need something similar we should create their own interface. The contains methods that are only ever used in one package, i don't want other packages to have to implement those.

packages/yum.go
adjackura: This is best practice for go options, you can add options without breaking the function interface.

packages/packages.go
iamsubratp: in order to inject the mockrunner, this has to be public since the injection happens in the test package

util/util.go
iamsubratp: yeah.. i just realized, runwithpty would be very specific to package package.

packages/packages.go
iamsubratp: ignore the last comment. that is not true. however, it had to be public otherwise my tests failed with panic, and changing the scope resolved it.

iamsubratp: sorry about the confusions from previous commments. i think i merged 2 issues that i faced yesterday

it does not have to public. changed it

packages/packages.go
iamsubratp: this is a specific example of using exec.Command in tests.
https://github.com/adjackura/osconfig/blob/5354872db63e6a6e86ba6c0aacb3253fabab95f0/inventory/packages/yum_test.go#L77

added here for reference

packages/yum.go
iamsubratp: @ryanwe now i understand what you meant. i think i accidentally left the options parsing in original function that caused the confusion. it should be good now

PR #208, opened 2020-08-06
Fix clog panic

adjackura:


clog/clog.go
```diff
@@ -77,7 +77,7 @@ func forContext(ctx context.Context) (*log, context.Context) {
 	cv := ctx.Value(ctxValueKey)
 	l, ok := cv.(*log)
 	if !ok {
-		l = &log{}
+		l = &log{labels: map[string]string{}}
```
iamsubratp, line 80: pulling code from fromContext as i could not commment there. 

```
func fromContext(ctx context.Context) *log {
	if ctx == nil {
		return &log{}
	}

	v := ctx.Value(ctxValueKey)
	l, ok := v.(*log)
	if !ok {
		l = &log{}
	}
	return l
}
```

we are creating new instances of log at multiple different places.
does this mean we need to pass an initialized map whenever we create a new instance of log?

PR #210, opened 2020-08-07
Fix WatchMetadata web error handling, fix clog CallDepth

adjackura:
We need to actually wait for slow network bringup, make sure WatchMetadata doesn't loop faster than 5s, wait up to 60s for slow network
logger.Log has an incorrect default call depth of 2, my calculation based on this was therefore incorrect.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/json"
-	"errors"
 	"flag"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -383,7 +382,10 @@ func getMetadata(suffix string) ([]byte, string, error) {
 func WatchConfig(ctx context.Context) error {
```
iamsubratp, line 382: since this watchconfig is becoming such a critical piecd, i think it is worth add mock tickers and test this method thoroughly in unit test. wdyt?

PR #218, opened 2020-08-12
Add framework for config task

adjackura:
Added fairly in depth tests to test the logic, tests will need to be updated once each config step is implemented.


agentendpoint/config_task.go
```diff
@@ -0,0 +1,462 @@
...
+	}
+	if err := c.client.reportTaskComplete(ctx, req); err != nil {
+		return fmt.Errorf("error reporting completed state: %v", err)
+	}
+	return nil
+}
+
+func (c *configTask) handleErrorState(ctx context.Context, msg string, err error) error {
+	if err == errServerCancel {
+		clog.Infof(ctx, "Canceling config run: %v", errServerCancel)
```
ryanwe, line 105: "Cancelling" with 2 'l's

agentendpoint/config_task.go
```diff
@@ -0,0 +1,462 @@
...
+
+func (c *configTask) reportContinuingState(ctx context.Context, configState agentendpointpb.ApplyConfigTaskProgress_State) error {
+	req := &agentendpointpb.ReportTaskProgressRequest{
+		TaskId:   c.TaskID,
+		TaskType: agentendpointpb.TaskType_APPLY_CONFIG_TASK,
+		Progress: &agentendpointpb.ReportTaskProgressRequest_ApplyConfigTaskProgress{
+			ApplyConfigTaskProgress: &agentendpointpb.ApplyConfigTaskProgress{State: configState},
+		},
+	}
+	res, err := c.client.reportTaskProgress(ctx, req)
```
ryanwe, line 121: Does this go through the retry util?

agentendpoint/config_task.go
```diff
@@ -0,0 +1,462 @@
...
+		return fmt.Errorf("error reporting task progress %s: %v", configState, err)
+	}
+	if res.GetTaskDirective() == agentendpointpb.TaskDirective_STOP {
+		return errServerCancel
+	}
+	return nil
+}
+
+func (c *configTask) validation(ctx context.Context) {
+	// First populate validation results and internal assignment state.
```
ryanwe, line 132: I thought validation was ensuring the agent understands the resources and checks for conflicts. Mutating state here breaks the SRP. Should those two things (populating pre-enforcement, internal assignment state and validating the task) be separate functions?

agentendpoint/config_task.go
```diff
@@ -0,0 +1,462 @@
...
+				result.GetExecutionSteps()[validationStepIndex] = &agentendpointpb.ApplyConfigTaskOutput_ResourceResult_ExcecutionStep{
+					Step: &agentendpointpb.ApplyConfigTaskOutput_ResourceResult_ExcecutionStep_Validation{
+						Validation: &agentendpointpb.ApplyConfigTaskOutput_ResourceResult_Validation{
+							Outcome: outcome,
+						},
+					},
+					ErrMsg: errMsg,
+				}
+
+				// Should we continue to validate all resources, or stop on the first unmarshal error?
```
ryanwe, line 187: If it's all the same, reporting an error for all the resources would be more helpful to users.

agentendpoint/config_task.go
```diff
@@ -0,0 +1,462 @@
...
+
+func (c *configTask) checkState(ctx context.Context) {
+	// First populate check state results (for policies that do not have validation errors).
+	for i, a := range c.Task.GetConfig().GetConfigAssignments() {
+		asgnmnt := c.assignments[a.GetConfigAssignment()]
+		aResult := c.results.GetResults()[i]
+		for i, p := range a.GetPolicies() {
+			plcy := asgnmnt.policies[p.GetId()]
+			// Skip state check if this policy already has an error from a previous step.
+			if plcy.hasError {
```
ryanwe, line 206: I thought that we wanted to fail fast if there is any policy validation error. Did that change?

agentendpoint/agentendpoint.go
```diff
@@ -227,6 +227,10 @@ func (c *Client) runTask(ctx context.Context) {
 			if err := c.RunExecStep(ctx, task); err != nil {
 				clog.Errorf(ctx, "Error running TaskType_EXEC_STEP_TASK: %v", err)
 			}
+		case agentendpointpb.TaskType_APPLY_CONFIG_TASK:
+			if err := c.RunApplyConfig(ctx, task); err != nil {
+				clog.Errorf(ctx, "Error running TaskType_APPLY_CONFIG_TASK: %v", err)
```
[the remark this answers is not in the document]
adjackura: This is in line with the other messages, we can change this error text across each of these, I'll do this in another PR

agentendpoint/config_task.go
```diff
@@ -0,0 +1,462 @@
...
+
+func (c *configTask) run(ctx context.Context) error {
+	clog.Infof(ctx, "Beginning apply config task.")
+	c.StartedAt = time.Now()
+	rcsErrMsg := "Error reporting continuing state"
+	if err := c.reportContinuingState(ctx, agentendpointpb.ApplyConfigTaskProgress_STARTED); err != nil {
+		return c.handleErrorState(ctx, rcsErrMsg, err)
+	}
+
+	if len(c.Task.GetConfig().GetConfigAssignments()) == 0 {
```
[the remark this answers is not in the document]
adjackura: Probably not, doesnt hurt for the agent to handle this though.

agentendpoint/config_task.go
adjackura: yes

agentendpoint/config_task.go
adjackura: That was my thinking as well

agentendpoint/config_task.go
adjackura: I think we went back and forth on this, I'll double check and update

agentendpoint/config_task.go
adjackura: The assignment state here is required for validation steps (unmarshal, check conflicts), this 'state' is just this particular config tasks understanding of the given assignments, not any persistent state in the agent. For example, it needs to have a place to unmarshal the resources to so it can then check for conflicts.

agentendpoint/config_task.go
```diff
@@ -0,0 +1,462 @@
...
+	// We need to generate base results first thing, each execution step
+	// just adds on.
+	c.generateBaseResults()
+
+	if err := c.reportContinuingState(ctx, agentendpointpb.ApplyConfigTaskProgress_VALIDATING); err != nil {
+		return c.handleErrorState(ctx, rcsErrMsg, err)
+	}
+	c.validation(ctx)
+
+	if err := c.reportContinuingState(ctx, agentendpointpb.ApplyConfigTaskProgress_CHECKING_DESIRED_STATE); err != nil {
```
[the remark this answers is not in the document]
adjackura: We may not, we have quick back to back calls for patch as well. We could drop all the states and replace with a basic APPLYING_CONFIGS that we only call once every 5s at most, but still needs to be at simple stop points, so like this:

report STARTED
generateBaseResults() // no real work
validation() // no real work
report APPLYING_CONFIGS
checkState() // may be fast, could be slow
report APPLYING_CONFIGS // only if its been 5s
enforceState()
report APPLYING_CONFIGS // only if its been 5s
postCheckState()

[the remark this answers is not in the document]
ryanwe: SGTM. We may also want to consider combining reportStart() and reportApplying() into a single API.

agentendpoint/config_task.go
```diff
@@ -0,0 +1,462 @@
...
+	req := &agentendpointpb.ReportTaskCompleteRequest{
+		TaskId:       c.TaskID,
+		TaskType:     agentendpointpb.TaskType_APPLY_CONFIG_TASK,
+		ErrorMessage: errMsg,
+		Output: &agentendpointpb.ReportTaskCompleteRequest_ApplyConfigTaskOutput{
+			ApplyConfigTaskOutput: &agentendpointpb.ApplyConfigTaskOutput{State: state, ConfigAssignmentResults: c.results},
+		},
+	}
+	if err := c.client.reportTaskComplete(ctx, req); err != nil {
+		return fmt.Errorf("error reporting completed state: %v", err)
```
manjunathganga, line 98: Do we need to qualify this error message is for config task?

agentendpoint/config_task.go
```diff
@@ -0,0 +1,462 @@
...
+		}
+	}
+
+	// Unmarshal all resources.
+	for i, a := range c.Task.GetConfig().GetConfigAssignments() {
+		ctx = clog.WithLabels(ctx, map[string]string{"config_assignment": a.GetConfigAssignment()})
+		asgnmnt := c.assignments[a.GetConfigAssignment()]
+		aResult := c.results.GetResults()[i]
+		for i, p := range a.GetPolicies() {
+			plcy := asgnmnt.policies[p.GetId()]
```
manjunathganga, line 162: do you have to update ctx to add policy id to the labels?
similar comment for resource id

agentendpoint/config_task.go
manjunathganga: We should continue to validate other resources so that all resource errors can be reported together and fixed.

agentendpoint/config_task.go
adjackura: So our current approach is to just fail for this policy, not the entire config run as policies are meant to be separate and not interact with each other.

agentendpoint/config_task.go
```diff
@@ -0,0 +1,462 @@
...
+		for i, p := range a.GetPolicies() {
+			ctx = clog.WithLabels(ctx, map[string]string{"policy_id": p.GetId()})
+			plcy := asgnmnt.policies[p.GetId()]
+			// Skip state check if this policy already has an error from a previous step.
+			if plcy.hasError {
+				clog.Debugf(ctx, "Policy has error, skipping state check.")
+				continue
+			}
+			pResult := aResult.GetOsPolicyResults().GetResults()[i]
+			for i, r := range p.GetResources() {
```
manjunathganga, line 234: nit: having variables named`r, res, result` in the same code block is confusing when reading the code. Consider renaming these variables to make the distinction clear.

agentendpoint/config_task.go
```diff
@@ -0,0 +1,459 @@
...
+				errMsg := ""
+				err := res.checkState(ctx)
+				if err != nil {
+					outcome = agentendpointpb.ApplyConfigTaskOutput_ResourceResult_CheckDesiredState_ERROR
+					plcy.hasError = true
+					errMsg = fmt.Sprintf("Error running desired state check: %v", err)
+					clog.Errorf(ctx, errMsg)
+				}
+
+				if res.enforcementNeeded {
```
manjunathganga, line 248: could this be renamed to `inDesiredState` as it seems to better align as a value set by checkState?

agentendpoint/config_task.go
```diff
@@ -0,0 +1,459 @@
...
+		}
+		c.postCheckState(ctx)
+	}
+
+	return c.reportCompletedState(ctx, "", agentendpointpb.ApplyConfigTaskOutput_SUCCEEDED)
+}
+
+// RunApplyConfig runs an apply config task.
+func (c *Client) RunApplyConfig(ctx context.Context, task *agentendpointpb.Task) error {
+	ctx = clog.WithLabels(ctx, task.GetServiceLabels())
```
manjunathganga, line 450: add task id to context?

agentendpoint/config_task.go
adjackura: Yes, I actually already renamed in in the follow up PR i was working on :)

agentendpoint/config_task.go
adjackura: Thats what this should do, any labels the service sends down.

agentendpoint/config_task.go
adjackura: It already will be based on the callers adding error context

agentendpoint/config_task.go
adjackura: Just missed it for this function, added

agentendpoint/config_task.go
```diff
@@ -283,16 +285,16 @@ func (c *configTask) enforceState(ctx context.Context) {
...
 			pResult := aResult.GetOsPolicyResults().GetResults()[i]
-			for i, r := range p.GetResources() {
-				ctx = clog.WithLabels(ctx, map[string]string{"resource_id": r.GetId()})
-				res := plcy.resources[r.GetId()]
+			for i, taskResource := range p.GetResources() {
+				ctx = clog.WithLabels(ctx, map[string]string{"resource_id": taskResource.GetId()})
+				res := plcy.resources[taskResource.GetId()]
 				// Only enforce resources that need it.
-				if !res.enforcementNeeded {
+				if !res.inDesiredState {
```
manjunathganga, line 292: Inverse the condition?
Same in all other places where enforcementNeeded was used before.

PR #219, opened 2020-08-14
Fix panic if any osinfo calls fail

adjackura:
Add more error details to WIndows inventory calls
Record osinfo details wven if some functions return errors.
Reorder windows osinfo functions in order of most likely to not return error.

### DIFF
--- a/osinfo/osinfo_windows.go
+++ b/osinfo/osinfo_windows.go
@@ -127,7 +127,7 @@ func getKernelInfo() (string, string, error) {
 	// This should be something like 040904b0 for US English UTF16LE.
 	langCodePage, err := getTranslation(info)
 	if err != nil {
-		return "", "", err
+		return "", "", fmt.Errorf("getTranslation() error: %v", err)
 	}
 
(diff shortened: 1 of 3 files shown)


osinfo/osinfo_windows.go
```diff
@@ -127,45 +127,40 @@ func getKernelInfo() (string, string, error) {
...
-	hn, err := os.Hostname()
-	if err != nil {
-		return oi, fmt.Errorf("os.Hostname() error: %v", err)
+	var ops []win32OperatingSystem
+	query := "SELECT Caption, Version FROM Win32_OperatingSystem"
+	if err := wmi.Query(query, &ops); err != nil {
+		return oi, fmt.Errorf("wmi.Query(%q) error: %v", query, err)
 	}
-	oi.Hostname = hn
+	oi.LongName = ops[0].Caption
```
ryanwe, line 162: Would it be safer to include bounds checking here?

adjackura: Probably a good idea of the calling function changes.

PR #221, opened 2020-08-20
Move all packages run calls to CommandRunner

adjackura:
This is mostly a replacement of the existing run function in the "packages" package to use our new CommandRunner interface and all the associated tests.
Tests were also refactored to use a common style and reduce code duplication.
CommandRunner.Run was updated to return separated stdout and stderr, the parsing functions now just use stdout to avoid any potential errors (fixes an issue parsing pip output.)


packages/packages_linux.go
```diff
@@ -142,7 +141,6 @@ func GetInstalledPackages(ctx context.Context) (Packages, error) {
 		if err != nil {
 			msg := fmt.Sprintf("error listing installed pip packages: %v", err)
 			clog.Debugf(ctx, "Error: %s", msg)
```
ryanwe, line 143: If these errors don't count toward the success/failure of this function, should we raise their log level? On one hand, we don't want to be too verbose for common issues. On the other, silently failing like this can be really confusing. I almost wonder if we'd want a `logOnce` function to avoid log spew.

util/mocks/mock_command_runner.go
```diff
@@ -1,8 +1,22 @@
+//  Copyright 2019 Google Inc. All Rights Reserved.
```
ryanwe, line 1: 2020

packages/packages_linux.go
adjackura: We specifically dont log pip and gem errors because they are very prone to issues, I just missed these functions when I removed the errors.

PR #223, opened 2020-08-27
Make sure we retry on failed Windows Updates

adjackura:
On a WUA error keep retrying for the full 10 cycles on any errors sleeping 60s between retries.


adjackura:
I don't want to use the retryutil functions for this as this is not retrying over a specific time period but a number of cycles and the exit criteria are no more updates available not just no errors.

PR #224, opened 2020-09-01
Rework COM object memory management

adjackura:
Used Application Verifier from Windows SDK to evaluate heap leaks
Also reorganized and how we call mains deferred functions so we actually log the shutdown call to cloud logging.

### DIFF
--- a/main.go
+++ b/main.go
@@ -25,7 +25,6 @@ import (
 	"os/signal"
 	"path/filepath"
 	"runtime"
-	"runtime/debug"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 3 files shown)


adjackura:
/hold

Running stress tests now.

main.go
```diff
@@ -264,4 +252,8 @@ func main() {
 	default:
 		run(ctx)
 	}
+
+	for _, f := range deferredFuncs {
```
ryanwe, line 256: Just curious, why not use `defer` instead?

PR #225, opened 2020-09-01
Update WUA session creation and error logging

adjackura:
Simplify session creation
Add more context to errors

### DIFF
--- a/packages/wua_windows.go
+++ b/packages/wua_windows.go
@@ -27,38 +27,41 @@ var wuaSession sync.Mutex
 
 // IUpdateSession is a an IUpdateSession.
 type IUpdateSession struct {
-	con *ole.Connection
-	ses *ole.Dispatch
+	*ole.IDispatch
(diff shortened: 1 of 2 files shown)


packages/wua_windows.go
```diff
@@ -27,38 +27,41 @@ var wuaSession sync.Mutex
...
+	*ole.IDispatch
 }
 
 func NewUpdateSession() (*IUpdateSession, error) {
 	wuaSession.Lock()
 	if err := ole.CoInitializeEx(0, ole.COINIT_MULTITHREADED); err != nil {
-		wuaSession.Unlock()
-		return nil, err
+		e, ok := err.(*ole.OleError)
+		// S_OK = 0, S_FALSE = 1, both are Success codes.
```
ryanwe, line 37: Are these available as constants somewhere? If not, maybe make them consts here in favor of self-documented code.

PR #226, opened 2020-09-02
Cleanup policy code, set API retry time to 30s

adjackura:
Set LookupEffectiveGuestPolicies retry time to 30s, this allows local configs to still run in a timely manner even if there is some service issue. Since GuestPolciies run every 10min they should get reapplied pretty soon anyway.
Cleaned up unneeded step of creating a blank repo file as the *Repositories functions will do that themselves
For GuestPolicies only gather package inventory when needed


adjackura:
/hold

Wait for merge until current release is complete

ryanwe (approved):

PR #227, opened 2020-09-03
Add understanding of resources and their verious methods to config code

adjackura:
Existing config package moved to agentconfig (unchanged)
The new config package is now for config management  code
Synced latest autogenerated libraries

Config code path makes use of two interfaces, one in config_task.go for the agent implementation of resources in general, and the second in config.go so that each resource type just needs to satisfy a basic common interface. The config packages role is the core implementation of how to manage each resource type (validate, check state, enforce state), it has no knowledge of polciies or assignments, just resources. The agentendpoint package handles all the logic of iterating through assignments and polcies, reporting state, as well as conflict detection inside and across policies (still needs to be implemented).


config/config.go
```diff
@@ -12,594 +12,91 @@
...
-
-	return fmt.Sprintf("%x", h.Sum(nil))
-}
-
-func getAgentConfig() config {
-	agentConfigMx.RLock()
-	defer agentConfigMx.RUnlock()
-	return *agentConfig
+	managedResources *ManagedResources
+	inDS             bool
```
manjunathganga, line 31: consider expanding DS to DesiredState for better readability.

config/config.go
```diff
@@ -12,594 +12,91 @@
...
+func (r *Resource) InDesiredState() bool {
+	return r.inDS
 }
 
-func (e *lastEtag) set(etag string) {
-	e.mu.Lock()
-	defer e.mu.Unlock()
-	e.Etag = etag
+// ManagedResources returns the resources that this resources manages.
+func (r *Resource) ManagedResources() *ManagedResources {
```
manjunathganga, line 41: Hmm, what does this mean? Do resources manage other resources?

config/config.go
```diff
@@ -12,594 +12,91 @@
...
+// CheckState or EnforceState should be run prior to calling InDesiredState.
+func (r *Resource) InDesiredState() bool {
+	return r.inDS
 }
 
-func (e *lastEtag) set(etag string) {
-	e.mu.Lock()
-	defer e.mu.Unlock()
-	e.Etag = etag
+// ManagedResources returns the resources that this resources manages.
```
manjunathganga, line 40: nit: resource

config/config.go
```diff
@@ -12,594 +12,91 @@
...
-	// GCE metadata hostname.
-	metadataHostEnv = "GCE_METADATA_HOST"
-	// InstanceMetadata is the compute metadata URL.
-	InstanceMetadata = "http://metadata.google.internal/computeMetadata/v1/instance"
-	// IdentityTokenPath is the instance identity token path.
-	IdentityTokenPath = "instance/service-accounts/default/identity?audience=osconfig.googleapis.com&format=full"
-	// ReportURL is the guest attributes endpoint.
-	ReportURL = InstanceMetadata + "/guest-attributes"
+// Resource is a single config resource.
+type Resource struct {
```
manjunathganga, line 26: Would it be better to qualify this as ConfigResource or OSPolicyResource since we also have another interface name resource

agentendpoint/config_task.go
```diff
@@ -126,54 +119,57 @@ func (c *configTask) reportContinuingState(ctx context.Context, configState agen
...
-
-	// Unmarshal all resources.
-	for i, a := range c.Task.GetConfig().GetConfigAssignments() {
-		ctx = clog.WithLabels(ctx, map[string]string{"config_assignment": a.GetConfigAssignment()})
-		assgnmnt := c.assignments[a.GetConfigAssignment()]
-		aResult := c.results.GetResults()[i]
 		for i, p := range a.GetPolicies() {
 			ctx = clog.WithLabels(ctx, map[string]string{"policy_id": p.GetId()})
-			plcy := assgnmnt.policies[p.GetId()]
+			policyMR := &config.ManagedResources{}
```
manjunathganga, line 150: Looks like this field is not getting updated?

agentendpoint/config_task.go
```diff
@@ -32,9 +33,15 @@ const (
 	postCheckDesiredStateStepIndex = 3
 )
 
+var newResource = func(r *agentendpointpb.ApplyConfigTask_Config_Resource) resourceIface {
```
manjunathganga, line 36: This is an overall comment for the file and not specific to this line.
The word resource appears in multiple contexts and reading for the first time, it is not immediately obvious which resource are we talking about without reading much further. Consider renaming the types so that it is easy to quickly decipher which flavor of resource we are talking about.

agentendpoint/config_task.go
manjunathganga, line 150: Consider refactoring each level of for loop into it's own function for easier readability.

adjackura: done

adjackura: It's meant to happen in detectPolicyConflicts (logic still to be implemented)

config/config.go
adjackura: renamed to OSPolicyResource as I think thats the naming we want to use

config/config.go
adjackura: Wasn't sure of a better name here, ManagedItems maybe? this is the actual objects this resource manages like packages, files or other item.

Now that I renamed Resource to OSPolicyResource this may not be as confusing.

agentendpoint/config_task.go
```diff
@@ -225,50 +286,64 @@ func (c *configTask) checkState(ctx context.Context) {
...
+		},
+		ErrMsg: errMsg,
+	}
+	return true
+}
+
+func enforceOSPolicyState(ctx context.Context, assgnmnt *assignment, osPolicy *agentendpointpb.ApplyConfigTask_Config_OSPolicy, pResult *agentendpointpb.ApplyConfigTaskOutput_OSPolicyResult) (actionTaken bool) {
+	ctx = clog.WithLabels(ctx, map[string]string{"policy_id": osPolicy.GetId()})
+	plcy := assgnmnt.policies[osPolicy.GetId()]
+	// Skip enforcement if this policy already has an error from a previous step.
```
manjunathganga, line 329: Should you also skip enforcement if the policy is in validation mode?

agentendpoint/config_task.go
```diff
@@ -277,51 +352,56 @@ func (c *configTask) enforceState(ctx context.Context) {
...
+			},
+		},
+		ErrMsg: errMsg,
+	}
+}
+
+func postCheckOSPolicyState(ctx context.Context, assgnmnt *assignment, osPolicy *agentendpointpb.ApplyConfigTask_Config_OSPolicy, pResult *agentendpointpb.ApplyConfigTaskOutput_OSPolicyResult) {
+	ctx = clog.WithLabels(ctx, map[string]string{"policy_id": osPolicy.GetId()})
+	plcy := assgnmnt.policies[osPolicy.GetId()]
+	// Skip post check if this policy already has an error from a previous step.
```
manjunathganga, line 396: Should you also skip enforcement if the policy is in validation mode?

manjunathganga (approved):

PR #229, opened 2020-09-15
Set RPM package gpgcheck=0 until osconfig allows API control.

jm2:
Right now the barrier to entry for RPM-based distros is higher than DEB-based ones in osconfig, as both package and repository signing are enabled by default for RPM-based distros, while only repository signing is enabled by default for DEB-based ones[1]. Leaving individual RPM GPG signing disabled IMO is a more reasonable default, at least until OSConfig allows this option to be set via the API. Repository GPG signing remains as-is.

[1] https://blog.packagecloud.io/eng/2014/10/28/howto-gpg-sign-verify-deb-packages-apt-repositories/

### DIFF
--- a/policies/yum.go
+++ b/policies/yum.go
@@ -35,14 +35,14 @@ func yumRepositories(ctx context.Context, repos []*agentendpointpb.YumRepository
 		name=repo1-name
 		baseurl=https://repo1-url
 		enabled=1
-		gpgcheck=1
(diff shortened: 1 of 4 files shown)


adjackura:
This is actually the opposite of the norm, yum repo metadata goes unsigned all the time but passing around unsigned rpms is pretty much never accepted best practice.

PR #230, opened 2020-09-16
Add validate and check state functionality for PackageResource

adjackura:
Validate fully implemented and unit tested
Check state still needs unit tests as well as implementation for msi/deb/rpm packages.

### DIFF
--- a/config/package_resouce.go
+++ b/config/package_resouce.go
@@ -175,23 +175,23 @@ var yumInstalled []packages.PkgInfo
 var zypperInstalled []packages.PkgInfo
 var rpmInstalled []packages.PkgInfo
 
-func needsInstall(installedPkgs []packages.PkgInfo, name string) bool {
+func pkgInstalled(installedPkgs []packages.PkgInfo, name string) bool {
 	for _, pkg := range installedPkgs {
 		if pkg.Name == name {
-			return false
+			return true
 		}
 	}
-	return true
+	return false
 }
 
-func (p *packageResouce) checkState() (bool, error) {
+func (p *packageResouce) checkState() (inDesiredState bool, err error) {
 	switch {
 	case p.policy.Apt.Install != nil:
-		if needsInstall(aptInstalled, p.policy.Apt.Install.GetName()) {
+		if pkgInstalled(aptInstalled, p.policy.Apt.Install.GetName()) {
 			return true, nil
 		}
 	case p.policy.Apt.Remove != nil:
-		if !needsInstall(aptInstalled, p.policy.Apt.Remove.GetName()) {
+		if !pkgInstalled(aptInstalled, p.policy.Apt.Remove.GetName()) {
 			return true, nil
 		}
 
@@ -200,11 +200,11 @@ func (p *packageResouce) checkState() (bool, error) {
 	case p.policy.Deb.Remove != nil:
 
 	case p.policy.GooGet.Install != nil:
-		if needsInstall(aptInstalled, p.policy.GooGet.Install.GetName()) {
+		if pkgInstalled(aptInstalled, p.policy.GooGet.Install.GetName()) {
 			return true, nil
 		}
 	case p.policy.GooGet.Remove != nil:
-		if !needsInstall(aptInstalled, p.policy.GooGet.Remove.GetName()) {
+		if !pkgInstalled(aptInstalled, p.policy.GooGet.Remove.GetName()) {
 			return true, nil
 		}
 
@@ -213,20 +213,20 @@ func (p *packageResouce) checkState() (bool, error) {
 	case p.policy.MSI.Remove != nil:
 
 	case p.policy.Yum.Install != nil:
-		if needsInstall(aptInstalled, p.policy.Yum.Install.GetName()) {
+		if pkgInstalled(aptInstalled, p.policy.Yum.Install.GetName()) {
 			return true, nil
 		}
 	case p.policy.Yum.Remove != nil:
-		if !needsInstall(aptInstalled, p.policy.Yum.Remove.GetName()) {
+		if !pkgInstalled(aptInstalled, p.policy.Yum.Remove.GetName()) {
 			return true, nil
 		}
 
 	case p.policy.Zypper.Install != nil:
-		if needsInstall(aptInstalled, p.policy.Zypper.Install.GetName()) {
+		if pkgInstalled(aptInstalled, p.policy.Zypper.Install.GetName()) {
 			return true, nil
 		}
 	case p.policy.Zypper.Remove != nil:
-		if !needsInstall(aptInstalled, p.policy.Zypper.Remove.GetName()) {
+		if !pkgInstalled(aptInstalled, p.policy.Zypper.Remove.GetName()) {
 			return true, nil
 		}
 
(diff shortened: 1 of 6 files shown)


packages/msi.go
```diff
@@ -0,0 +1,34 @@
...
+package packages
+
+import (
+	"runtime"
+
+	"github.com/GoogleCloudPlatform/osconfig/util"
+)
+
+var (
+	msiexec string
```
manjunathganga, line 24: nit: msiExec?

config/package_resouce.go
```diff
@@ -14,26 +14,229 @@
...
+	"fmt"
+
+	agentendpointpb "github.com/GoogleCloudPlatform/osconfig/internal/google.golang.org/genproto/googleapis/cloud/osconfig/agentendpoint/v1alpha1"
+	"github.com/GoogleCloudPlatform/osconfig/packages"
+)
 
 type packageResouce struct {
 	*agentendpointpb.ApplyConfigTask_Config_Resource_PackageResource
+
+	policy ManagedPackage
```
manjunathganga, line 28: Should this be named managedPackage?

config/package_resouce.go
```diff
@@ -14,26 +14,229 @@
...
+}
+
+// AptPackage describes an apt package policy.
```
manjunathganga, line 31: The word policy might lead to some confusion since we already have an os policy

config/package_resouce.go
```diff
@@ -14,26 +14,229 @@
...
+func needsInstall(installedPkgs []packages.PkgInfo, name string) bool {
+	for _, pkg := range installedPkgs {
+		if pkg.Name == name {
+			return false
+		}
+	}
+	return true
 }
 
 func (p *packageResouce) checkState() (bool, error) {
```
manjunathganga, line 187: Would it be better to rename the method or have a named return value so that it is clear what the boolean values mean?
https://g3doc.corp.google.com/go/g3doc/style/decisions.md?cl=head#named-result-parameters

Currently, it is not clear what true or false return values mean.

config/package_resouce.go
```diff
@@ -14,26 +14,229 @@
...
+		if !needsInstall(aptInstalled, p.policy.Apt.Remove.GetName()) {
+			return true, nil
+		}
+
+	// TODO: implement check for deb
+	case p.policy.Deb.Install != nil:
+	case p.policy.Deb.Remove != nil:
+
+	case p.policy.GooGet.Install != nil:
+		if needsInstall(aptInstalled, p.policy.GooGet.Install.GetName()) {
```
manjunathganga, line 203: gooInstalled?

config/package_resouce.go
```diff
@@ -14,26 +14,229 @@
...
+			return true, nil
+		}
+	case p.policy.GooGet.Remove != nil:
+		if !needsInstall(aptInstalled, p.policy.GooGet.Remove.GetName()) {
```
manjunathganga, line 207: gooInstalled

config/package_resouce.go
```diff
@@ -14,26 +14,229 @@
...
+			return true, nil
+		}
+
+	// TODO: implement check for msi
+	case p.policy.MSI.Install != nil:
+	case p.policy.MSI.Remove != nil:
+
+	case p.policy.Yum.Install != nil:
+		if needsInstall(aptInstalled, p.policy.Yum.Install.GetName()) {
```
manjunathganga, line 216: yumInstalled? Similar comment for all other packages.

config/package_resouce.go
```diff
@@ -14,26 +14,229 @@
...
+			return nil, fmt.Errorf("cannot manage Yum package %q because yum does not exist on the system", pr.GetName())
+		}
+
+		switch p.GetDesiredState() {
+		case agentendpointpb.ApplyConfigTask_Config_Resource_PackageResource_INSTALLED:
+			p.policy.Yum.Install = pr
+		case agentendpointpb.ApplyConfigTask_Config_Resource_PackageResource_REMOVED:
+			p.policy.Yum.Remove = pr
+		}
+	case *agentendpointpb.ApplyConfigTask_Config_Resource_PackageResource_Zypper_:
```
manjunathganga, line 139: nit: Is the underscore at the end expected? Not seeing for other package managers.

config/package_resouce.go
```diff
@@ -14,26 +14,229 @@
...
+		}
+	default:
+		return nil, errors.New("SystemPackage field not set or references unknown package manager")
+	}
+
+	return &ManagedResources{Packages: []ManagedPackage{p.policy}}, nil
+}
+
+// TODO: implement a caching system for installed packages.
+var aptInstalled []packages.PkgInfo
```
manjunathganga, line 171: Would it be better to have a map here keyed by pkg name so that lookups will be faster. for example in needsInstall function.

config/package_resouce.go
adjackura: Done, also renamed function and reworked logic to make it clear.

config/package_resouce.go
adjackura: needsInstall is the function name, renamed to be clearer

config/package_resouce.go
adjackura: []packages.PkgInfo is what the inventory lookup call returns. Once I implement the cache system I may have that convert to a map though and remove the need for the intermediary function.

config/package_resouce.go
adjackura: It's an artifact of the autogenerated client library, has to do with the wrapped type and lowercase vs uppercase

config/package_resouce.go
adjackura: This is just the comment, and it is because of the os policy, just the package resource in that policy. The reason these are public is so we can run conflict resolution later on

config/package_resouce.go
adjackura: That's the type, which is public (will be used by agentendpoint package)

packages/msi.go
adjackura: Since its the name of the binary (msiexec.exe) i'm keeping it lowercase.

config/package_resouce.go
manjunathganga: Got it.
Should the comment be reworded as "AptPackage describes an apt package **_resource_**"
Same for all other comments for other package types.

config/package_resouce.go
manjunathganga: Sorry, I was referring to the argument passed to _needsInstall_ function. In this case, it should be _gooInstalled_, correct?

config/package_resouce.go
manjunathganga: Did you push the changes. Still seeing the old method name.

config/package_resouce.go
adjackura: ack, originally I was using this field for something slightly different thus the odd naming, changed

config/package_resouce.go
adjackura: That was a comment for the wrong function. Added in the named return

config/package_resouce.go
adjackura: oh lol, yeah nice catch, I will have full unit tests for this once it's fully implemented.

config/package_resouce.go
adjackura: done

manjunathganga (approved):

PR #232, opened 2020-09-21
Add CheckState/Validate for PackageResource

adjackura:


config/package_resouce.go
```diff
@@ -82,164 +92,264 @@ func (p *packageResouce) validate() (*ManagedResources, error) {
...
+		cache = &rpmInstalled
+		refreshFunc = packages.InstalledRPMPackages
+	default:
+		return fmt.Errorf("unknown or unpopulated ManagedPackage package type: %+v", mp)
+	}
+
+	// Cache already populated.
+	if *cache != nil {
+		return nil
+	}
```
manjunathganga, line 201: When the cache is already populated, when do we re-populate it?

If a package is installed outside the config management system, would the cache reflect that?

adjackura: I was going back and forth on how to implement this, for now i created a catch type with a refreshed timestamp

config/package_resource.go
```diff
@@ -195,8 +201,8 @@ func populateInstalledCache(ctx context.Context, mp ManagedPackage) error {
 		return fmt.Errorf("unknown or unpopulated ManagedPackage package type: %+v", mp)
 	}
 
-	// Cache already populated.
-	if *cache != nil {
+	// Cache already populated within the last 3 min.
+	if cache.cache != nil && cache.refreshed.After(time.Now().Add(-3*time.Minute)) {
```
manjunathganga, line 205: Where is this field getting updated?

adjackura: it wasnt, just fixed

manjunathganga (approved):

PR #233, opened 2020-09-24
Add RepositoryResource

adjackura:
Still need to add tests for APT GPG key setup, this should be a feature match for current GuestPolicies.

### DIFF
--- a/config/repository_resource.go
+++ b/config/repository_resource.go
@@ -260,7 +260,7 @@ func (r *repositoryResource) validate() (*ManagedResources, error) {
 		filePath = filepath.Join(agentconfig.YumRepoDir(), "osconfig_managed_%s.repo")
 
 	case *agentendpointpb.ApplyConfigTask_Config_Resource_RepositoryResource_Zypper:
-		if !packages.YumExists {
+		if !packages.ZypperExists {
 			return nil, errors.New("cannot manage yum repository because yum does not exist on the system")
 		}
 		r.managedRepository.Zypper = &ZypperRepository{RepositoryResource: r.GetZypper()}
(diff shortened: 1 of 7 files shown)


config/repository_resource.go
```diff
@@ -0,0 +1,321 @@
...
+	case *agentendpointpb.ApplyConfigTask_Config_Resource_RepositoryResource_Yum:
+		if !packages.YumExists {
+			return nil, errors.New("cannot manage yum repository because yum does not exist on the system")
+		}
+		r.managedRepository.Yum = &YumRepository{RepositoryResource: r.GetYum()}
+		r.managedRepository.RepoFileContents = yumRepoContents(r.GetYum())
+		filePath = filepath.Join(agentconfig.YumRepoDir(), "osconfig_managed_%s.repo")
+
+	case *agentendpointpb.ApplyConfigTask_Config_Resource_RepositoryResource_Zypper:
+		if !packages.YumExists {
```
manjunathganga, line 263: ZypperExists?

config/repository_resource.go
```diff
@@ -0,0 +1,321 @@
...
+}
+
+func (r *repositoryResource) checkState(ctx context.Context) (inDesiredState bool, err error) {
+	// Check APT gpg key if applicable.
+	if r.managedRepository.Apt != nil && r.managedRepository.Apt.GpgFileContents != nil {
+		match, err := contentsMatch(r.managedRepository.Apt.GpgFilePath, r.managedRepository.Apt.GpgChecksum)
+		if !match {
+			return false, nil
+		}
+		if err != nil {
```
manjunathganga, line 298: Should this error check happen before the check for match as an error also would return a mismatch?

config/repository_resource.go
adjackura: Fixed

config/repository_resource.go
adjackura: yep, fixed

config/repository_resource.go
```diff
@@ -260,7 +260,7 @@ func (r *repositoryResource) validate() (*ManagedResources, error) {
 		filePath = filepath.Join(agentconfig.YumRepoDir(), "osconfig_managed_%s.repo")
 
 	case *agentendpointpb.ApplyConfigTask_Config_Resource_RepositoryResource_Zypper:
-		if !packages.YumExists {
+		if !packages.ZypperExists {
 			return nil, errors.New("cannot manage yum repository because yum does not exist on the system")
```
manjunathganga, line 264: The error message also needs to be updated :-)

manjunathganga (approved):

PR #234, opened 2020-09-25
Set full version during Windows build

adjackura:


packaging/googet/google-osconfig-agent.goospec
```diff
@@ -37,6 +29,6 @@
   }],
   "build": {
     "linux": "/bin/bash",
-    "linuxArgs": ["-c", "GOOS=windows $GO build -ldflags='-s -w -X main.version={{.version}}' -mod=readonly -o google_osconfig_agent.exe"]
+    "linuxArgs": ["-c", "GOOS=windows $GO build -ldflags='-s -w -X main.version={{$package_version}}' -mod=readonly -o google_osconfig_agent.exe"]
```
jaiminsh, line 32: Should this be version as "version": "{{$package_version}}" ?

adjackura: quoting here is complicated as this has two levels of nesting, this works as is ($package_version does not have spaces, that would cause a build failure) so I'd prefer not to change it (bash quoting can get complicated)

PR #236, opened 2020-10-07
Report installed packages on COS

edjee:
Passed gofmt, goimports, goline, go vet, go build, go test.
The resulted binary worked as expected on a COS instance.


packages/packages.go
```diff
@@ -64,6 +66,7 @@ type Packages struct {
 	Deb           []PkgInfo     `json:"deb,omitempty"`
 	Zypper        []PkgInfo     `json:"zypper,omitempty"`
 	ZypperPatches []ZypperPatch `json:"zypperPatches,omitempty"`
+	Cos           []PkgInfo     `json:"cos,omitempty"`
```
adjackura, line 69: We aren't super diligent about this, but I assume you may prefer COS over Cos

packages/packages_linux.go
```diff
@@ -127,6 +127,9 @@ func GetInstalledPackages(ctx context.Context) (Packages, error) {
 			pkgs.Deb = deb
 		}
 	}
+	if cos, err := InstalledCOSPackages(); err == nil {
```
adjackura, line 130: Your missing the check for COSPkgInfoExists

adjackura (changes requested):

packages/packages.go
edjee: Fixed in the new commit.

packages/packages_linux.go
edjee: Fixed in the new commit.

By the way, the previous commit I made seems lost. How can I preserve the patch history, like keeping the past "patch sets" for a specific CL in Gerrit?

packages/cos.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"fmt"
+	"strconv"
+
+	"cos.googlesource.com/cos/tools.git/src/pkg/cos"
+	"github.com/GoogleCloudPlatform/osconfig/osinfo"
+)
+
+// COSPackageInfoExists returns whether COS package information exists.
+func COSPackageInfoExists() bool {
```
adjackura, line 26: Leave this as a var in packages.go and populated it in an init() function in this file. That way this check is only done once not every run, and it's a bit simpler to test.

edjee: Done in the second commit.

adjackura (approved):

PR #237, opened 2020-10-07
add initial e2e tests for inventory reporting

wj-chen:
Created a new test suite for inventory reporting. Test case does basic validations on the corresponding RPC invocation and its payload. Test setup was branched from the inventory test suite.

The WaitForSerialOutput verification function was referenced from https://github.com/GoogleCloudPlatform/osconfig/commit/9a3f7f3c83544dba6fbf4ecb0b6f1be6ad2dc693#diff-c54902af2d84c9ff1ed5525e24a7e3f3R52 and modified to support a list of regular expressions as one of the inputs.

/hold Waiting for the backend RPC visibility issue to be fixed.
/assign @adjackura @ryanwe 
/woof

### DIFF
--- a/e2e_tests/test_suites/inventoryreporting/inventory_reporting.go
+++ b/e2e_tests/test_suites/inventoryreporting/inventory_reporting.go
@@ -1,0 +1,167 @@
+//  Copyright 2020 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package inventoryreporting
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"path"
+	"regexp"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/GoogleCloudPlatform/compute-image-tools/go/e2e_test_utils/junitxml"
+	"github.com/GoogleCloudPlatform/osconfig/e2e_tests/compute"
+	"github.com/GoogleCloudPlatform/osconfig/e2e_tests/config"
+	gcpclients "github.com/GoogleCloudPlatform/osconfig/e2e_tests/gcp_clients"
+	testconfig "github.com/GoogleCloudPlatform/osconfig/e2e_tests/test_config"
+	"github.com/GoogleCloudPlatform/osconfig/e2e_tests/utils"
+	api "google.golang.org/api/compute/v1"
+)
+
+const (
+	testSuiteName = "OSInventoryReporting"
+)
+
+// TestSuite is an OSInventoryReporting test suite.
+func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junitxml.TestSuite, logger *log.Logger, testSuiteRegex, testCaseRegex *regexp.Regexp) {
+	defer tswg.Done()
+
+	if testSuiteRegex != nil && !testSuiteRegex.MatchString(testSuiteName) {
+		return
+	}
+
+	testSuite := junitxml.NewTestSuite(testSuiteName)
+	defer testSuite.Finish(testSuites)
+
+	logger.Printf("Running TestSuite %q", testSuite.Name)
+
+	var wg sync.WaitGroup
+	tests := make(chan *junitxml.TestCase)
+	for _, setup := range headImageTestSetup() {
+		wg.Add(1)
+		go inventoryReportingTestCase(ctx, setup, tests, &wg, logger, testCaseRegex)
+	}
+
+	go func() {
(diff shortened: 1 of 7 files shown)


e2e_tests/test_suites/inventoryreporting/inventory_reporting.go
```diff
@@ -0,0 +1,152 @@
+//  Copyright 2018 Google Inc. All Rights Reserved.
```
ryanwe, line 1: 2020

e2e_tests/test_suites/inventoryreporting/inventory_reporting.go
```diff
@@ -0,0 +1,152 @@
...
+func runInventoryReportingTest(ctx context.Context, testSetup *inventoryTestSetup, testCase *junitxml.TestCase, logwg *sync.WaitGroup) {
+	testCase.Logf("Creating compute client")
+
+	computeClient, err := gcpclients.GetComputeClient()
+	if err != nil {
+		testCase.WriteFailure("Error getting compute client: %v", err)
+		return
+	}
+
+	testSetup.hostname = fmt.Sprintf("inventory-test-%s-%s", path.Base(testSetup.testName), utils.RandString(5))
```
ryanwe, line 81: would "inventoryreporting-test-%s-%s" be too long? (to differentiate from other inventory tests)

e2e_tests/test_suites/inventoryreporting/inventory_reporting.go
```diff
@@ -0,0 +1,152 @@
...
+		testCase.WriteFailure("Error waiting for osconfig agent install: %v", err)
+		return
+	}
+
+	// Build regexes for verification.
+	patterns := []string{
+		`.*Calling ReportInventory with request.*`,
+		fmt.Sprintf(`.*"hostname":.*"%s".*`, testSetup.hostname),
+		fmt.Sprintf(`.*"short_name":.*"%s".*`, testSetup.shortName),
+	}
```
ryanwe, line 113: Can there also be a pattern check that the call was successful? Or if this does that, can you add a comment? When I see a log message like "Calling ..." I think it's going to happen but hasn't occurred yet.

e2e_tests/test_suites/inventoryreporting/test_setup.go
```diff
@@ -0,0 +1,131 @@
+//  Copyright 2019 Google Inc. All Rights Reserved.
```
ryanwe, line 1: 2020

ryanwe (commented):
LGTM. But I'll defer to Andrew.

e2e_tests/test_suites/inventoryreporting/inventory_reporting.go
```diff
@@ -0,0 +1,152 @@
...
+	// Skip packages test for cos as it is not currently supported.
+	if !strings.Contains(testCase.Name, "cos") {
```
ryanwe, line 115: I don't think you need this filter anymore: https://github.com/GoogleCloudPlatform/osconfig/pull/236

e2e_tests/test_suites/inventoryreporting/inventory_reporting.go
```diff
@@ -0,0 +1,152 @@
...
+		testCase.WriteFailure("Error getting compute client: %v", err)
+		return
+	}
+
+	testSetup.hostname = fmt.Sprintf("inventory-test-%s-%s", path.Base(testSetup.testName), utils.RandString(5))
+
+	var metadataItems []*api.MetadataItems
+	metadataItems = append(metadataItems, testSetup.startup)
+	metadataItems = append(metadataItems, compute.BuildInstanceMetadataItem("enable-os-inventory", "true"))
+	metadataItems = append(metadataItems, compute.BuildInstanceMetadataItem("osconfig-disabled-features", "tasks,guestpolicies"))
```
adjackura, line 86: You should only need enable-os-inventory, or you can just enable all of osconfig and then specifically disable tasks and guest policies if you want

e2e_tests/test_suites/inventoryreporting/inventory_reporting.go
adjackura: The debug log should also record the response, so we should also just check for that.

e2e_tests/test_suites/inventoryreporting/inventory_reporting.go
wj-chen: In practice, without `osconfig-disabled-features` the agent on the test VMs would attempt to listen for patch task and run the guest policies task (which could take some time). I'll share some screenshots with you offline.

e2e_tests/test_suites/inventoryreporting/inventory_reporting.go
wj-chen: Just checked and we are not debug-logging the response for report inventory yet, but i can add that in a separate PR. For the time being I added another regex for "Finished task OSInventory" to match the console output with and a negative lookup for "Error reporting inventory" to fail on. Both changes have been tested and should be sufficient until I add in the logging for response and update the tests again once new version is released to staging and unstable.

e2e_tests/test_suites/inventoryreporting/inventory_reporting.go
adjackura: We must have it enabled at the project then, not that big an issue.

adjackura (approved):

ryanwe (approved):
/lgtm

PR #239, opened 2020-10-12
Make inventory WUAUpdates call spawn a new process, retry on metadata unmarshal error

### LINKED ISSUE #154: OSConfig Agent "unexpected end of JSON input"
I have a VM instance running Debian GNU/Linux 9 (stretch), it has the OSConfig Agent installed and running.  I have encountered the following error several times. 

```
 {
  "insertId": "ezj5okfzvhagf",
  "jsonPayload": {
    "localTimestamp": "2020-03-03T18:37:28.4900Z",
    "message": "unexpected end of JSON input"
  },
  "resource": {
    "type": "gce_instance",
    "labels": {
      "instance_id": "<INSTANCE_ID>",
      "zone": "us-central1-a",
      "project_id": "<PROJECT_ID>"
    }
  },
  "timestamp": "2020-03-03T18:37:28.737755265Z",
  "severity": "ERROR",
  "logName": "projects/<PROJECT_ID>/logs/OSConfigAgent",
  "sourceLocation": {
    "file": "main.go",
    "line": "154",
    "function": "main.runLoop"
  },
  "receiveTimestamp": "2020-03-03T18:37:30.853440003Z"
}
```
I have installed nodejs 12.16.1.

adjackura:
We have issues with the WUA api leaking memory, this is an issue with the inventory calls because they run so often.  To solve this the list calls for inventory spawn a new instance of the agent to query the api, this is only implemented for inventory calls as patch calls do not run that often before an agent restart and the miniscule memory buildup per run will never be noticed.

Make sure we close clients when done with them.

The metadata unmarshal sometimes returns an error, this most often will be because of a bad send from the metadata server and can just be retried safely to avoid logging this error. (fixes #154)

### DIFF
--- a/main.go
+++ b/main.go
@@ -258,6 +258,15 @@ func main() {
 	}
 
 	switch action := flag.Arg(0); action {
+	// wuaupdates just runs the packages.WUAUpdates function and returns it's output
+	// as JSON on stdout. This avoids memmory issues with the WUA api since this is
+	// called often for Windows inventory runs.
+	case "wuaupdates":
+		if err := wuaUpdates(flag.Arg(1)); err != nil {
+			fmt.Fprint(os.Stderr, err)
+			os.Exit(1)
+		}
+		os.Exit(0)
 	case "", "run":
 		runService(ctx)
 	default:
--- a/main.go
+++ b/main.go
@@ -226,6 +226,7 @@ func runServiceLoop(ctx context.Context) {
 					logger.Errorf(err.Error())
 				}
 				client.ReportInventory(ctx)
+				client.Close()
 			})
 		}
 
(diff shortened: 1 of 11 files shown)


main.go
```diff
@@ -258,6 +259,15 @@ func main() {
 	}
 
 	switch action := flag.Arg(0); action {
+	// wuaupdates just runs the packages.WUAUpdates function and returns it's output
+	// as JSON on stdout. This avoids memmory issues with the WUA api since this is
```
wj-chen, line 263: "memory"

packages/packages_windows.go
```diff
@@ -15,17 +15,40 @@ package packages
...
 	"fmt"
+	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/GoogleCloudPlatform/osconfig/clog"
 	"github.com/GoogleCloudPlatform/osconfig/util"
 )
 
+// In order to work around memmory issues with the WUA library we spawn a
```
wj-chen, line 29: "memory"

wj-chen (commented):
LGTM

PR #241, opened 2020-10-16
Add sleep after Unavailable errors for agentendpoint

### LINKED ISSUE #228: High CPU usage on restricted environment
We experience a very high CPU usage from the google_osconfig_agent proces. More than 60% average.

Our environment is restricted in a few ways:
- We're using shared VPC networks
- Access to Google API's are redirected to restricted.googleapis.com
- Outbound internet is filtered using a proxy server. All outbound traffic is denied except for a whitelist (see below)
- no_proxy is set to ``localhost,127.0.0.1,10.0.0.0/8,169.254.169.254,metadata.google.internal,.googleapis.com`` to prevent traffic go through the proxy for the APIs.

I started the agent manually with:
``google_osconfig_agent -debug -stdout ospatch``

Then it shows the following debug log:
```
2020-09-14T11:10:21.5994Z OSConfigAgent Info: OSConfig Agent (version 20200901.01-g1.el7) started.
2020-09-14T11:10:21.6005Z OSConfigAgent Debug: Creating new agentendpoint client.
2020-09-14T11:10:21.6007Z OSConfigAgent Debug: Running WaitForTaskNotification
2020-09-14T11:10:21.6008Z OSConfigAgent Debug: Checking local state file for saved task.
2020-09-14T11:10:21.6009Z OSConfigAgent Debug: Setting up ReceiveTaskNotification stream watcher.
2020-09-14T11:10:21.6011Z OSConfigAgent Debug: Creating new agentendpoint client.
2020-09-14T11:10:21.6110Z OSConfigAgent Debug: Calling ReceiveTaskNotification with request:
{
  "instance_id_token":  "",
  "agent_version":  "20200901.01-g1.el7"
}
2020-09-14T11:10:21.6113Z OSConfigAgent Debug: Calling RegisterAgent with request:
{
  "instance_id_token":  "",
  "agent_version":  "20200901.01-g1.el7",
  "supported_capabilities":  [
    "PATCH_GA",
    "GUEST_POLICY_BETA"
  ]
}
```

But then keeps looping every few milliseconds:
```
2020-09-14T11:15:39.8288Z OSConfigAgent Debug: Waiting on ReceiveTaskNotification stream Recv().
2020-09-14T11:15:39.8292Z OSConfigAgent Debug: Stream canceled, will reconnect: rpc error: code = Unavailable desc = 404:Not Found
2020-09-14T11:15:39.8293Z OSConfigAgent Debug: Calling ReceiveTaskNotification with request:
{
  "instance_id_token":  "",
  "agent_version":  "20200901.01-g1.el7"
}
2020-09-14T11:15:39.8294Z OSConfigAgent Debug: Waiting on ReceiveTaskNotification stream Recv().
2020-09-14T11:15:39.8299Z OSConfigAgent Debug: Stream canceled, will reconnect: rpc error: code = Unavailable desc = 404:Not Found
2020-09-14T11:15:39.8299Z OSConfigAgent Debug: Calling ReceiveTaskNotification with request:
{
  "instance_id_token":  "",
  "agent_version":  "20200901.01-g1.el7"
}
```

This causes a very high CPU usage on our VM's.

It feels like the agent is not able to reach some HTTP(S) url, but I can't pinpoint what I need to unblock to get the agent working.

adjackura:
Fixes #228

### DIFF
--- a/agentendpoint/agentendpoint.go
+++ b/agentendpoint/agentendpoint.go
@@ -46,12 +46,11 @@ import (
 const apiRetrySec = 600
 
 var (
-	errServerCancel       = errors.New("task canceled by server")
-	errServiceNotEnabled  = errors.New("service is not enabled for this project")
-	errServiceUnavailable = errors.New("Unavailable")
-	errResourceExhausted  = errors.New("ResourceExhausted")
-	taskStateFile         = agentconfig.TaskStateFile()
-	sameStateTimeWindow   = -5 * time.Second
+	errServerCancel      = errors.New("task canceled by server")
+	errServiceNotEnabled = errors.New("service is not enabled for this project")
+	errResourceExhausted = errors.New("ResourceExhausted")
+	taskStateFile        = agentconfig.TaskStateFile()
+	sameStateTimeWindow  = -5 * time.Second
 )
 
 // Client is a an agentendpoint client.
@@ -316,9 +315,6 @@ func (c *Client) waitForTask(ctx context.Context) error {
 	}
 	if s, ok := status.FromError(err); ok {
 		switch s.Code() {
-		case codes.Unavailable:
-			clog.Warningf(ctx, "Stream connection error, will reconnect: %v", err)
-			return errServiceUnavailable
 		case codes.PermissionDenied:
 			// Service is not enabled for this project.
 			return errServiceNotEnabled
@@ -326,7 +322,6 @@ func (c *Client) waitForTask(ctx context.Context) error {
 			return errResourceExhausted
 		}
 	}
-	// TODO: Add more error checking (handle more API erros vs non API errors) and backoff where appropriate.
 	return err
 }
 
--- a/agentendpoint/agentendpoint.go
+++ b/agentendpoint/agentendpoint.go
@@ -46,11 +46,12 @@ import (
 const apiRetrySec = 600
 
 var (
-	errServerCancel      = errors.New("task canceled by server")
-	errServiceNotEnabled = errors.New("service is not enabled for this project")
-	errResourceExhausted = errors.New("ResourceExhausted")
-	taskStateFile        = agentconfig.TaskStateFile()
-	sameStateTimeWindow  = -5 * time.Second
+	errServerCancel       = errors.New("task canceled by server")
+	errServiceNotEnabled  = errors.New("service is not enabled for this project")
+	errServiceUnavailable = errors.New("Unavailable")
+	errResourceExhausted  = errors.New("ResourceExhausted")
+	taskStateFile         = agentconfig.TaskStateFile()
+	sameStateTimeWindow   = -5 * time.Second
 )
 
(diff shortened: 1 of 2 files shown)


agentendpoint/agentendpoint.go
```diff
@@ -373,16 +370,27 @@ func (c *Client) WaitForTaskNotification(ctx context.Context) {
...
+					sleep = retryutil.RetrySleep(resourceExhausted, 5)
 				} else {
-					resourceExhausted = 1
+					// Retry any other errors with a modest backoff. Only retry up to 10
+					// times, at that point return, the client will be recreated during the next
+					// cycle.
+					errs++
+					clog.Warningf(ctx, "Error waiting for task (attempt %d of 10): %v", errs, err)
+					resourceExhausted = 0
+					if errs > 10 {
```
ryanwe, line 384: Would it be easier to reason about failures if this used a time-based threshold instead, for example, giving up after 2 minutes (I think this gives up after around 1 minute but can vary a lot depending on the random jitter of the retry delay).

agentendpoint/agentendpoint.go
```diff
@@ -373,16 +370,27 @@ func (c *Client) WaitForTaskNotification(ctx context.Context) {
...
+						c.Close()
+						return
+					}
+					sleep = retryutil.RetrySleep(errs, 0)
 				}
 				time.Sleep(sleep)
+				continue
 			}
+			errs = 0
```
ryanwe, line 393: Just an observation: it seems a little odd that this function is designed to wait forever for a task when things work, retry a few times to overcome intermittent failures, but return and let the caller retry for other failures. This manifests itself as having a few variables that track different errors and get reset in different scenarios. Would it make sense to consolidate the API retry logic for both short-term and long-term failures in one place - possibly separating retry logic for different types of errors?

ryanwe (approved):
/lgtm
/hold

agentendpoint/agentendpoint.go
adjackura: We just want to retry enough for very temporary issues to resolve themselves. Timing is very complicated as we also have the request timeouts to deal with. Time based might be easier to read if all we cared about was time since first failure, but would be more complicated to implement and ensure we try at least x number of times, so I'd end up implementing both time based on retry based. Since min retry time for 10 errors is over 6min that felt like a good point to just quit and wait for the next reload (max 10mins)

agentendpoint/agentendpoint.go
adjackura: The current logic is more "continue forever unless there is a fatal error" all the errors we stop on are thought of as fatal and continuing to retry connections not super useful. The fact that main will recreate the connection is not a retry, more how the connection logic is implemented. 

I could rewrite this to never return and these "fatal" errors just force us into a longer retry loop. That could make this function make more sense (only way to stop it is to call Close() or cancel the context). It wouldn't change any logic elsewhere though as all that is till required for other situations.

PR #242, opened 2020-10-16
Ensure we record epoch for rpm packages

adjackura:


manjunathganga (commented):
Looks good.

Can we add tests?

adjackura:
We can't really add any unit tests for this as all that's changed is the output that rpmquery gives us, and that's not unit testable. We don't currently have a way for a more in depth integration test where we actually call external commands and check for expected output other than our e2e tests and those don't have absolute of inventory cases at this time.

manjunathganga (approved):

PR #246, opened 2020-10-20
Support deb and rpm in PackageResource

adjackura:
Merged latest two commits from master as one was related to the code I was changing.

### DIFF
--- a/config/package_resource.go
+++ b/config/package_resource.go
@@ -18,11 +18,15 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
 	"github.com/GoogleCloudPlatform/osconfig/clog"
 	"github.com/GoogleCloudPlatform/osconfig/packages"
+	"github.com/GoogleCloudPlatform/osconfig/util"
 
 	agentendpointpb "github.com/GoogleCloudPlatform/osconfig/internal/google.golang.org/genproto/googleapis/cloud/osconfig/agentendpoint/v1alpha1"
 )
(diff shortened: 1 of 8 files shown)


adjackura:
This PR only updates files in config/ and packages/, the other updates are from upstream changes.

config/package_resource.go
```diff
@@ -101,8 +116,16 @@ func (p *packageResouce) validate(ctx context.Context) (*ManagedResources, error
 		if !packages.DpkgExists {
 			return nil, fmt.Errorf("cannot manage Deb package because dpkg does not exist on the system")
 		}
+		switch p.GetDesiredState() {
+		case agentendpointpb.ApplyConfigTask_OSPolicy_Resource_PackageResource_INSTALLED:
```
manjunathganga, line 120: Why is **Installed** not an applicable desired state for deb?
Similar comment for other package types in this file

adjackura: Installed is the only applicable desired state, any other state is not applicable. 
I reorganized the statement to be much more readable.

config/package_resource.go
```diff
@@ -101,8 +116,14 @@ func (p *packageResouce) validate(ctx context.Context) (*ManagedResources, error
 		if !packages.DpkgExists {
 			return nil, fmt.Errorf("cannot manage Deb package because dpkg does not exist on the system")
 		}
+		if p.GetDesiredState() != agentendpointpb.ApplyConfigTask_OSPolicy_Resource_PackageResource_INSTALLED {
```
manjunathganga, line 119: Don't we support other desired states(example: REMOVED) for Deb and other package resources that are checking only for Installed state here? 

Do we plan to add support for other desired states in the future PRs?

adjackura: Currently I don't plan on supporting anything but INSTALLED as it really doesnt make much sense, we should also add this check in the API. If you want to remove a deb package use the apt functionality, same with rpm and yum/zypper.

manjunathganga: Got it. Can you file a bug to update our API documentation and also to add this validation on the service side?

Looking through the cases here, I see that only APT and Googet  allows for REMOVED state. Checking to see if that is what you see as the final support for these package types or if you want to add REMOVED support for more package types in the future. If it is the latter, adding TODOs might help to differentiate the cases.

Also, I don't see a case statement for RPM.

adjackura: Created a bug to track this on the API side. apt, googet, yum and zypper all allow install and remove. deb, rpm and msi only allow install, those three should have a simple if check in validation.

manjunathganga (approved):

PR #253, opened 2020-11-05
use retryutil for ReportInventory call

wj-chen:
The logic tracks an additional number of "attempts" for reporting full inventory to prevent going into an infinite loop in case service continuously keeps asking for full inventory. Otherwise, each ReportInventory call (with full inventory or not) gets retried according to the retryutil specifications.

/hold Tested manually but still need to run e2e tests
/assign @adjackura @jaiminsh 
/woof

### DIFF
--- a/agentendpoint/inventory.go
+++ b/agentendpoint/inventory.go
@@ -12,6 +12,7 @@ import (
 	"github.com/GoogleCloudPlatform/osconfig/clog"
 	"github.com/GoogleCloudPlatform/osconfig/inventory"
 	"github.com/GoogleCloudPlatform/osconfig/packages"
+	"github.com/GoogleCloudPlatform/osconfig/retryutil"
 	"github.com/GoogleCloudPlatform/osconfig/util"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
@@ -20,7 +21,6 @@ import (
 
 const (
 	inventoryURL = agentconfig.ReportURL + "/guestInventory"
-	maxRetries   = 5
 )
 
 // ReportInventory writes inventory to guest attributes and reports it to agent endpoint.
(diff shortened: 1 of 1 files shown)


agentendpoint/inventory.go
```diff
@@ -61,23 +62,28 @@ func (c *Client) report(ctx context.Context, state *inventory.InstanceInventory)
...
+			res, err = c.reportInventory(ctx, inventory, reportFull)
+			if err != nil {
+				return err
+			}
+			clog.Debugf(ctx, "ReportInventory response:\n%s", util.PrettyFmt(res))
+			return nil
+		}); err != nil {
 			clog.Errorf(ctx, "Error reporting inventory: %v", err)
+			break
+		} else if attempts > maxRetries {
```
jaiminsh, line 80: I am curious, why do we have a logic of reporting inventory multiple times. Shouldn't the logic be 
1. Call ReportInventory with reportFull = false
2. If the ReportFullInventory is true, call ReportInventory with reportFull = true.

In what case would the agent be asked to report inventory again? Wouldn't that mean a bug in the service?

wj-chen: Each response comes back with a reportFullInventory flag which might still be `true` even if full inventory was just sent, most likely due to unforeseen issues in either agent not sending or the service not receiving inventory as expected.

We could disregard the flag in the second request and try again next time inventory task runs on the agent if we are ok with that behavior.

jaiminsh: I would suggest to change the behavior with what I suggested in the earlier comment instead of blindly retrying. We dont know how retries would solve the unforeseen issues in either the agent or service except from rpc retrying at the request level for the different codes we receive from the service, which we have in place.

wj-chen: Simplified!

agentendpoint/inventory.go
```diff
@@ -20,7 +21,7 @@ import (
 
 const (
 	inventoryURL = agentconfig.ReportURL + "/guestInventory"
-	maxRetries   = 5
+	maxRetries   = 3
```
jaiminsh, line 24: No need for this now :)

wj-chen: Ah yes :)

adjackura (approved):

PR #259, opened 2020-11-16
Update test Windows images

adjackura:
Remove EOL 1809 image
Add 2004 image
Remove semiannual images from HEAD tests as they don't add much.


ryanwe (approved):
/lgtm

PR #265, opened 2020-12-03
fix: UpdateSession object not release.

dean2021:
fix: UpdateSession object not release.


adjackura:
Pretty sure this will clear the memory prematurely and WUA calls will fail.

adjackura:
Closing as this should have been resolved a log while ago.

PR #268, opened 2020-12-10
Add ExecResource

adjackura:
Add ExecResource
Move downloading of data from checkState to validate in both PackageResouce and FileResource as checkState gets run multiple times and we don't want to keep downloading.


config/exec_resource.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		if v, ok := err.(*exec.ExitError); ok {
+			code = v.ExitCode()
+		}
+	}
+	return stdout, stderr, code, err
+}
+
+func (e *execResource) checkState(ctx context.Context) (inDesiredState bool, err error) {
+	// For validate we expect an exit code of 100 for "correct state" and 101 for "incorrect state".
```
manjunathganga, line 152: Is this captured in our API documentation? If not, can you add it?

config/exec_resource.go
```diff
@@ -0,0 +1,184 @@
...
+	stdout, stderr, code, err := e.run(ctx, e.validatePath, e.GetValidate())
+	switch code {
+	case -1:
```
manjunathganga, line 155: Is this a code that the customer returns or is this what the OS returns?

config/exec_resource.go
```diff
@@ -0,0 +1,184 @@
...
+		return false, err
+	case 100:
```
manjunathganga, line 157: Can we add a comment here as to why we did not choose 0 for this case since it might be a natural question that will arise when reading this code in the future.

config/exec_resource.go
```diff
@@ -0,0 +1,184 @@
...
+func (e *execResource) enforceState(ctx context.Context) (inDesiredState bool, err error) {
+	// For enforce we expect an exit code of 100 for "success" and anything positive code is a failure".
+	stdout, stderr, code, err := e.run(ctx, e.enforcePath, e.GetEnforce())
+	switch code {
+	case -1:
+		return false, err
+	case 100:
+		return true, nil
+	default:
+		return false, fmt.Errorf("unexpected return code from validate: %d, stdout: %s, stderr: %s", code, stdout, stderr)
```
manjunathganga, line 175: typo: enforceState

config/exec_resource.go
adjackura: Will do, I also need to update the API a bit for this resource as we don't use the allowed exit codes.

config/exec_resource.go
adjackura: The way this logic works is if e.run returns code -1 that means it was an error running exec (like file not found or something) so we just log the error. this is a workaround for checking err != nil as we can't do that directly because all non 0 return codes return an exec error.

manjunathganga (approved):

PR #269, opened 2020-12-11
Drop Windows 1903 and CentOS 6 from tests

adjackura:
Both are EOL, CentOS 6 repos do not exist and so no yum calls work


ryanwe (approved):
/lgtm

PR #272, opened 2020-12-29
Carry over patch task labels during reboot, minor updates to tests, additional debug logging, allow KB prefix to excludes

adjackura:


ryanwe (approved):
/lgtm

PR #283, opened 2021-01-27
Make writes atomic, add unused "allowDowngrades" option to apt, fix a few recipe issues

adjackura:


policies/recipes/steps.go
```diff
@@ -494,10 +505,10 @@ func writeScript(path, contents string) error {
 		f.Close()
 		return err
 	}
-	if err := f.Close(); err != nil {
+	if err := f.Chmod(0755); err != nil {
 		return err
 	}
-	if err := os.Chmod(path, 0755); err != nil {
+	if err := f.Close(); err != nil {
```
ryanwe, line 511: Should this use AtomicWrite too?

policies/recipes/steps.go
```diff
@@ -80,18 +83,26 @@ func stepCopyFile(step *agentendpointpb.SoftwareRecipe_Step_CopyFile, artifacts
...
 	}
 
-	return nil
+	if err := tmp.Chmod(permissions); err != nil {
+		return err
+	}
+
+	if err := tmp.Close(); err != nil {
+		return err
+	}
```
ryanwe, line 103: Would an AtomicFileCopy be helpful? The only difference is io.Copy vs. io.Write.

adjackura: Not needed by any other function right now, will add if I find other uses

ryanwe (approved):
/lgtm

PR #291, opened 2021-03-15
Fix ExecResource for config task, add minimal unit test

adjackura:
More unit tests still need to be added for the rest of the ExecResource


manjunathganga (commented):
/lgtm

PR #294, opened 2021-03-15
Fix ExecResource permissions, add logs to fetcher

adjackura:


config/exec_resource.go
```diff
@@ -55,6 +56,7 @@ func (e *execResource) download(ctx context.Context, execR *agentendpointpb.OSPo
 			} else {
 				name = "script"
 			}
+			perms = os.FileMode(0755)
```
ryanwe, line 59: Should `OSPolicy_Resource_ExecResource_Exec_SHELL` set this too?

adjackura: No, fir SHELL we don't execute this file

ryanwe (approved):

PR #296, opened 2021-03-18
config_task: Rework config step logic

adjackura:
Added additional debug log for printing each policy before execution
ConfigSteps will only contain the steps that ran for each resource

Logic change:

For each policy in policies:
- For each resource in this policies resources
  - If management of a previous resource ended in failure, skip managing this resource
  - Validate, setup, and check desired state
  - Enforce desired state if required
  - If enforcement was required record that post check state is required

If post check was triggered, check the state for every resource in each policy.

### DIFF
--- a/agentendpoint/config_task.go
+++ b/agentendpoint/config_task.go
@@ -175,7 +175,7 @@ func checkConfigResourceState(ctx context.Context, plcy *policy, rCompliance *ag
 	rCompliance.State = state
 }
 
-func enforceConfigResourceState(ctx context.Context, plcy *policy, rCompliance *agentendpointpb.OSPolicyResourceCompliance, configResource *agentendpointpb.OSPolicy_Resource) bool {
+func enforceConfigResourceState(ctx context.Context, plcy *policy, rCompliance *agentendpointpb.OSPolicyResourceCompliance, configResource *agentendpointpb.OSPolicy_Resource) (enforcementActionTaken bool) {
 	ctx = clog.WithLabels(ctx, map[string]string{"resource_id": configResource.GetId()})
 	res := plcy.resources[configResource.GetId()]
 	// Only enforce resources that need it.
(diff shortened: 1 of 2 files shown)


agentendpoint/config_task.go
```diff
@@ -402,14 +295,41 @@ func (c *configTask) run(ctx context.Context) error {
...
-	c.checkState(ctx)
 
 	if err := c.reportContinuingState(ctx, agentendpointpb.ApplyConfigTaskProgress_APPLYING_CONFIG); err != nil {
 		return c.handleErrorState(ctx, rcsErrMsg, err)
 	}
-	c.enforceState(ctx)
+
+	c.policies = map[string]*policy{}
+	for i, osPolicy := range c.Task.GetOsPolicies() {
+		ctx = clog.WithLabels(ctx, map[string]string{"config_assignment": osPolicy.GetOsPolicyAssignment(), "policy_id": osPolicy.GetId()})
```
manjunathganga, line 306: os_policy_assignment

manjunathganga, line 306: os_policy_id

agentendpoint/config_task.go
```diff
@@ -402,14 +295,41 @@ func (c *configTask) run(ctx context.Context) error {
...
+			rCompliance := pResult.GetOsPolicyResourceCompliances()[i]
+			validateConfigResource(ctx, plcy, policyMR, rCompliance, configResource)
+			if plcy.hasError {
+				break
+			}
+			checkConfigResourceState(ctx, plcy, rCompliance, configResource)
+			if plcy.hasError {
+				break
+			}
+			if enforceConfigResourceState(ctx, plcy, rCompliance, configResource) {
```
manjunathganga, line 324: Consider naming the return value for this method as it could be potentially confusing while reading to determine if it is indicating whether enforcement was attempted or whether enforcement was successful

adjackura: Done here and in the function

PR #302, opened 2021-04-01
e2e_tests: add OSPolicies tests for package install

adjackura:
Should only run for unstable tests currently
All code under internal/ is autogenerated code


manjunathganga (commented):
/lgtm

PR #306, opened 2021-04-02
e2e_tests: Add more OSPolicies package and repo tests

adjackura:
Added package removal OSPolices test
Added package install from new repo OSPolices test
Added sles OSPolices tests
Update tests names for both GuestPolices and OSPolices


manjunathganga (commented):
/lgtm

PR #307, opened 2021-04-05
Disable OSPolicies by default since it is an unreleased feature

adjackura:


agentendpoint/agentendpoint.go
```diff
@@ -234,8 +235,13 @@ func (c *Client) runTask(ctx context.Context) {
 				clog.Errorf(ctx, "Error running TaskType_EXEC_STEP_TASK: %v", err)
 			}
 		case agentendpointpb.TaskType_APPLY_CONFIG_TASK:
-			if err := c.RunApplyConfig(ctx, task); err != nil {
-				clog.Errorf(ctx, "Error running TaskType_APPLY_CONFIG_TASK: %v", err)
+			if agentconfig.OSPoliciesEnabled() || osPoliciesEnabled {
+				if err := c.RunApplyConfig(ctx, task); err != nil {
+					clog.Errorf(ctx, "Error running TaskType_APPLY_CONFIG_TASK: %v", err)
```
ryanwe, line 240: Will users understand this message? I know we don't expect to see it often, but would "Error applying OS policy configurations: %v" be more helpful?

ryanwe (approved):

PR #308, opened 2021-04-05
Fix e2e_tests

adjackura:
Fix inventory poll interval
Wait for policy assignment rollout to complete
Delete old inventory metadata report


ryanwe (approved):
/lgtm

PR #317, opened 2021-04-08
Config task: only run post check on resources that have already been evaluated

adjackura:
When any resource is enforced mark all previous resources (for all policies) as needing post check.
With this change any resource that is evaluated AFTER the last resource that was "enforced" the agent will not re check.

### DIFF
--- a/agentendpoint/config_task.go
+++ b/agentendpoint/config_task.go
@@ -26,8 +26,8 @@ import (
 	agentendpointpb "google.golang.org/genproto/googleapis/cloud/osconfig/agentendpoint/v1"
 )
 
-var newResource = func(r *agentendpointpb.OSPolicy_Resource) resourceIface {
-	return resourceIface(&config.OSPolicyResource{OSPolicy_Resource: r})
+var newResource = func(r *agentendpointpb.OSPolicy_Resource) *resource {
+	return &resource{resourceIface: resourceIface(&config.OSPolicyResource{OSPolicy_Resource: r})}
 }
 
 type configTask struct {
@@ -40,9 +40,8 @@ type configTask struct {
 	StartedAt time.Time `json:",omitempty"`
 
 	// ApplyConfigTaskOutput result
-	results           []*agentendpointpb.ApplyConfigTaskOutput_OSPolicyResult
-	postCheckRequired bool
-	policies          map[string]*policy
+	results  []*agentendpointpb.ApplyConfigTaskOutput_OSPolicyResult
+	policies map[string]*policy
 }
 
 type applyConfigTask struct {
@@ -50,8 +49,13 @@ type applyConfigTask struct {
 }
 
 type policy struct {
-	resources map[string]resourceIface
+	resources map[string]*resource
 	hasError  bool
+}
+
+type resource struct {
+	resourceIface
+	needsPostCheck bool
 }
 
 type resourceIface interface {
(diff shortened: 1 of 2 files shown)


agentendpoint/config_task.go
```diff
@@ -328,22 +342,38 @@ func (c *configTask) run(ctx context.Context) error {
...
+// Mark all resources that have already completed as "needs post check".
+func (c *configTask) markPostCheckRequired() {
+	for _, osPolicy := range c.Task.GetOsPolicies() {
+		plcy, ok := c.policies[osPolicy.GetId()]
+		if !ok || plcy.hasError {
+			continue
+		}
+		for _, configResource := range osPolicy.GetResources() {
+			res, ok := plcy.resources[configResource.GetId()]
+			if !ok || res == nil {
```
manjunathganga, line 369: When would it be nil?

agentendpoint/config_task.go
```diff
@@ -328,22 +342,38 @@ func (c *configTask) run(ctx context.Context) error {
...
+	c.postCheckState(ctx)
 
 	return c.reportCompletedState(ctx, "", agentendpointpb.ApplyConfigTaskOutput_SUCCEEDED)
 }
 
+// Mark all resources that have already completed as "needs post check".
+func (c *configTask) markPostCheckRequired() {
+	for _, osPolicy := range c.Task.GetOsPolicies() {
+		plcy, ok := c.policies[osPolicy.GetId()]
+		if !ok || plcy.hasError {
```
manjunathganga, line 364: Consider adding a comment about when the `!ok` condition will be met. It might not be immediately obvious when reading this in the future.

Also, a comment about why we ignore policies with errors will also help the reasons (if and when we want to modify this logic) in the future.

manjunathganga, line 364: Do we have tests to cover these scenarios?

agentendpoint/config_task.go
```diff
@@ -234,16 +242,21 @@ func (c *configTask) postCheckState(ctx context.Context) {
 	// No prepopulate run for post check as we will always check every resource.
 	for i, osPolicy := range c.Task.GetOsPolicies() {
 		ctx = clog.WithLabels(ctx, map[string]string{"config_assignment": osPolicy.GetOsPolicyAssignment(), "policy_id": osPolicy.GetId()})
-		plcy := c.policies[osPolicy.GetId()]
-
+		plcy, ok := c.policies[osPolicy.GetId()]
+		if !ok {
```
manjunathganga, line 246: When would this happen? We will execute every policy right?

agentendpoint/config_task.go
```diff
@@ -234,16 +242,21 @@ func (c *configTask) postCheckState(ctx context.Context) {
...
 		// Skip state check if this policy already has an error from a previous step.
 		if plcy.hasError {
 			clog.Debugf(ctx, "Policy has error, skipping post check.")
 			continue
 		}
-
 		pResult := c.results[i]
 		for i, configResource := range osPolicy.GetResources() {
+			res, ok := plcy.resources[configResource.GetId()]
+			if !ok || res == nil {
```
manjunathganga, line 257: Won't we return every resource in the response even if they are not executed? In that case, would we ever hit this condition?

adjackura: Same as above, no we should never hit this but a bug in the code that somehow gets past tests would cause a agent crash loop and we want to avoid that. I updated the code to log this error state

agentendpoint/config_task.go
adjackura: Normally it shouldn't, with the current flow this will always be populated. In the scenario where logic changes I want to avoid a nil pointer panic though since that will send the agent into a crash loop.

agentendpoint/config_task.go
adjackura: These should not happen, unit tests and e2e_tests would pick up if we hit this though as the ReportTaskCompleteRequest and compliance report would not be populated correctly

agentendpoint/config_task.go
adjackura: With the current logic never (production code we never set this to nil), this was a helper for unit tests where we can run these functions out of order and/or pre-populate fields with empty values.

agentendpoint/config_task.go
manjunathganga: /lgtm

PR #320, opened 2021-04-15
e2e_tests: move to using inventory api for inventory reporting

adjackura:
We no longer need to scrape serial output to determine of inventory was reported.


wj-chen:
Nice thanks for updating these tests!

PR #324, opened 2021-04-15
e2e_tests: reduce size of instances, use pd-balanced, rerun failed tests once

adjackura:
Move all linux tests to use e2-medium
Move all windows tests (other than patch) to e2-standard-2
Move all instances to use pd-balanced disks
Check inventory items for InventoryReporting tests
Rerun failed tests once to try to reduce flakes for a single test run
 - inventory tests are not rerun as that setup was difficult to rerun simply, but they don't fail often
 - both test cases will report on testgrid, if we don't like how that shows up we can instead not record the first failing test


ryanwe (approved):
/lgtm

PR #334, opened 2021-04-23
Examples for os policy assignments

manjunathganga:
/assign ryanwe
/assign adjackura


adjackura:
You probably want to be more specific with the file names as none are fully generic. Some only work on certain linux distros, some only work on windows vs linux, and we probably will want more more examples for each distro/os in the future.

adjackura (approved):

PR #335, opened 2021-04-23
e2e_tests: increase the deadline for OSPolicy tests and only start after a zone has been secured

adjackura:


e2e_tests/test_suites/ospolicies/ospolicies.go
```diff
@@ -166,6 +163,9 @@ func runTest(ctx context.Context, testCase *junitxml.TestCase, testSetup *osPoli
 	testProjectConfig := testconfig.GetProject()
 	zone := testProjectConfig.AcquireZone()
 	defer testProjectConfig.ReleaseZone(zone)
+	// No test should take longer than 60 min
```
manjunathganga, line 166: Do you want to add a comment about "only start after a zone has been secured " so that someone won't refactor and change the order in the future?

PR #342, opened 2021-04-28
e2e_tests: re-add retries to OSPolicies

adjackura:
Dropped timeout to 20min, longest test takes just over 13min it seems (including zone wait time)
Add locking with delete call to somewhat limit qps


e2e_tests/test_suites/ospolicies/ospolicies.go
```diff
@@ -301,7 +300,9 @@ func getTestCaseFromTestSetUp(testSetup *osPolicyTestSetup) (*junitxml.TestCase,
 }
 
 func cleanupOSPolicyAssignment(ctx context.Context, client *osconfigZonalV1alpha.OsConfigZonalClient, testCase *junitxml.TestCase, name string) {
+	gpMx.Lock()
```
manjunathganga, line 303: nit: Can you add a comment about why we need to acquire a lock here?

PR #345, opened 2021-04-29
ConfigTask: add additional info logs

adjackura:
Add debug log for full ApplyConfigTask dump
Add info log for starting each policy
Add info log for recording state for each resource
Allow for ExecResource to not have Enforce set (for example if policy is in VALIDATION mode)

### DIFF
--- a/config/exec_resource.go
+++ b/config/exec_resource.go
@@ -93,7 +93,7 @@ func (e *execResource) download(ctx context.Context, execR *agentendpointpb.OSPo
 			return "", err
 		}
 	default:
-		return "", fmt.Errorf("unrecognized Source type for FileResource: %q", execR.GetSource())
+		return "", fmt.Errorf("unrecognized Source type for ExecResource: %q", execR.GetSource())
 	}
 
 	return name, nil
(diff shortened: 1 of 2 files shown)


config/exec_resource.go
```diff
@@ -111,15 +111,22 @@ func (e *execResource) validate(ctx context.Context) (*ManagedResources, error)
 		return nil, err
 	}
 
-	e.enforcePath, err = e.download(ctx, e.GetEnforce())
-	if err != nil {
-		return nil, err
+	// Assume lack of Enforce means policy is in VALIDATE mode.
```
manjunathganga, line 114: Instead of assuming, can the mode be passed so that we can raise an error if the pre-condition is not satisfied?

adjackura: Not with how this is setup, resources know nothing about policies and there is no per resource customization options.

manjunathganga: I see. can you add a todo for it for reference.

adjackura: This isn't a TODO I really want to fix as it is just not possible with the resource setup. I don't think this is that big of an issue though seeing as the API should prevent this anyway, also there will be an error at enforce execution time so the end result would be the same with the agent returning an error for this resource.

PR #350, opened 2021-05-18
Config Task: add error message and ExecResource output recording

adjackura:
Pull latest proto libraries
Populate ErrorMessage field for config task steps that end in failure (max size of 512)
Add ExecResourceOutput recording (max size of 100K)
If an error with an enforce step is encountered, end the policy application but still run post check on all resources.


config/exec_resource.go
```diff
@@ -190,12 +195,52 @@ func (e *execResource) enforceState(ctx context.Context) (inDesiredState bool, e
...
+	clog.Debugf(ctx, "Reading %q for ExecResource output", outputFilePath)
+
+	f, err := os.Open(outputFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("error opening OutputFilePath: %v", err)
+	}
+	defer f.Close()
+
+	// Make a byte slice with a capacity of 1 over maxSize (for simplicity).
+	output := make([]byte, 0, maxExecOutputSize+1)
```
ryanwe, line 220: Instead of this, can you check if `err == io.EOF` instead? Basically validating that by reading `maxExecOutputSize` bytes, you've either reach the EOF of not?

config/exec_resource.go
```diff
@@ -190,12 +195,52 @@ func (e *execResource) enforceState(ctx context.Context) (inDesiredState bool, e
...
+	// Read up to capactity.
+	n, err := f.Read(output[:cap(output)])
+	output = output[:len(output)+n]
```
ryanwe, line 223: My Golang skills are lacking here, but are there any places where the code will copy the bytes instead of updating or passing references?

Also, because this can be large, is there any value in calling GC() after sending this to the service instead of waiting on the runtime?

config/repository_resource.go
```diff
@@ -311,6 +311,9 @@ func (r *repositoryResource) enforceState(ctx context.Context) (inDesiredState b
 	return true, nil
 }
 
+func (r *repositoryResource) populateOutput(rCompliance *agentendpointpb.OSPolicyResourceCompliance) {
+}
```
ryanwe, line 315: Why are these empty `populateOutput()` functions needed and why is this one formatted differently?

agentendpoint/config_task.go
```diff
@@ -26,6 +26,10 @@ import (
 	agentendpointpb "google.golang.org/genproto/googleapis/cloud/osconfig/agentendpoint/v1"
 )
 
+// This is the maximum size in bytes of the OSPolicyResourceConfigStep ErrorMessage field.
```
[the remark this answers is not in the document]
adjackura: Bytes, and yes multi-byte chars will get cut

config/repository_resource.go
adjackura: No idea why it was formatted this way, ill see if I can fix that manually. The reason for the empty methods is so that each resource type implements the interface.

config/exec_resource.go
adjackura: No, we are working with slices here so we are always passing by reference.
As far as force calling GC(), perhaps, but this really isn't that much data and the runtime should be pretty on top of this, I expect reading any amount of data a user would notice will trigger the GC to run since normal usage is 2-3mb.
Ill do some testing and what we might be able to do is modify the GC percentage for the agent to trigger GC earlier since we really don't expect memory usage to go up that much

config/exec_resource.go
adjackura: EOF wont happen until the following read, the docs don't explain this clearly but most Read implementations will only return 0, EOF, so for shorter files i would have to make an additional read call to check that we are in fact at the end of the file. I updated the code slightly here to trim that last byte if output is greater.

agentendpoint/config_task.go
```diff
@@ -363,11 +381,19 @@ func (c *configTask) run(ctx context.Context) error {
 				continue
 			}
 
-			if enforcementActionTaken := enforceConfigResourceState(ctx, plcy, rCompliance, configResource); enforcementActionTaken {
-				// On any change we trigger post check for all previous resouces.
+			// Only errors in validate and check state constitute a serious error,
```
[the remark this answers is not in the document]
adjackura: renamed to validateOrCheckError

agentendpoint/config_task.go
```diff
@@ -50,18 +54,19 @@ type applyConfigTask struct {
...
 type policy struct {
 	resources map[string]*resource
-	hasError  bool
 }
 
 type resource struct {
 	resourceIface
-	needsPostCheck bool
+	needsPostCheck  bool
+	hasSeriousError bool
```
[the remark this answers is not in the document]
adjackura: Renamed to better explain.

config/repository_resource.go
adjackura: Welp the formatter wants it this way, can't really argue with gofmt

ryanwe (approved):
🚢

PR #353, opened 2021-05-25
Run fieldalignment on all structs

adjackura:
https://pkg.go.dev/golang.org/x/tools@v0.1.0/go/analysis/passes/fieldalignment

### DIFF
--- a/agentendpoint/inventory.go
+++ b/agentendpoint/inventory.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"reflect"
-	"runtime/debug"
 	"strings"
 	"time"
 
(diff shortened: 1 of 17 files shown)


agentendpoint/inventory.go
```diff
@@ -28,6 +29,7 @@ func (c *Client) ReportInventory(ctx context.Context) {
 	state := inventory.Get(ctx)
 	write(ctx, state, inventoryURL)
 	c.report(ctx, state)
+	debug.FreeOSMemory()
```
ryanwe, line 32: nit: can you mention this in the commit description too?

ryanwe (approved):
/hold

agentendpoint/inventory.go
adjackura: ugh nice catch, I meant to remove this, but this means my testing was not as useful as I thought :(

ryanwe (approved):
/unhold
/lgtm

PR #355, opened 2021-06-01
Agent efficiency improvements/bugfixes/logging updates

adjackura:
Update agentendpoint api calls to new debug logging system
If debug logging is not enabled don't create the debug log message at all to avoid unnecessary memory usage
Set debug logging settings on every config refresh so the agent does not need to be restarted
Don't log error for config post check state if resource is nil (now expected ins some situations)
Use pointers for data structures during inventory runs to avoid unnecessarily copying data, none of this was stored on the stack anyway.
Use int64 in some places to be more explicit and allow for non 64bit builds
Create OSCONFIG_FREE_OS_MEMORY environment variable that will FreeOSMemory after each task
Create OSCONFIG_DISABLE_INVENTORY_WRITE environment variable that will disable inventory writes


ospatch/updates.go
```diff
@@ -30,22 +30,22 @@ const (
...
+	var btime int64
 	scnr := bufio.NewScanner(f)
 	for scnr.Scan() {
 		if bytes.HasPrefix(scnr.Bytes(), []byte("btime")) {
 			split := bytes.SplitN(scnr.Bytes(), []byte(" "), 2)
 			if len(split) != 2 {
 				return 0, fmt.Errorf("error parsing btime from %s: %q", stat, scnr.Text())
 			}
-			btime, err = strconv.Atoi(string(bytes.TrimSpace(split[1])))
+			btime, err = strconv.ParseInt(string(bytes.TrimSpace(split[1])), 10, 0)
```
ryanwe, line 48: Should the bitSize be `64`? Otherwise, this is identical to [Atoi](https://golang.org/pkg/strconv/#Atoi)

> Atoi is equivalent to ParseInt(s, 10, 0), converted to type int.

tasker/tasker.go
```diff
@@ -17,15 +17,19 @@ package tasker
...
 )
 
 var (
-	tc chan *task
-	wg sync.WaitGroup
-	mx sync.Mutex
+	tc           chan *task
+	wg           sync.WaitGroup
+	mx           sync.Mutex
+	freeOSMemory = strings.ToLower(os.Getenv("OSCONFIG_FREE_OS_MEMORY"))
```
ryanwe, line 32: Should this be part of `agentconfig` too?

agentendpoint/agentendpoint.go
```diff
@@ -128,41 +127,33 @@ func (c *Client) reportInventory(ctx context.Context, inventory *agentendpointpb
...
-	req.InstanceIdToken = token
-
-	// Additional logging for verifications in e2e tests.
-	payloadSummary := fmt.Sprintf("hostname %s, short name %s, %d installed packages, %d available packages", inventory.OsInfo.Hostname, inventory.OsInfo.ShortName, len(inventory.InstalledPackages), len(inventory.AvailablePackages))
-	clog.Debugf(ctx, "Calling ReportInventory with request containing %s", payloadSummary)
+	resp, err := c.raw.ReportInventory(ctx, req)
 
-	return c.raw.ReportInventory(ctx, req)
+	req.InstanceIdToken = "<redacted>"
+	clog.DebugRPC(ctx, "ReportInventory", req, resp)
```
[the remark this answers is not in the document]
adjackura: I can separate the request and response but we explicitly dont want a helper for the redacting as there is no efficient way to do that (efficiency even on the debug logs is still much more important than code duplication)

agentconfig/agentconfig.go
```diff
@@ -670,3 +673,13 @@ func RestartFile() string {
 func UserAgent() string {
 	return "google-osconfig-agent/" + Version()
 }
+
+// DisableInventoryWrite returns if the DisableInventoryWrite setting is set.
```
ryanwe, line 677: "returns true"

agentconfig/agentconfig.go
```diff
@@ -670,3 +673,13 @@ func RestartFile() string {
...
+func DisableInventoryWrite() bool {
+	return disableInventoryWrite != "true" && disableInventoryWrite != "1"
```
ryanwe, line 679: Should this be `disableInventoryWrite == "true" || disableInventoryWrite == "1"`? Or even `strings.EqualFold(disableInventoryWrite, "true") || disableInventoryWrite == "1"`?

agentconfig/agentconfig.go
```diff
@@ -670,3 +673,13 @@ func RestartFile() string {
...
+}
+
+// FreeOSMemory returns if the FreeOSMemory setting is set.
```
ryanwe, line 682: "returns true"

ryanwe (approved):

PR #362, opened 2021-06-16
Use exec.CommandContext so that canceling the context also kills any running processes

adjackura:
Add better debug logging to DefaultRunner.Run()


ryanwe (approved):
/lgtm

PR #371, opened 2021-07-22
Support Windows Application Inventory

coodie:
This change adds support for WindowsApplication.

We consider something WindowsApplication when it can be uninstalled and has a name. There are specific places in Windows Registry that Windows installation process updates whenever new application is installed. Agent reads these places and updates Inventory with them.

TODO:
- We need to add E2E tests for that feature, however currently we need to update public osconfig protos in order to do that: https://pkg.go.dev/google.golang.org/genproto/googleapis/cloud/osconfig/v1alpha#Inventory_SoftwarePackage as it misses WindowsApplication Inventory Item.


adjackura:
FYI see https://github.com/GoogleCloudPlatform/osconfig/pull/372 to unblock e2e_tests

agentendpoint/inventory.go
```diff
@@ -16,6 +16,7 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	agentendpointpb "google.golang.org/genproto/googleapis/cloud/osconfig/agentendpoint/v1"
+	date "google.golang.org/genproto/googleapis/type/date"
```
adjackura, line 19: no need for the import rename here

packages/windows_application_windows.go
```diff
@@ -0,0 +1,122 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package packages
+
+import (
+	//"golang.org/x/sys/windows"
```
adjackura, line 18: remove these

packages/windows_application_windows.go
```diff
@@ -0,0 +1,122 @@
...
+
+func GetWindowsApplications(ctx context.Context) ([]*WindowsApplication, error) {
+	directories := []string{
+		`SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall`,
+		`SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall`,
+	}
+	var allApps []*WindowsApplication
+
+	for _, dir := range directories {
+		clog.Debugf(ctx, "loading windows applications from: %v", dir)
```
adjackura, line 83: Log messages should be capitalized (here and elsewhere)

packages/windows_application_windows.go
```diff
@@ -0,0 +1,122 @@
...
+		apps, err := loadWindowsApplicationsFromRegistryDirectory(ctx, dir)
+		if err != nil {
+			clog.Debugf(ctx, "error loading windows applications from registry: %v, error: %v", dir, err)
+			continue
+		}
+		allApps = append(allApps, apps...)
+	}
+	clog.Debugf(ctx, "Loaded windows applications")
+	for _, app := range allApps {
```
adjackura, line 92: Drop this debug log

packages/windows_application_windows.go
```diff
@@ -0,0 +1,122 @@
...
+	if errName == nil && errUninstall == nil {
+		return &WindowsApplication{
+			DisplayName:    displayName,
+			DisplayVersion: getStringValueReturnEmptyIfError(k, "DisplayVersion"),
+			Publisher:      getStringValueReturnEmptyIfError(k, "Publisher"),
+			InstallDate:    parseDate(getStringValueReturnEmptyIfError(k, "InstallDate")),
+			HelpLink:       getStringValueReturnEmptyIfError(k, "HelpLink"),
+		}
+	}
+	return nil
```
ryanwe, line 72: Would it be helpful to add debug logs? Will the errors always be that the item isn't an installed application or can this fail in other ways?

packages/windows_application_windows.go
```diff
@@ -0,0 +1,122 @@
...
+
+	var result []*WindowsApplication
+	subkeys, err := dirKey.ReadSubKeyNames(0)
+	if err != nil {
+		return nil, err
+	}
+	for _, subkey := range subkeys {
+		k, err := registry.OpenKey(dirKey, subkey, registry.QUERY_VALUE)
+		if err != nil {
+			continue
```
adjackura, line 113: Log error and then continue?

packages/windows_application_windows.go
```diff
@@ -0,0 +1,122 @@
...
+		`SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall`,
+		`SOFTWARE\Wow6432Node\Microsoft\Windows\CurrentVersion\Uninstall`,
+	}
+	var allApps []*WindowsApplication
+
+	for _, dir := range directories {
+		clog.Debugf(ctx, "loading windows applications from: %v", dir)
+		apps, err := loadWindowsApplicationsFromRegistryDirectory(ctx, dir)
+		if err != nil {
+			clog.Debugf(ctx, "error loading windows applications from registry: %v, error: %v", dir, err)
```
ryanwe, line 86: Should this be clog.Warn or clog.Error? Being unable to read the registry seems like something users should be able to see in the logs.

packages/windows_application_windows.go
```diff
@@ -0,0 +1,122 @@
...
+			continue
+		}
+		allApps = append(allApps, apps...)
+	}
+	clog.Debugf(ctx, "Loaded windows applications")
+	for _, app := range allApps {
+		clog.Debugf(ctx, "%v", app)
+	}
```
ryanwe, line 94: Can these logs be combined into a single log entry like we do when installing packages?

packages/windows_application_windows.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return t
+	}
+
+	return time.Date(int(year), time.Month(month), int(day), 0, 0, 0, 0, time.Now().Location())
+}
+
+func getWindowsApplicationFromRegistryKey(ctx context.Context, k *registry.Key) *WindowsApplication {
+	displayName, _, errName := k.GetStringValue("DisplayName")
+	_, _, errUninstall := k.GetStringValue("UninstallString")
```
adjackura, line 61: Do you have examples of entries that don't have "UninstallString"? I just want some good data on why we would not want to include these.

packages/windows_application_windows.go
```diff
@@ -0,0 +1,122 @@
...
+	defer dirKey.Close()
+
+	var result []*WindowsApplication
+	subkeys, err := dirKey.ReadSubKeyNames(0)
+	if err != nil {
+		return nil, err
+	}
+	for _, subkey := range subkeys {
+		k, err := registry.OpenKey(dirKey, subkey, registry.QUERY_VALUE)
+		if err != nil {
```
ryanwe, line 112: Should this error be logged?

packages/windows_application_windows.go
```diff
@@ -0,0 +1,122 @@
...
+	//"time"
+	"context"
+	"strconv"
+	"time"
+
+	"github.com/GoogleCloudPlatform/osconfig/clog"
+	"golang.org/x/sys/windows/registry"
+)
+
+func getStringValueReturnEmptyIfError(k *registry.Key, value string) string {
```
adjackura, line 29: GetStringValue already returns "" on any error, this is actually best practice with Go functions. You should be able to remove the need for this function and just ignore error (make sure to add a comment as to why)

ryanwe (commented):
Looks great overall. Is it possible to add unit tests to windows_application_windows.go?

packages/windows_application_windows.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	day, err := strconv.ParseInt(dateString[6:8], 10, 32)
+	if err != nil {
+		return t
+	}
+
+	return time.Date(int(year), time.Month(month), int(day), 0, 0, 0, 0, time.Now().Location())
+}
+
+func getWindowsApplicationFromRegistryKey(ctx context.Context, k *registry.Key) *WindowsApplication {
```
adjackura, line 59: just getWindowsApplication

packages/windows_application_windows.go
```diff
@@ -0,0 +1,122 @@
...
+		allApps = append(allApps, apps...)
+	}
+	clog.Debugf(ctx, "Loaded windows applications")
+	for _, app := range allApps {
+		clog.Debugf(ctx, "%v", app)
+	}
+	return allApps, nil
+}
+
+func loadWindowsApplicationsFromRegistryDirectory(ctx context.Context, directory string) ([]*WindowsApplication, error) {
```
adjackura, line 98: how about getWindowsApplications

packages/windows_application_windows.go
```diff
@@ -0,0 +1,122 @@
...
+func getStringValueReturnEmptyIfError(k *registry.Key, value string) string {
+	value, _, err := k.GetStringValue(value)
+	if err != nil {
+		return ""
+	}
+	return value
+}
+
+func parseDate(dateString string) time.Time {
+	var t time.Time
```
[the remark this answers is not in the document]
coodie: time.Time isn't a reference type -> can't be nil. In this case error can be considered that it's not valid date, but we discussed on chat/gvc/somewhere else that it should just be "empty", so this function might as well just return the empty value if date isn't valid.

packages/windows_application_windows.go
coodie: In this case there would be a big flood of debug logs as quite a lot of entries inside that directory empty don't satisfy these conditions.

packages/packages.go
```diff
@@ -62,18 +62,19 @@ var (
...
+	Deb                []*PkgInfo            `json:"deb,omitempty"`
+	Zypper             []*PkgInfo            `json:"zypper,omitempty"`
+	ZypperPatches      []*ZypperPatch        `json:"zypperPatches,omitempty"`
+	COS                []*PkgInfo            `json:"cos,omitempty"`
+	Gem                []*PkgInfo            `json:"gem,omitempty"`
+	Pip                []*PkgInfo            `json:"pip,omitempty"`
+	GooGet             []*PkgInfo            `json:"googet,omitempty"`
+	WUA                []*WUAPackage         `json:"wua,omitempty"`
+	QFE                []*QFEPackage         `json:"qfe,omitempty"`
+	WindowsApplication []*WindowsApplication `json:"windowsApplication,omitempty"`
```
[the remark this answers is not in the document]
coodie: Why so? From my understanding E2E tests actually read this data back from guest attributes.

packages/windows_application_windows.go
coodie: This function exists because it returns one value so I can use it directly when creating WindowsApplication structure like that:

```
    return &WindowsApplication{
      DisplayName:    displayName,
      DisplayVersion: getStringValueReturnEmptyIfError(k, "DisplayVersion"),
      Publisher:      getStringValueReturnEmptyIfError(k, "Publisher"),
      InstallDate:    parseDate(getStringValueReturnEmptyIfError(k, "InstallDate")),
      HelpLink:       getStringValueReturnEmptyIfError(k, "HelpLink"),
    }  
```

packages/windows_application_windows.go
ryanwe: Got it. And it looks like I misread the code. You are logging all unexpected errors. 👍

packages/windows_application_windows.go
coodie: I've been examining windows registry in these locations on my local machine (by running some powershell commands and examining the output), and can't find anything that's weird and suspicious. I'm not sure if it's good enough of a test, but based on that we can probably include these entries.

packages/windows_application_windows.go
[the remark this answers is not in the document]
adjackura: In this case instantiate during the return so you don't create the unused variable. Another option would be to use a pointer so it can be nil (would be the preference if you also returned an error), but in this case that may cause additional logic in the caller so just returning time.Time{} is fine.

packages/packages.go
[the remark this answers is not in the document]
adjackura: We do not want to add this to the legacy inventory, only ReportInventory.

packages/windows_application_windows.go
adjackura: That's overly complicated and introduces an addition nil check for no good reason, just do it this way:

```
displayVersion, _, _ := k.GetStringValue("DisplayVersion"))
publisher, _, _ := k.GetStringValue("Publisher"))
return &WindowsApplication{ 
  DisplayName: displayName, 
  DisplayVersion: displayVersion,
  Publisher: publisher,
  ...
 }
```

packages/windows_application_windows.go
adjackura: My point is you check errUninstall == nil, it can see an entry not having "UninstallString" (probably not uncommon) and we would not record those applications

packages/windows_application_windows.go
adjackura: Just remove the loop completly. If you want to log all the apps you found we can create one log message

packages/windows_application_windows.go
coodie: Yep. I haven't found example of application that doesn't have UninstallString, but has other fields that actually make sense.

At the moment I'm seeing entries like: 

HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\MPlayer2
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\IE40
HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\SchedulingAgent

that only have one entry:

| Name      | Type   | Data            |
|-----------|--------|-----------------|
| (Default) | REG_SZ | (value not set) |

This entry for example:
- HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Connection Manager

has:

| Name            | Type      | Data            |
|-----------------|-----------|-----------------|
| (Default)       | REG_SZ    | (value not set) |
| SystemComponent | REG_DWORD | 0x00000001 (1)  |

While this:
- HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\WIC

has:

| Name      | Type      | Data            |
|-----------|-----------|-----------------|
| (Default) | REG_SZ    | (value not set) |
| NoRemove  | REG_DWORD | 0x00000001 (1)  |

There are few more examples. Some of these can be considered "applications" like mplayer (I can actually start it and run it from windows start menu). Not quite sure what are the others but it seems like it's something installed by default on Windows (not using typical windows installer).

adjackura (approved):

PR #373, opened 2021-07-22
Send basic inventory with RegisterAgent

adjackura:
Make sure we run RegisterAgent immediately on agent start and agent enablement
Seperate internal agent periodics (just restart checker for now) from the service specifc ones
Update some core library dependencies


agentendpoint/agentendpoint.go
```diff
@@ -101,7 +102,20 @@ func (c *Client) RegisterAgent(ctx context.Context) error {
 		return err
 	}
 
-	req := &agentendpointpb.RegisterAgentRequest{AgentVersion: agentconfig.Version(), SupportedCapabilities: agentconfig.Capabilities()}
+	oi, err := osinfo.Get()
```
jaiminsh, line 105: Should we populate this only if inventory is turned on?

main.go
```diff
@@ -170,6 +170,10 @@ func runTaskLoop(ctx context.Context, c chan struct{}) {
 		// Set debug logging settings so that customers don't need to restart the agent.
 		logger.SetDebugLogging(agentconfig.Debug())
 		clog.DebugEnabled = agentconfig.Debug()
+		if agentconfig.TaskNotificationEnabled() && taskNotificationClient == nil {
+			// Call RegisterAgent now since we just we either just started running or were just enabled.
+			registerAgent(ctx)
```
jaiminsh, line 175: Not blocking this PR, but we should also call RegisterInventory here. Something to track

adjackura: Inventory happens on it's own schedule (default once every 10 min) so between agent being enabled and an inventory run will be <10min which is totally fine.

agentendpoint/agentendpoint.go
adjackura: Fixed

PR #381, opened 2021-09-15
Added log messages for patch reports for apt-update

dowgird:


dowgird:
As discussed offline, I rethought how this should be implemented.

PR #382, opened 2021-09-16
Add more info logs, actually cleanup unmanaged repos

adjackura:
During a config run there will now be clear info logs when actions are taken or after a check has been completed.
We need to remove repos that config is no longer managing, we do this by clearing out all unmanaged repos that match our naming.
Actually limit error messages to the 512 byte limit.

For the new logging, here is an example of info logs, errors are always logged at any of the satges.
Execution Resource (success case)
- Validation successful
- Check state: resource state is NON_COMPLIANT
- Enforcing desired state for resource
- Desired state enforcement is successful
- Check state post enforcement: resource state is COMPLIANT

Execution Resource (resource is already in desired state)
- Validation successful
- Check state: resource state is COMPLIANT

### DIFF
--- a/agentendpoint/config_task.go
+++ b/agentendpoint/config_task.go
@@ -159,7 +159,7 @@ func validateConfigResource(ctx context.Context, res *resource, policyMR *config
 		if err := detectPolicyConflicts(res.ManagedResources(), policyMR); err != nil {
 			outcome = agentendpointpb.OSPolicyResourceConfigStep_FAILED
 			hasError = true
-			errMessage = fmt.Sprintf("Validate: resource conflict in policy: %v", err)
+			errMessage = truncateMessage(fmt.Sprintf("Validate: resource conflict in policy: %v", err), maxErrorMessage)
 			clog.Errorf(ctx, errMessage)
 		} else {
 			clog.Infof(ctx, "Validate: resource %q validation successful.", configResource.GetId())
--- a/agentendpoint/config_task.go
+++ b/agentendpoint/config_task.go
@@ -336,7 +336,7 @@ func (c *configTask) cleanupRepos(ctx context.Context) {
 			clog.Errorf(ctx, "Error globing directory: %v", err)
 		}
 		for _, match := range matches {
-			if err := removeIfNoMatch(match, managedRepos); err != nil {
+			if err := removeFileIfNoMatch(match, managedRepos); err != nil {
 				clog.Errorf(ctx, "Error cleaning up old repo: %v", err)
 			}
 		}
--- a/agentendpoint/config_task.go
+++ b/agentendpoint/config_task.go
@@ -311,7 +311,7 @@ func (c *configTask) generateBaseResults() {
 	}
 }
 
-func removeIfNoMatch(a string, s []string) error {
+func removeFileIfNoMatch(a string, s []string) error {
 	for _, b := range s {
 		if a == b {
 			return nil
--- a/agentendpoint/config_task.go
+++ b/agentendpoint/config_task.go
@@ -17,8 +17,11 @@ package agentendpoint
 import (
 	"context"
 	"fmt"
+	"os"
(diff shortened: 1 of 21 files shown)


agentendpoint/config_task.go
```diff
@@ -146,15 +152,17 @@ func validateConfigResource(ctx context.Context, res *resource, policyMR *config
...
-		errMessage = fmt.Sprintf("Error validating resource: %v", err)
+		errMessage = truncateMessage(fmt.Sprintf("Validate: resource %q error: %v", configResource.GetId(), err), maxErrorMessage)
 		clog.Errorf(ctx, errMessage)
 	} else {
 		// Detect any resource conflicts within this policy.
 		if err := detectPolicyConflicts(res.ManagedResources(), policyMR); err != nil {
 			outcome = agentendpointpb.OSPolicyResourceConfigStep_FAILED
 			hasError = true
-			errMessage = fmt.Sprintf("Resource conflict in policy: %v", err)
+			errMessage = fmt.Sprintf("Validate: resource conflict in policy: %v", err)
```
manjunathganga, line 162: Do you also wrap it with `truncateMessage` here?

agentendpoint/config_task.go
```diff
@@ -201,7 +209,7 @@ func enforceConfigResourceState(ctx context.Context, res *resource, rCompliance
 	clog.Debugf(ctx, "Running step 'enforce state' on resource %q.", configResource.GetId())
 	// Only enforce resources that need it.
 	if res.InDesiredState() {
-		clog.Debugf(ctx, "No enforcement required for %q.", configResource.GetId())
+		clog.Debugf(ctx, "Enforce state: No enforcement required for %q.", configResource.GetId())
```
manjunathganga, line 212: Info?

agentendpoint/config_task.go
```diff
@@ -301,7 +311,43 @@ func (c *configTask) generateBaseResults() {
 	}
 }
 
+func removeIfNoMatch(a string, s []string) error {
+	for _, b := range s {
+		if a == b {
+			return nil
+		}
+	}
+	return os.Remove(a)
```
manjunathganga, line 320: s?

config/exec_resource.go
```diff
@@ -186,6 +186,7 @@ func (e *execResource) checkState(ctx context.Context) (inDesiredState bool, err
 }
 
 func (e *execResource) enforceState(ctx context.Context) (inDesiredState bool, err error) {
+	clog.Infof(ctx, "Running Enforce for ExecResource.")
```
manjunathganga, line 189: nit: s/Enforce/enforce

Ignore if you actually meant to use upper-case E

agentendpoint/config_task.go
adjackura: This function checks if "a" is in "s" and removes it if it is not.

agentendpoint/config_task.go
adjackura: I purposely did not make this info as the check state output should make it clear, and this would log every "normal" run which i want to avoid.

agentendpoint/config_task.go
manjunathganga: Ah, that makes sense. I thought `os` is a variable. 

Would it help if the method is renamed to `removeRepoIfNoMatch`. Right now, `removeIfNoMatch` does not immediately convey the meaning in the context of config_task.go.

manjunathganga (commented):
Looks good overall. Just added one comment.

agentendpoint/config_task.go
adjackura: Renamed

manjunathganga: Did you forget to upload the changes. I still see the old name.

adjackura: Yep, all updated now

PR #384, opened 2021-09-22
Fix rpm/deb package formating for inventory reporting

adjackura:
Map deb to apt based on apt existing
Map rpm to zypper or yum based on those package systems existing


manjunathganga (commented):
/lgtm

Issue #192, opened 2020-07-23
Please store configuration in /etc/ /etc/google-osconfig-agent on Linux


rbalint:
`/etc/osconfig` is too generic and packaging also use `google-osconfig-agent` in the service name and also renames the binary go google_osconfig_agent (on Debian and Ubuntu).

PR #92, opened 2019-10-30
Fix bugs in agentendpoint code

adjackura:
Add more debug logging
Add LookupEffectiveGuestPolicies call, currently unused
Sleep the poll interval between calls (10m) if the osconfig service in not enabled
Sleep longer between API call retries on ResourceExhausted errors
Fix race in tests
Decode jwt token correctly

PR #385, opened 2021-09-23
Deprecate old config directory in favor of new cache directory

adjackura:
Fixes [#92](https://github.com/GoogleCloudPlatform/osconfig/issues/192)


agentconfig/agentconfig.go
```diff
@@ -690,6 +703,11 @@ func RestartFile() string {
 	return restartFileLinux
 }
 
+// OldRestartFile is the location of the restart required file.
+func OldRestartFile() string {
+	return restartFileLinux
```
manjunathganga, line 708: oldRestartFileLinux?

adjackura: That would have been bad, good catch. I will be running extensive tests before this gets pushed FYI.

manjunathganga (commented):
/lgtm

PR #386, opened 2021-09-23
ReportInventory: Fix bug in deb/rpm inventory, reduce calls to append

adjackura:
Because we are appending a known number of packages we can preallocate a temporary slice which we then pass to append. This avoids the constant growing of the slice because of multiple append calls.

### DIFF
--- a/agentendpoint/inventory.go
+++ b/agentendpoint/inventory.go
@@ -146,7 +146,7 @@ func formatPackages(ctx context.Context, pkgs *packages.Packages, shortName stri
 		softwarePackages = append(softwarePackages, temp...)
 	}
 	if pkgs.Zypper != nil {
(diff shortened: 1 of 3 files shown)


agentendpoint/inventory.go
```diff
@@ -104,88 +104,117 @@ func formatInventory(ctx context.Context, state *inventory.InstanceInventory) *a
...
-			})
+			}
 		}
+		softwarePackages = append(softwarePackages, temp...)
 	}
-	if packages.Zypper != nil {
-		for _, pkg := range packages.Zypper {
-			softwarePackages = append(softwarePackages, &agentendpointpb.Inventory_SoftwarePackage{
+	if pkgs.Zypper != nil {
+		temp := make([]*agentendpointpb.Inventory_SoftwarePackage, len(pkgs.Yum))
```
wj-chen, line 149: len(pkgs.Zypper)

PR #389, opened 2021-09-27
Fixed the runWithPty function to set ctty to child's filedesc

dowgird:
* changed the runWithPty function to use child's stdin for the
    controlling terminal - this is what the exec package expects

  * changed handling of errors returned by cmd.Run not to ignore all
    errors

  * changed the error return statements to return std{in,err} instead of
    nil because these can be useful for troubleshooting errors

### DIFF
--- a/packages/pty_linux.go
+++ b/packages/pty_linux.go
@@ -84,7 +84,7 @@ func runWithPty(cmd *exec.Cmd) ([]byte, []byte, error) {
 	cmd.SysProcAttr = &syscall.SysProcAttr{
 		Setctty: true,
 		Setsid:  true,
-		Ctty:    int(tty.Fd()),
+		Ctty:    0, // Stdin of the child process
 	}
 
 	var stdout bytes.Buffer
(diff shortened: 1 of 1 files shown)


packages/pty_linux.go
```diff
@@ -114,8 +114,17 @@ func runWithPty(cmd *exec.Cmd) ([]byte, []byte, error) {
...
 	err = cmd.Run()
+	if err != nil {
+		// Yum returns non-zero exit values on non-error conditions.
+		// Only fail on errors which are *not* in this category.
+		if _, ok := err.(*exec.ExitError); !ok {
+			return stdout.Bytes(), stderr.Bytes(), err
+		}
+	}
+
 	if err := tty.Close(); err != nil {
```
adjackura, line 125: We still need to close the tty if there is an error in Run

dowgird: Done.

adjackura (approved):

PR #391, opened 2021-09-28
Config: Add package info caching

adjackura:
Keep a cache of package info for config package resource so we don't need to download the package on every run
This cache is keyed to the source of the resource so multiple resources that use the exact same source will use the same cache item
On every run update the timestamp for each accessed item in the cache so that we can clean up the cache of items that have not been touched for >7 days
Failure modes are very forgiving and we just invalidate the entire cache if necessary which will result in redownloading packages and recreating the cache.

### DIFF
--- a/config/package_resource.go
+++ b/config/package_resource.go
@@ -141,7 +141,7 @@ func loadPackageInfoCache(ctx context.Context) {
 	}
 	var cache packageInfoCache
 	if err := json.Unmarshal(data, &cache); err != nil {
-		clog.Debugf(ctx, "Error reading the package info cache: %v", err)
+		clog.Debugf(ctx, "Error unmarshaling the package info cache: %v", err)
 		packageInfoCacheStore = packageInfoCache{}
 		return
 	}
@@ -154,7 +154,7 @@ func getPackageInfoFromCache(ctx context.Context, pkgFile *agentendpointpb.OSPol
 	if err != nil {
 		// Just ignore the error and return early
 		// The error mode here is just always redownload the file.
-		clog.Debugf(ctx, "Error reading the package info cache: %v", err)
+		clog.Debugf(ctx, "Error creating the package info cache key: %v", err)
 		return nil
 	}
 	packageInfo, ok := packageInfoCacheStore[key]
--- a/config/package_resource.go
+++ b/config/package_resource.go
@@ -38,6 +38,7 @@ var (
 	packageInfoCacheFile = filepath.Join(agentconfig.CacheDir(), "config_package_info.cache")
 	// Clear out the entry if the last lookup is > 7 days ago.
 	packageInfoCacheTimeout = -168 * time.Hour
+	packageInfoCacheStore   packageInfoCache
 )
 
 type packageResouce struct {
@@ -126,29 +127,37 @@ func getPackageInfoCacheKey(pkgFile *agentendpointpb.OSPolicy_Resource_File) (st
 	return base64.RawStdEncoding.EncodeToString(raw), nil
 }
 
-func getPackageInfoCache() packageInfoCache {
+func loadPackageInfoCache(ctx context.Context) {
+	if packageInfoCacheStore != nil {
+		return
+	}
 	data, err := ioutil.ReadFile(packageInfoCacheFile)
 	if err != nil {
 		// Just ignore the error and return early
-		// The error mode here is just always redownload the file.
-		return nil
+		// The error mode here is to just always redownload the file.
+		clog.Debugf(ctx, "Error reading the package info cache: %v", err)
+		packageInfoCacheStore = packageInfoCache{}
+		return
 	}
 	var cache packageInfoCache
 	if err := json.Unmarshal(data, &cache); err != nil {
-		return nil
+		clog.Debugf(ctx, "Error reading the package info cache: %v", err)
+		packageInfoCacheStore = packageInfoCache{}
+		return
 	}
-	return cache
+	packageInfoCacheStore = cache
 }
 
-func getPackageInfoFromCache(pkgFile *agentendpointpb.OSPolicy_Resource_File) *packages.PkgInfo {
-	cache := getPackageInfoCache()
+func getPackageInfoFromCache(ctx context.Context, pkgFile *agentendpointpb.OSPolicy_Resource_File) *packages.PkgInfo {
+	loadPackageInfoCache(ctx)
 	key, err := getPackageInfoCacheKey(pkgFile)
 	if err != nil {
 		// Just ignore the error and return early
 		// The error mode here is just always redownload the file.
+		clog.Debugf(ctx, "Error reading the package info cache: %v", err)
 		return nil
 	}
-	packageInfo, ok := cache[key]
+	packageInfo, ok := packageInfoCacheStore[key]
 	if !ok {
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


config/package_resource.go
```diff
@@ -100,6 +110,78 @@ func (p *packageResouce) validateFile(file *agentendpointpb.OSPolicy_Resource_Fi
...
+	LastLookup time.Time
+}
+
+func getPackageInfoCacheKey(pkgFile *agentendpointpb.OSPolicy_Resource_File) (string, error) {
+	// We use the base64 encoded binary proto as the key.
+	raw, err := proto.Marshal(pkgFile)
+	if err != nil {
+		return "", err
+	}
+	return base64.RawStdEncoding.EncodeToString(raw), nil
```
manjunathganga, line 126: Would it be better to compute the hash of the contents of the file to reduce the key size? A large file could lead to a very large key otherwise.

config/package_resource.go
```diff
@@ -100,6 +110,78 @@ func (p *packageResouce) validateFile(file *agentendpointpb.OSPolicy_Resource_Fi
...
+	}
+	return cache
+}
+
+func getPackageInfoFromCache(pkgFile *agentendpointpb.OSPolicy_Resource_File) *packages.PkgInfo {
+	cache := getPackageInfoCache()
+	key, err := getPackageInfoCacheKey(pkgFile)
+	if err != nil {
+		// Just ignore the error and return early
+		// The error mode here is just always redownload the file.
```
manjunathganga, line 148: nit: s/is/is to

config/package_resource.go
```diff
@@ -100,6 +110,78 @@ func (p *packageResouce) validateFile(file *agentendpointpb.OSPolicy_Resource_Fi
...
+		return "", err
+	}
+	return base64.RawStdEncoding.EncodeToString(raw), nil
+}
+
+func getPackageInfoCache() packageInfoCache {
+	data, err := ioutil.ReadFile(packageInfoCacheFile)
+	if err != nil {
+		// Just ignore the error and return early
+		// The error mode here is just always redownload the file.
```
manjunathganga, line 133: nit: s/is/is to

config/package_resource.go
```diff
@@ -100,6 +110,78 @@ func (p *packageResouce) validateFile(file *agentendpointpb.OSPolicy_Resource_Fi
...
+	return cache
+}
+
+func getPackageInfoFromCache(pkgFile *agentendpointpb.OSPolicy_Resource_File) *packages.PkgInfo {
+	cache := getPackageInfoCache()
+	key, err := getPackageInfoCacheKey(pkgFile)
+	if err != nil {
+		// Just ignore the error and return early
+		// The error mode here is just always redownload the file.
+		return nil
```
manjunathganga, line 149: Also log the error?

config/package_resource.go
manjunathganga: Also log the error?

config/package_resource.go
```diff
@@ -100,6 +110,78 @@ func (p *packageResouce) validateFile(file *agentendpointpb.OSPolicy_Resource_Fi
...
+	// We use the base64 encoded binary proto as the key.
+	raw, err := proto.Marshal(pkgFile)
+	if err != nil {
+		return "", err
+	}
+	return base64.RawStdEncoding.EncodeToString(raw), nil
+}
+
+func getPackageInfoCache() packageInfoCache {
+	data, err := ioutil.ReadFile(packageInfoCacheFile)
```
manjunathganga, line 130: Do we have to read the file every time or is it only required the first time?

config/package_resource.go
```diff
@@ -16,18 +16,28 @@ package config
...
 	"github.com/GoogleCloudPlatform/osconfig/packages"
 	"github.com/GoogleCloudPlatform/osconfig/util"
 
 	agentendpointpb "google.golang.org/genproto/googleapis/cloud/osconfig/agentendpoint/v1"
+	"google.golang.org/protobuf/proto"
+)
+
+var (
+	packageInfoCacheFile = filepath.Join(agentconfig.CacheDir(), "config_package_info.cache")
+	// Clear out the entry if the last lookup is > 7 days ago.
```
manjunathganga, line 39: Should we be using one of the standard supported cache packages listed below instead of implementing one.

The standard packages should help to avoid running into memory issues if there are too many entries in the map within this time period

https://g3doc.corp.google.com/util/cache/go/README.md?cl=head

WDYT?

adjackura: Those are in memory caches (and not opened source), we don't need a complex cache, just a local set of files. File size should really not be an issue because the agent will already be hold this in memory (it's just the API calls). I have an idea on how we can reduce file sizes though (one cache per assignment) that I am going to look at.

config/package_resource.go
adjackura: We are not caching the file, just the proto field File for a package resource. This is always very small and most likely much smaller than a hash would be (sha256 hash is 64 in hex format)

config/package_resource.go
```diff
@@ -126,60 +127,70 @@ func getPackageInfoCacheKey(pkgFile *agentendpointpb.OSPolicy_Resource_File) (st
...
-		return nil
+		// The error mode here is to just always redownload the file.
+		clog.Debugf(ctx, "Error reading the package info cache: %v", err)
+		packageInfoCacheStore = packageInfoCache{}
+		return
 	}
 	var cache packageInfoCache
 	if err := json.Unmarshal(data, &cache); err != nil {
-		return nil
+		clog.Debugf(ctx, "Error reading the package info cache: %v", err)
```
manjunathganga, line 144: unmarshaling?

config/package_resource.go
```diff
@@ -126,60 +127,70 @@ func getPackageInfoCacheKey(pkgFile *agentendpointpb.OSPolicy_Resource_File) (st
...
+	if packageInfoCacheStore != nil {
+		return
+	}
 	data, err := ioutil.ReadFile(packageInfoCacheFile)
 	if err != nil {
 		// Just ignore the error and return early
-		// The error mode here is just always redownload the file.
-		return nil
+		// The error mode here is to just always redownload the file.
+		clog.Debugf(ctx, "Error reading the package info cache: %v", err)
```
manjunathganga, line 138: Should these be warnings?

config/package_resource.go
```diff
@@ -126,60 +127,70 @@ func getPackageInfoCacheKey(pkgFile *agentendpointpb.OSPolicy_Resource_File) (st
...
 
-func getPackageInfoFromCache(pkgFile *agentendpointpb.OSPolicy_Resource_File) *packages.PkgInfo {
-	cache := getPackageInfoCache()
+func getPackageInfoFromCache(ctx context.Context, pkgFile *agentendpointpb.OSPolicy_Resource_File) *packages.PkgInfo {
+	loadPackageInfoCache(ctx)
 	key, err := getPackageInfoCacheKey(pkgFile)
 	if err != nil {
 		// Just ignore the error and return early
 		// The error mode here is just always redownload the file.
+		clog.Debugf(ctx, "Error reading the package info cache: %v", err)
```
manjunathganga, line 157: Should it be "Error creating package info cache key"?

manjunathganga (commented):
Looks good overall. Some minor comments.

config/package_resource.go
adjackura: I dont want spam, users can't solve the issue so I want the logs to be very minimal.

PR #392, opened 2021-10-04
Added patch report log messages for Yum and Apt

dowgird:
This is a demo for Yum and Apt, if this looks good I'll update the rest of the package managers, PTAL

### DIFF
--- a/ospatch/apt_upgrade.go
+++ b/ospatch/apt_upgrade.go
@@ -98,7 +98,6 @@ func RunAptGetUpgrade(ctx context.Context, opts ...AptGetUpgradeOption) error {
 		return nil
 	}
 
-	clog.Infof(ctx, "Updating %s", msg)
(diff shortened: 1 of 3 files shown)


ospatch/apt_upgrade.go
```diff
@@ -97,7 +97,15 @@ func RunAptGetUpgrade(ctx context.Context, opts ...AptGetUpgradeOption) error {
 		clog.Infof(ctx, "Running in dryrun mode, not updating %s", msg)
 		return nil
 	}
+
 	clog.Infof(ctx, "Updating %s", msg)
```
adjackura, line 101: This log line should be removed now as it is duplicated in logPackages

dowgird: Done.

adjackura (approved):

PR #401, opened 2021-11-11
Create cloud-ops-linux.yaml

smaksimov:


adjackura (changes requested):
Seperate the script downloading into a file resource, and make sure to use dry run with exec resource validation

PR #404, opened 2021-11-16
Add retry logic for RegisterAgent

adjackura:
We will retry RegisterAgent forever now, this will block other functions on first start but this is WAI as we can not do anything else until RegisterAgent has completed anyway

### DIFF
--- a/main.go
+++ b/main.go
@@ -76,13 +76,23 @@ func (s *serialPort) Write(b []byte) (int, error) {
 
 var deferredFuncs []func()
 
+// RegisterAgent is a blocking call, the RPC itself has retry logic baked it
+// with jitter and backoff up to a total of 10 minutes.
+// If client creation or register agent (after retries) fail we then wait for
+// 5 minutes and try again.
 func registerAgent(ctx context.Context) {
-	if agentconfig.TaskNotificationEnabled() || agentconfig.GuestPoliciesEnabled() {
-		if client, err := agentendpoint.NewClient(ctx); err != nil {
-			logger.Errorf(err.Error())
-		} else if err := client.RegisterAgent(ctx); err != nil {
-			logger.Errorf(err.Error())
+	for {
+		if agentconfig.TaskNotificationEnabled() || agentconfig.GuestPoliciesEnabled() {
+			if client, err := agentendpoint.NewClient(ctx); err != nil {
+				logger.Errorf(err.Error())
+			} else if err := client.RegisterAgent(ctx); err != nil {
+				logger.Errorf(err.Error())
+			} else {
+				// RegisterAgent completed successfully.
+				return
+			}
 		}
+		time.Sleep(5 * time.Minute)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -189,6 +199,7 @@ func runTaskLoop(ctx context.Context, c chan struct{}) {
 		clog.DebugEnabled = agentconfig.Debug()
 		if agentconfig.TaskNotificationEnabled() && taskNotificationClient == nil {
 			// Call RegisterAgent now since we just we either just started running or were just enabled.
```
manjunathganga, line 201: nit: s/we just we either/we just either

main.go
```diff
@@ -189,6 +199,7 @@ func runTaskLoop(ctx context.Context, c chan struct{}) {
...
+			// This call is blocking until as we can't continue unless register agent has completed.
```
manjunathganga, line 202: nit: s/until as/until successful as

main.go
```diff
@@ -76,13 +76,23 @@ func (s *serialPort) Write(b []byte) (int, error) {
 
 var deferredFuncs []func()
 
+// RegisterAgent is a blocking call, the RPC itself has retry logic baked it
```
manjunathganga, line 79: nit: s/baked it/baked in it?

main.go
```diff
@@ -76,13 +76,23 @@ func (s *serialPort) Write(b []byte) (int, error) {
...
+// If client creation or register agent (after retries) fail we then wait for
+// 5 minutes and try again.
 func registerAgent(ctx context.Context) {
-	if agentconfig.TaskNotificationEnabled() || agentconfig.GuestPoliciesEnabled() {
-		if client, err := agentendpoint.NewClient(ctx); err != nil {
-			logger.Errorf(err.Error())
-		} else if err := client.RegisterAgent(ctx); err != nil {
-			logger.Errorf(err.Error())
+	for {
+		if agentconfig.TaskNotificationEnabled() || agentconfig.GuestPoliciesEnabled() {
```
manjunathganga, line 85: Curious to know why this check is required. Is this to handle older agents?

I think a comment might help here since it is not obvious why these checks are required.

main.go
```diff
@@ -76,13 +76,23 @@ func (s *serialPort) Write(b []byte) (int, error) {
...
+			if client, err := agentendpoint.NewClient(ctx); err != nil {
+				logger.Errorf(err.Error())
+			} else if err := client.RegisterAgent(ctx); err != nil {
+				logger.Errorf(err.Error())
+			} else {
+				// RegisterAgent completed successfully.
```
manjunathganga, line 91: Would it be helpful to add a log statement here so that the lack of it will be a signal that registeragent was not successful?

adjackura: The agent logs the failures and all the retries as errors, there really isn't a need to log the success case as it's not helpful to customers. When debug logging is enabled it will log the RPC request/response

main.go
adjackura: Updated, the logic here to move this check to the caller so this is clearer

PR #412, opened 2022-01-25
Made registerAgent stop retrying on PreconditionFailed

dowgird:


main.go
```diff
@@ -85,13 +87,24 @@ func registerAgent(ctx context.Context) {
...
+			}
 		} else {
 			// RegisterAgent completed successfully.
 			return
 		}
 		time.Sleep(5 * time.Minute)
 	}
+	// Agent rejected, just wait for the possible parent context termination.
+	select {
+	case <-ctx.Done():
```
adjackura, line 106: We can't hang forever, unless we know that this error will never ever change. We should instead just keep it in the loop and keep retrying (but only log the first time.)

PR #418, opened 2022-02-21
Add --allow-downgrades flag to apt-get calls

Gulio:
When apt-get calls fails because of a package downgrade, try again with --allow-downgrades flag

Downgrading is a default behavior for many package managers even apt (just apt not apt-get command) and should be considered the same way as upgrading a package in terms of safety concerns.


adjackura:
Overall I think this looks good but I want an e2e test put in first (it should fail) and then we con submit this and see the test pass.

adjackura (approved):

PR #420, opened 2022-02-23
Create e2e test that runs apt-get in a state that makes it downgrade a package

Gulio:
Apt-get to downgrade a package when using -y flag needs and additional --allow-downgrades flag

This new test should fail in the current state of the code and https://github.com/GoogleCloudPlatform/osconfig/pull/418 should fix that

### DIFF
--- a/e2e_tests/test_suites/patch/patch.go
+++ b/e2e_tests/test_suites/patch/patch.go
@@ -41,7 +41,8 @@ import (
 )
 
 const (
-	testSuiteName = "OSPatch"
+	testSuiteName              = "OSPatch"
+	AptDowngradeScriptLocation = "./set_apt_downgrade.sh"
 )
 
 var (
@@ -446,7 +447,7 @@ func patchConfigWithPrePostSteps() *osconfigpb.PatchConfig {
 }
 
 func patchConfigWithPreStepsForcingDowngrade() *osconfigpb.PatchConfig {
-	linuxPreStepConfig := &osconfigpb.ExecStepConfig{Executable: &osconfigpb.ExecStepConfig_LocalPath{LocalPath: "./set_apt_downgrade.sh"}, Interpreter: osconfigpb.ExecStepConfig_SHELL}
+	linuxPreStepConfig := &osconfigpb.ExecStepConfig{Executable: &osconfigpb.ExecStepConfig_LocalPath{LocalPath: AptDowngradeScriptLocation}, Interpreter: osconfigpb.ExecStepConfig_SHELL}
 
 	preStep := &osconfigpb.ExecStep{LinuxExecStepConfig: linuxPreStepConfig}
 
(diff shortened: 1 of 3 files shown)


e2e_tests/test_suites/patch/test_setup.go
```diff
@@ -189,6 +205,15 @@ func aptHeadImageTestSetup() []*patchTestSetup {
 	return imageTestSetup(mapping)
 }
 
+func aptDownradeImageTestSetup() []*patchTestSetup {
+	// This maps a specific patchTestSetup to test setup names and associated images.
+	mapping := map[*patchTestSetup]map[string]string{
+		aptDowngradeSetup: utils.DowngradeAptImages,
```
adjackura, line 211: just use HeadAptImages here?

e2e_tests/test_suites/patch/test_setup.go
```diff
@@ -78,6 +78,13 @@ curl -X PUT --data "${new}" $uri -H "Metadata-Flavor: Google"
 	linuxLocalPrePatchScript = `
 echo 'curl -X PUT --data "1" http://metadata.google.internal/computeMetadata/v1/instance/guest-attributes/osconfig_tests/pre_step_ran -H "Metadata-Flavor: Google"' >> ./linux_local_pre_patch_script.sh
 chmod +x ./linux_local_pre_patch_script.sh
+`
+
+	setUpDowngradeScript = `
```
adjackura, line 83: Instead of making this a pre patch script just take these actions here. Your already setting this ins the startup script on line 115 _after_ osconfig gets installed so you should be safe to just pin the package directly and then you can remove the PreStep completly.

e2e_tests/test_suites/patch/patch.go
```diff
@@ -128,6 +129,15 @@ func TestSuite(ctx context.Context, tswg *sync.WaitGroup, testSuites chan *junit
 		}
 		go runTestCase(tc, f, tests, &wg, logger, testCaseRegex)
 	}
+	// Test that pre- and post-patch steps run as expected.
```
adjackura, line 132: comment needs to be updated

adjackura (changes requested):

e2e_tests/test_suites/patch/test_setup.go
Gulio: I haven't tested how adding other debians than 10 works for this test I am pinning a package from a certain date that was meant for debian 10

Also HeadAptImages have ubuntus

e2e_tests/test_suites/patch/test_setup.go
Gulio: Fixed

e2e_tests/test_suites/patch/patch.go
Gulio: Fixed

e2e_tests/test_suites/patch/test_setup.go
adjackura: But this is an apt change test, we should test all apt images

adjackura (approved):

PR #427, opened 2022-03-15
adding authentication signature for onprem vms

vibhorgoogle:


vibhorgoogle:
its still work in progress

PR #439, opened 2022-08-12
replace /usr/bin/pip with /usr/bin/pip3

atetubou:
Python2 is deprecated now.

### DIFF
--- a/packages/pip.go
+++ b/packages/pip.go
@@ -34,7 +34,7 @@ var (
 
 func init() {
 	if runtime.GOOS != "windows" {
-		pip = "/usr/bin/pip"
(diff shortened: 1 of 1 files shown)


packages/pip.go
```diff
@@ -34,7 +34,7 @@ var (
 
 func init() {
 	if runtime.GOOS != "windows" {
-		pip = "/usr/bin/pip"
+		pip = "/usr/bin/pip3"
```
adjackura, line 37: This isn't safe as for all distros that we support, also we are not sure if pip is something we want to continue to scan by default because of a number of issues we have with it.

PR #445, opened 2022-10-13
Delete yum transaction files if created.

wrigri:


packages/yum.go
```diff
@@ -131,6 +134,35 @@ func parseYumUpdates(data []byte) []*PkgInfo {
...
+	   is saved look like:
+
+	     Exiting on user command
+	     Your transaction was saved, rerun it with:
+	      yum load-transaction /tmp/yum_save_tx.2022-10-12.20-26.j3auah.yumtx
+	*/
+	lines := bytes.Split(bytes.TrimSpace(data), []byte("\n"))
+
+	var filename string
+	for _, ln := range lines {
```
dorileo, line 148: I'd suggest a change, but feel free to ignore it (it's more a readability/stylistic suggestion):

```go
	for _, ln := range lines {
		flds := bytes.Fields(ln)
		if len(flds) == 3 && string(flds[1]) == "load-transaction" &&  strings.HasPrefix(filename, "/tmp/yum_save_tx.") {
			return string(flds[2])
		}
	}
```

packages/yum.go
```diff
@@ -180,6 +212,17 @@ func listAndParseYumPackages(ctx context.Context, opts ...YumUpdateOption) ([]*P
 		return nil, nil
 	}
 
+	// Some versions of yum will leave a transaction file in /tmp when update
+	// is run with --assumeno. This will attempt to delete that file.
+	yumTXFile := getYumTXFile(stdout)
```
dorileo, line 217: I'm wondering, if yum fails won't `ptyrunner.Run()` return an error? Also, even if yum doesn't fail it's very likely that yum will produce stdout and we're checking for stdout before the execution reaches this block and returning. What am I missing?

adjackura: on 211 we return `if stdout == nil`

packages/yum.go
```diff
@@ -180,6 +212,17 @@ func listAndParseYumPackages(ctx context.Context, opts ...YumUpdateOption) ([]*P
...
+	if yumTXFile != "" {
+		clog.Debugf(ctx, "Removing yum tx file: %s", yumTXFile)
+		err := os.Remove(yumTXFile)
```
adjackura, line 220: if err:= ....; err != nil {

packages/yum.go
wrigri: Updated. The original was easier to develop with, but once I had a unit test for it, I think this is a little cleaner. (I didn't copy/paste your suggestion, but it's mostly the same.)

packages/yum.go
wrigri: Yes, I don't think this is an issue. If there isn't anything in stdout, it will exit above. If there is anything in stdout, then this function call will be safe to call no matter what's in stdout.

packages/yum.go
wrigri: Updated.

adjackura (approved):

PR #454, opened 2023-01-19
Call FQDN

Gulio:
Not calling FQDN makes linux check ndots settings in `resolv.conf` and sometimes do a lot of DNS queries


paulinakania (approved):
LGTM

adjackura (approved):

PR #472, opened 2023-04-05
Set DEBIAN_FRONTEND=noninteractive for apt-get interactions

burov:
Agent doesn't support interactive frontend, apt-get will always use default options in all interactions if DEBIAN_FRONTEND=noninteractive is set.


dorileo:
Please, don't hardcode setting the environment variable like this but make it honor whatever is defined on the user's environment, that way we can turn it on and off by setting the env var, for systemd units one can even use `Environment=DEBIAN_FRONTEND=noninteractive` directive.

adjackura (approved):

PR #483, opened 2023-08-29
[Code clean-up] Migrate empty interface to any

burov:


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/GoogleCloudPlatform/osconfig
 
-go 1.17
+go 1.18
```
dorileo, line 3: Why not 1.19 and align with guest-agent?

burov: I use minimal version required for the change, as well i think we might bump it to 1.21 pretty soon due to the latest fix in golang DNS resolution library, so i think it might be a good idea to bump it to 1.19 to make 1.21 bump safer in a future

Gulio (approved):

dorileo (approved):

PR #486, opened 2023-09-15
Write inventory to attributes only if enabled

burov:
* Add enable-guest-attributes to the config.
* Write inventory to guest attributes only if guest attributes enabled.


paulinakania:
Can you add a test for this please?

michaljankowiak (approved):

PR #491, opened 2023-10-24
Removing obsolete OS images and adding new ones

michaljankowiak:
Fixing failing e2e tests. Removing old OS images that no longer exist. Adding new images and releases.


paulinakania (approved):
LGTM

PR #531, opened 2024-02-07
Update packaging as of Ubuntu devel packaging

utkarsh2102:
Hello,

This PR is an initiative to align the Debian packaging of osconfig-agnet in Ubuntu and upstream (here).

The diff represents the changes we have in Ubuntu. I'm forwarding them here as they're all relevant and follow the standard Debian packaging practices. These changes are what we have in Ubuntu devel (current development branch).

These changes change how packaging is done and don't affect the resulting binaries (that is, there is no functional change). Also, it's worth noting that since the Go dependencies have been vendored, it should be sort of ignored while reviewing the PR. Other than that, the changes are small.

Please let me know if you have any questions or concerns.


utkarsh2102:
@dorileo @a-crate, hey, this is ready for your review! :)

a-crate:
Hi Utkarsh, I can't review this one because my team doesn't own this code or this repo. It looks like prow has assigned some reviewers from the OSConfig team though.

PR #537, opened 2024-03-19
Enable OSConfig agent to read GPG keys files with multiple entities

MahmoudOuka:
Fixes OSConfig-agent to read all the entities in a GPG file. Previously, only the first entity was processed.

/assign @dowgird
/cc @dowgird
/cc @adjackura

### DIFF
--- a/config/repository_resource.go
+++ b/config/repository_resource.go
@@ -31,7 +31,6 @@ import (
 	"github.com/GoogleCloudPlatform/osconfig/util"
 	"golang.org/x/crypto/openpgp"
 	"golang.org/x/crypto/openpgp/armor"
-	"golang.org/x/crypto/openpgp/packet"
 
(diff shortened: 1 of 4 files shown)


config/repository_resource.go
```diff
@@ -176,40 +175,50 @@ func zypperRepoContents(repo *agentendpointpb.OSPolicy_Resource_RepositoryResour
...
+		return true
+	}
+
+	return false
+}
+
+func fetchGPGKey(key string) (openpgp.EntityList, error) {
 	resp, err := http.Get(key)
 	if err != nil {
-		return nil, fmt.Errorf("error downloading gpg key: %v", err)
```
dowgird, line 182: The standard approach is to use `%w` for `err` so as to chain the errors (here and below).

dowgird (approved):

PR #618, opened 2024-06-04
Fix the logic of pick region for Artifact Registry function

MahmoudOuka:
/assign @dowgird
/cc @dowgird
/cc @ekremenetskii
/cc @YahiaBadr

### DIFF
--- a/e2e_tests/utils/utils.go
+++ b/e2e_tests/utils/utils.go
@@ -97,15 +97,18 @@ func getRegionFromZone(zone string) string {
 // pickTestRegionForArtifactRegistry selects a random zone from the configured zones to pull osconfig-agent package from AR & selected-region
 func pickTestRegionForArtifactRegistry() string {
 	zones := config.Zones()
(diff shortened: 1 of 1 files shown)


e2e_tests/utils/utils.go
```diff
@@ -97,15 +97,18 @@ func getRegionFromZone(zone string) string {
...
-	for zone := range zones {
-		if currentIndex == randomIndex {
-			randomZone = zone
-		}
+	if len(zones) == 0 {
+		// default region for tests
+		return "us-central1"
+	}
+
+	zoneKeys := make([]string, 0, len(zones))
```
dowgird, line 106: There's maps.Keys() function which does this but it is in experimental, so I think it's okay not to use it.

dowgird (approved):

e2e_tests/utils/utils.go
MahmoudOuka: Thanks for letting me know, I don't know it. So, I used the classic way to do this.

PR #626, opened 2024-06-19
Adding Architecture to the packages being installed/updated in yum repo

darthmelonder:
Currently, OsConfig Agent does not support installing packages with multiarchitecture (where both 32bit and 64 bit architecture packages are present). To support this, we are going to thereby use "packageName.arch" instead of "packageName" in yum install <packageName> command.


MahmoudOuka (approved):

packages/packages.go
```diff
@@ -81,7 +81,7 @@ type Packages struct {
 
 // PkgInfo describes a package.
 type PkgInfo struct {
-	Name, Arch, Version string
+	Name, Arch, Version, NameWithArch string
```
burov, line 84: Why we need NameWithArch if we already have Name and Arch in the structure?

burov (changes requested):

packages/packages.go
burov: I suggest to replace Arch with a specific type rather than just string, it might store both raw and standardized architecture, that will help us to avoid create new field that is required only to specific use-case for specific package manager.

darthmelonder: The already existing Arch is a little different, it resolves the different architecture this way (https://github.com/GoogleCloudPlatform/osconfig/blob/a235d3366ebc7183a106d62dddd22cfe1f18822d/osinfo/osinfo.go#L31) while reporting to inventory. 

I am not comfortable changing this for our existing reporting, hence the new field.

ospatch/yum_update.go
```diff
@@ -108,7 +108,7 @@ func RunYumUpdate(ctx context.Context, opts ...YumUpdateOption) error {
 
 	var pkgNames []string
 	for _, pkg := range fPkgs {
-		pkgNames = append(pkgNames, pkg.Name)
+		pkgNames = append(pkgNames, pkg.NameWithArch)
```
burov, line 111: As i explained in a comment below it is not very hard to pass data required for construction NameWithArch in idiomatic way, i totally against adding fields that is required for one specific use case to structure that used across code base hugely

burov (changes requested):

packages/packages.go
darthmelonder: And manual testing by me verified that "yum install foo.noarch" succeeds whereas the resolved "yum install foo.all" will fail here.

ospatch/yum_update.go
darthmelonder: Makes sense, updated the structure.

packages/packages.go
```diff
@@ -81,9 +81,15 @@ type Packages struct {
...
 type PkgInfo struct {
-	Name, Arch, Version string
+	Name, Version string
+	Arch          OSArch
+	Source        Source
+}
 
-	Source Source
+// OSArch describes a OS Architecture specified for package.
+type OSArch struct {
```
burov, line 90: Type can be called just `Architecture`

ospatch/yum_update.go
```diff
@@ -108,7 +108,11 @@ func RunYumUpdate(ctx context.Context, opts ...YumUpdateOption) error {
 
 	var pkgNames []string
 	for _, pkg := range fPkgs {
-		pkgNames = append(pkgNames, pkg.Name)
+		nameWithArch := pkg.Name
```
burov, line 111: it might be a good idea to create separate function to construct name, rather than do it in place, function like would encapsulate the logic effectively 
```
func packageNameFrom(name string, arch Architecture) {
...
}
```
or 

```
func packageNameFrom(pkg *PkgInfo) {
...
}
```

packages/packages.go
burov, line 90: I suggest to create constructor function like

```
func NewArchitecture(raw string) Architecture {
    return Architrecture{StandardArch: osinfo.Architecture(raw), RawArch: raw}
}
```

That would create all code that initialize structure cleaner

packages/packages.go
```diff
@@ -81,9 +81,15 @@ type Packages struct {
...
+	StandardArch string
```
burov, line 91: I'm thinking of how to make a bit more readable usage of this two fields, i think that might be reasonable to create two methods 

```
func (a Architecture) Raw() string {
    return a.RawArch
}

func (a Architecture) Architecture() string {
    return a.StandardArch
}
```

This will give a signal that raw architecture is rather additional not main data stored in the structure and that in most scenarios standard architecture should be used

burov: As well it might be a good idea to comment fields as current names is not straightforward for me

packages/yum.go
```diff
@@ -131,7 +131,7 @@ func parseYumUpdates(data []byte) []*PkgInfo {
 			}
 			break
 		}
-		pkgs = append(pkgs, &PkgInfo{Name: string(pkg[0]), Arch: osinfo.Architecture(string(pkg[1])), Version: string(pkg[2])})
+		pkgs = append(pkgs, &PkgInfo{Name: string(pkg[0]), Arch: OSArch{StandardArch: osinfo.Architecture(string(pkg[1])), RawArch: string(pkg[1])}, Version: string(pkg[2])})
```
burov, line 134: Line become to long, that might be a good idea to separate structure construction and changing slice

burov (changes requested):

ospatch/yum_update.go
```diff
@@ -78,6 +78,14 @@ func YumDryRun(dryrun bool) YumUpdateOption {
 	}
 }
 
+func packageNameWithArch(pkg *packages.PkgInfo) string {
+	nameWithArch := pkg.Name
```
burov, line 82: Small note about name of the function, as function is not always returns package with architecture, maybe just rename it to packageNameFrom. But i'm not insisting on that

ospatch/yum_update.go
```diff
@@ -78,6 +78,14 @@ func YumDryRun(dryrun bool) YumUpdateOption {
...
+	if len(pkg.Arch.RawArch) > 0 {
```
burov, line 83: For the consistency i'm encourage you to create method to expose the data rather than just use field, but i will not insist on that

burov (approved):

ospatch/yum_update.go
burov: I think usage of mixed model might be misleading, if data model is not straightforward enough to just expose it (make all fields public) we might need to introduce an API (a set of methods that restrict user's usage patterns to the correct one) the problem with current model (i'm not thinking it is blocking for this PR submission, but might be harmful in the future) both fields are still public you're not restricting wrong usage patterns. And one last question that might be a good hint, is there a benefit of not having the method?

burov (approved):

PR #648, opened 2024-07-29
Fix for old state file on Windows

MarcMarc:


MahmoudOuka:
I think we will need to remove this [condition](https://github.com/GoogleCloudPlatform/osconfig/blob/e54f7841c62df7b83af0f371ca78aecdbb181736/agentendpoint/task_state.go#L54) from loadState function because the changes that is done on OldTaskStateFile() will be useless for Windows as long as this condition exist.

MahmoudOuka (approved):

PR #786, opened 2025-02-03
Fix the vet linter errors for non-constant format strings in logging calls.

zoltak-g:


burov (approved):
I think it is a good start, we still might want to change it to specific formats if it possible

Issue #755, opened 2024-12-16
"RebootConfig = always" logic in pre and post-patch seems wrong


chrisboulton:
We recently had an incident where an old version of a daemon was running in memory while the new version had been installed on disk (during an OS patch run). This was odd to me, because we configure all of our OS patch jobs with `RebootConfig = always` (https://cloud.google.com/compute/docs/osconfig/rest/v1/PatchConfig#rebootconfig):

> ALWAYS: Always reboot the machine after the update completes.

Here's the transcript of `OSConfigAgent`:

```
OSConfigAgent[517]: 2024-12-03T02:43:03.9463Z OSConfigAgent Info: Beginning ExecStepTask
OSConfigAgent[517]: 2024-12-03T02:43:23.9832Z OSConfigAgent Info: Command exit code: 0, out:
OSConfigAgent[517]: 2024-12-03T02:43:24.0294Z OSConfigAgent Info: Successfully completed ApplyConfigTask
OSConfigAgent[517]: 2024-12-03T02:43:34.4493Z OSConfigAgent Info: Beginning ApplyPatchesTask
OSConfigAgent[517]: 2024-12-03T02:43:34.4845Z OSConfigAgent Info: System indicates a reboot is required.
OSConfigAgent[517]: 2024-12-03T02:44:04.6225Z OSConfigAgent Warning: Error waiting for task (attempt 1 of 10): rpc error: code = Canceled desc = context canceled
OSConfigAgent[517]: 2024-12-03T02:44:04.6227Z OSConfigAgent Info: OSConfig Agent (version 20240926.03-g1) shutting down.

-- reboot happens here --

OSConfigAgent[518]: 2024-12-03T02:44:41.2684Z OSConfigAgent Info: OSConfig Agent (version 20240926.03-g1) started.
OSConfigAgent[518]: 2024-12-03T02:44:44.5830Z OSConfigAgent Info: Beginning ApplyPatchesTask
OSConfigAgent[518]: 2024-12-03T02:45:02.8319Z OSConfigAgent Info: Updating 2 packages: ["tzdata all 2024b-0+deb12u1" "vault x86_64 1.18.2-1"]
OSConfigAgent[518]: 2024-12-03T02:45:28.1150Z OSConfigAgent Info: Success. Updated 2 packages: ["tzdata all 2024b-0+deb12u1" "vault x86_64 1.18.2-1"]
[...]

PR #788, opened 2025-02-03
Rename `RebootCount` to `PostPatchRebootCount` and only increment it for post-patch reboots.

zoltak-g:
`RebootCount` was used only for forcing post-patch reboots via `RebootConfig = always`, but incremented for every reboot. As a result, the forced post-patch reboot was skipped if pre-patch reboot was required.

Note that the rename should be fine, since reading the old state file will result with counter being reset (so worst case, we will have one additional restart, but not more)

Closes: #755

### DIFF
--- a/agentendpoint/patch_task.go
+++ b/agentendpoint/patch_task.go
@@ -34,9 +34,10 @@ func systemRebootRequired(ctx context.Context) (bool, error) {
 type patchStep string
 
 const (
-	prePatch  = "PrePatch"
-	patching  = "Patching"
-	postPatch = "PostPatch"
+	prePatch              = "PrePatch"
+	patching              = "Patching"
+	postPatch             = "PostPatch"
+	totalRebootCountLimit = 5
 )
 
 type patchTask struct {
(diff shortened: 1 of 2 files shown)


zoltak-g:
Note that the rename should be fine, since reading the old state file will result with counter being reset (so worst case, we will have one additional restart, but not more)

agentendpoint/patch_task.go
```diff
@@ -45,11 +45,11 @@ type patchTask struct {
...
-	TaskID      string
-	Task        *applyPatchesTask
-	StartedAt   time.Time `json:",omitempty"`
-	PatchStep   patchStep `json:",omitempty"`
-	RebootCount int
+	TaskID               string
+	Task                 *applyPatchesTask
+	StartedAt            time.Time `json:",omitempty"`
+	PatchStep            patchStep `json:",omitempty"`
+	PostPatchRebootCount int
```
burov, line 52: I'm a bit scared that we might get prePatch reboots loop, do you think it might be a good idea to leave RebootCount counter and break the loop of prePartch reboots if the RebootCount counter is bigger than 3 or 5 smth like that?

agentendpoint/patch_task.go
```diff
@@ -200,7 +200,16 @@ func (r *patchTask) rebootIfNeeded(ctx context.Context, prePatch bool) error {
 		return nil
 	}
 
-	r.RebootCount++
+	if totalRebootCount := r.PrePatchRebootCount + r.PostPatchRebootCount; totalRebootCount > totalRebootCountLimit {
```
burov, line 203: Are we get back to the same scenario were we will skip the upgrades, 

I don't think we resetting the state of the counters anywhere, so after 5 pre-patch reboots we will never trigger post-patch reboot, am i get it right ?

burov (approved):
Should work for the scenario you mentioned, i think we still might have some problems with pre-patch reboots, but we will have bigger problems without this change, thank you!

PR #795, opened 2025-03-14
Add tests for retryutil package.

burov:
* Write tests for retryutil package.
* Refactor RetryAPICall to simplify the code.
* Fix bug with error mapping in RetryAPICall.

I was not able to trigger the case with metadata.NotDefinedError in tests, i fix the logic, but i suspect we can get rid of this case.

### DIFF
--- a/retryutil/retry.go
+++ b/retryutil/retry.go
@@ -29,6 +29,8 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+var currentSleeper sleeper = defaultSleeper{}
+
 // RetrySleep returns a pseudo-random sleep duration.
 func RetrySleep(base int, extra int) time.Duration {
 	// base=1 and extra=0 => 1*1+[0,1] => 1-2s
@@ -63,7 +65,7 @@ func RetryFunc(ctx context.Context, maxRetryTime time.Duration, desc string, f f
 		}
 
 		clog.Errorf(ctx, "Error %s, attempt %d, retrying in %s: %v", desc, i, ns, err)
-		time.Sleep(ns)
+		currentSleeper.Sleep(ns)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


retryutil/retry.go
```diff
@@ -29,6 +29,8 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+var currentSleeper sleeper = defaultSleeper{}
+
 // RetrySleep returns a pseudo-random sleep duration.
 func RetrySleep(base int, extra int) time.Duration {
```
zoltak-g, line 35: This is not that actionable in context of this PR, but I find this function to be pretty mysterious.

1) I would love to see a readable formula in comment (e.g `(b+e)*b + [0, b+e]`) instead of only examples. It took me a while to parse both examples and the line with actual expression. But maybe that's just me.

2) At first glance I have no intuition what numbers I should pass there, apart from looking the examples. I have no idea whether the original author had anything particular in mind or "just have a particular interval in mind and try to find two integers which more or less cover it"...  Ideally, the comment would build such intuition.

retryutil/retry_test.go
```diff
@@ -0,0 +1,244 @@
...
+		{base: 1, extra: 5, expectedLowerBound: 6 * time.Second, expectedUpperBound: 12 * time.Second},
+		{base: 2, extra: 5, expectedLowerBound: 14 * time.Second, expectedUpperBound: 21 * time.Second},
+		{base: 3, extra: 5, expectedLowerBound: 24 * time.Second, expectedUpperBound: 32 * time.Second},
+
+		{base: 1, extra: 10, expectedLowerBound: 11 * time.Second, expectedUpperBound: 22 * time.Second},
+		{base: 2, extra: 10, expectedLowerBound: 24 * time.Second, expectedUpperBound: 36 * time.Second},
+		{base: 3, extra: 10, expectedLowerBound: 39 * time.Second, expectedUpperBound: 52 * time.Second},
+	}
+
+	for _, tt := range tests {
```
zoltak-g, line 33: nit: maybe worth re-running each case N times (with N being like 100)?

retryutil/retry_test.go
```diff
@@ -0,0 +1,244 @@
...
+		expectedToFail       int
+		failWith             error
+		expectedError        error
+		funcCalledLowerBound int
+		funcCalledUpperBound int
+	}{
+		{
+			name:                 "Function does not fail",
+			maxRetryTime:         time.Minute,
+			callsCollector:       callsCollector,
```
zoltak-g, line 55: Do you have plans to use a different calls collector anywhere? If not - it is worth getting rid of it from the test case definition, as it only bloats the spec.

retryutil/retry_test.go
```diff
@@ -0,0 +1,244 @@
...
+			expectedToFail:       5,
+			failWith:             status.Error(codes.InvalidArgument, "invalid"),
+			expectedError:        fmt.Errorf("code: \"InvalidArgument\", message: \"invalid\", details: []"),
+			funcCalledLowerBound: 1,
+			funcCalledUpperBound: 1,
+		},
+		{
+			name:                 "Function fail with retriable API error, retry does not helped",
```
zoltak-g, line 146: nit: "help", same in next case

retryutil/retry_test.go
```diff
@@ -0,0 +1,244 @@
...
+			failWith:             nil,
+			expectedError:        nil,
+			funcCalledLowerBound: 1,
+			funcCalledUpperBound: 1,
+		},
+		{
+			name:                 "Function fail with non API error",
+			maxRetryTime:         time.Second,
+			callsCollector:       callsCollector,
+			expectedToFail:       5,
```
zoltak-g, line 129: I think I am missing sth here...

My current understanding of `callsCollector` is that it will start failing at `expectedToFail`-th call? So for this case it would return `nil` and would never be retried? (and therefore never fail?)

retryutil/retry_test.go
burov: Good point, i thought about this collector a bit differently when i started implementation, removed.

retryutil/retry_test.go
burov: Probably my naming judgment was not the best when i write it, expectedToFail is how many times it expected to fail before success

burov: I rename it to `expectedToFailTimes` that should give a bit more context

retryutil/retry_test.go
burov: I have nothing against it as well it potentially increase chances of catching issues on edge of the range,  done

retryutil/retry.go
burov: I agree the function is not very straightforward, i'll think of how to make comment for the function better.

zoltak-g (approved):

PR #799, opened 2025-03-17
Update hash computation algorithm

burov:
* Introduce new function that compute hash that independent from the order of inventory items.
* Write tests for computeHash and computeStableHash functions.
* Benchmark computeHash and computeStableHash to be able to compare approaches.

Benchmark results: 
goos: linux
goarch: amd64
pkg: github.com/GoogleCloudPlatform/osconfig/agentendpoint
cpu: Intel(R) Xeon(R) CPU @ 2.20GHz
Benchmark_computeHash-24          	   93759	     12064 ns/op
Benchmark_computeStableHash-24    	   17388	     68429 ns/op

### DIFF
--- a/agentendpoint/inventory.go
+++ b/agentendpoint/inventory.go
@@ -1,9 +1,14 @@
 package agentendpoint
 
 import (
+	"bytes"
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
+	"io"
 	"reflect"
+	"sort"
 	"strings"
 	"time"
 
@@ -13,6 +18,7 @@ import (
 	"github.com/GoogleCloudPlatform/osconfig/inventory"
 	"github.com/GoogleCloudPlatform/osconfig/packages"
 	"github.com/GoogleCloudPlatform/osconfig/retryutil"
+	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"cloud.google.com/go/osconfig/agentendpoint/apiv1/agentendpointpb"
@@ -358,4 +364,58 @@ func formatWindowsApplication(pkg *packages.WindowsApplication) *agentendpointpb
 			InstallDate:    &d,
 			HelpLink:       pkg.HelpLink,
 		}}
+}
+
+func computeFingerprint(ctx context.Context, inventory *agentendpointpb.Inventory) (string, error) {
+	fingerprint := sha256.New()
+	b, err := proto.Marshal(inventory)
+	if err != nil {
+		return "", err
+	}
+	io.Copy(fingerprint, bytes.NewReader(b))
+
+	return hex.EncodeToString(fingerprint.Sum(nil)), nil
+}
+
+func computeStableFingerprint(ctx context.Context, inventory *agentendpointpb.Inventory) (string, error) {
+	fingerprint := sha256.New()
+	b, err := proto.Marshal(inventory.GetOsInfo())
+	if err != nil {
+		return "", err
+	}
+	io.Copy(fingerprint, bytes.NewReader(b))
+
+	installedPackages := inventory.GetInstalledPackages()
+	availablePackages := inventory.GetAvailablePackages()
+
+	entries := make([]string, 0, len(installedPackages)+len(availablePackages))
+
+	for _, pkg := range installedPackages {
+		entries = append(entries, fingerprintForPackage(pkg))
+	}
+
+	for _, pkg := range availablePackages {
+		entries = append(entries, fingerprintForPackage(pkg))
+	}
+
+	sort.Strings(entries)
+
+	for _, entry := range entries {
+		if _, err := io.WriteString(fingerprint, entry); err != nil {
+			return "", err
+		}
+	}
+
+	return hex.EncodeToString(fingerprint.Sum(nil)), nil
+}
+
+func fingerprintForPackage(pkg *agentendpointpb.Inventory_SoftwarePackage) string {
+	//Inventory_WindowsUpdatePackage struct contains repeated fields
+	//we should rely on fields that are stable and sufficient enough to uniquely identify the package.
+	if wua := pkg.GetWuaPackage(); wua != nil {
+		return fmt.Sprintf("%s-%s-%d", wua.GetTitle(), wua.GetUpdateId(), wua.GetRevisionNumber())
+	}
+
+	// For all packages other then wua we can just rely on proto String() method.
+	return pkg.String()
 }
--- a/agentendpoint/inventory_test.go
+++ b/agentendpoint/inventory_test.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math/rand"
 	"net/http"
 	"net/http/httptest"
 	"reflect"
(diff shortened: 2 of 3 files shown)


agentendpoint/inventory.go
```diff
@@ -359,3 +365,46 @@ func formatWindowsApplication(pkg *packages.WindowsApplication) *agentendpointpb
...
+	b, err := proto.Marshal(inventory)
+	if err != nil {
+		return "", err
+	}
+	io.Copy(hash, bytes.NewReader(b))
+
+	return hex.EncodeToString(hash.Sum(nil)), nil
+}
+
+func computeStableHash(ctx context.Context, inventory *agentendpointpb.Inventory) (string, error) {
```
zoltak-g, line 380: I think we typically call them "fingerprints"

agentendpoint/inventory.go
```diff
@@ -359,3 +365,46 @@ func formatWindowsApplication(pkg *packages.WindowsApplication) *agentendpointpb
...
+	}
+	io.Copy(hash, bytes.NewReader(b))
+
+	installedPackages := inventory.GetInstalledPackages()
+	availablePackages := inventory.GetAvailablePackages()
+
+	entries := make([]string, 0, len(installedPackages)+len(availablePackages))
+
+	for _, pkg := range installedPackages {
+		entries = append(entries, pkg.String())
```
zoltak-g, line 394: I wonder if this is good enough without additional normalization of packages themselves?

For example - `WUAPackage` struct has multiple fields that are arrays. I do not think we can assume that order of items within those arrays is stable. There is also a (small) risk that we will introduce more array fields in the future, making this problem potentially worse.


This is likely an edge case, though. I think it is much less likely for a single package entry to not be stable while printed by package manager vs the overall order of packages. So maybe this is good enough improvement for now

agentendpoint/inventory_test.go
```diff
@@ -464,3 +465,87 @@ func TestReport(t *testing.T) {
 		})
 	}
 }
+
+func Test_computeHash(t *testing.T) {
+	ctx := context.Background()
+
+	expectedHash := "ebbafb112c34d2eed453bdf8748be7a8ae69ec05719b0a61618640f286754181"
```
zoltak-g, line 472: I wonder if this check makes much sense. Was this string mostly copy-pasted from the failed test output? It looks like an example of "change-detector test" not a logic test.

Maybe it would more sense to check things like
- whether the length / characters are as expected
- that change to the inventory changes the hash (in particular, same packages but in a different order, which would clearly show a functional difference from the stable hash alternative)

agentendpoint/inventory_test.go
```diff
@@ -464,3 +465,87 @@ func TestReport(t *testing.T) {
...
+			}
+
+			installedPackages := inventory.GetInstalledPackages()
+			availablePackages := inventory.GetAvailablePackages()
+
+			rand.Shuffle(len(installedPackages), func(i, j int) {
+				inventory.InstalledPackages[i], inventory.InstalledPackages[j] = inventory.InstalledPackages[j], inventory.InstalledPackages[i]
+			})
+
+			rand.Shuffle(len(availablePackages), func(i, j int) {
```
zoltak-g, line 506: How about creating a custom `Source` from a static seed, just in case this test starts to fail and someone has to reproduce it?

agentendpoint/inventory_test.go
```diff
@@ -464,3 +465,87 @@ func TestReport(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("unable to generate hash, err - %s", err)
+	}
+
+	if hash != expectedHash {
+		t.Fatalf("got unexpected hash, expected %q, got %q", expectedHash, hash)
+	}
+}
+
+func Test_computeStableHash_testHashStability(t *testing.T) {
```
zoltak-g, line 483: Maybe also a test, where you hash two different inventories and stable hashes are actually different?

I think right now 

```
func computeStableHash(ctx context.Context, inventory *agentendpointpb.Inventory) (string, error) {
  return "", nil
}
```

would pass tests

agentendpoint/inventory_test.go
zoltak-g: Hm... I see that this is to some extend tested later, but at the same time it is nice to still have short simple test cases just to explain contract.

agentendpoint/inventory.go
```diff
@@ -359,3 +365,46 @@ func formatWindowsApplication(pkg *packages.WindowsApplication) *agentendpointpb
 			HelpLink:       pkg.HelpLink,
 		}}
 }
+
+func computeHash(ctx context.Context, inventory *agentendpointpb.Inventory) (string, error) {
```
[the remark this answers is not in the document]
burov: I keep in for a bit of time and if we didn't get any complains we can remove this funciton.

agentendpoint/inventory.go
burov: Will rename, thank you!

agentendpoint/inventory_test.go
burov: Yeah, it actually looks like change detector test, let me rewrite it a bit

agentendpoint/inventory.go
burov: done

agentendpoint/inventory_test.go
burov: refactored

agentendpoint/inventory.go
burov: I changed algorithm a bit, for WUA package i just used compound key of (Title, UpdateId, RevisionId) that should be enough for all other items i continue to use pkg.String(), WDYT ?

agentendpoint/inventory.go
```diff
@@ -359,3 +365,57 @@ func formatWindowsApplication(pkg *packages.WindowsApplication) *agentendpointpb
...
+			return "", err
+		}
+	}
+
+	return hex.EncodeToString(fingerprint.Sum(nil)), nil
+}
+
+func fingerprintForPackage(pkg *agentendpointpb.Inventory_SoftwarePackage) string {
+	//Inventory_WindowsUpdatePackage struct contains repeated fields
+	//we should rely on fields that stable but enough to identify package
```
zoltak-g, line 414: I have trouble with parsing this.

Did you mean sth like  "we should rely on fields that are stable and sufficient to identify the package"?

agentendpoint/inventory_test.go
```diff
@@ -464,3 +484,164 @@ func TestReport(t *testing.T) {
...
+				t.Errorf("unexpected error while calculating fingerprint, err %v", err)
+			}
+			stableFingerprintBefore, err := computeStableFingerprint(ctx, inventory)
+			if err != nil {
+				t.Errorf("unexpected error while calculating fingerprint, err %v", err)
+			}
+
+			// Reorder pacakges in the inventory, that should change result of computeFingerprint
+			// and do not change the result of computeStableFingerptint.
+			reshufflePackagesOrder(inventory)
```
zoltak-g, line 591: One note - the `reshufflePackageOrder` is now a bit error prone, since it always re-initializes the `Rand` with same seed. 

For example - I can easily imagine someone using a copy of `inventory` here instead of re-shuffling the same object. Which would result in each iteration checking exactly the same permutation.


It is not critical, so I won't push on this, but I think ideally the Rand would be re-initialized per each ~test case.

agentendpoint/inventory.go
burov: Yeah sorry, i'll rewrite the comment

agentendpoint/inventory_test.go
burov: Done, WDYT ? i built infrastructure, to use time.Now(), but you can stick with static for debug

agentendpoint/inventory.go
burov: Done

zoltak-g (approved):

PR #805, opened 2025-04-05
Report RPM inventory as YUM instead of empty SoftwarePackage when neither Zypper nor YUM tools are installed.

1pnq:
User custom setup may have `rpm` package manager installed without `yum` (redhat family default) or `zypper` (suse family default). In setup like this, agent reports empty SoftwarePackage with no details for every `rpm` package to **agentendpoint**, as a result whole inventory reporting fails with `Unsupported software package.`.
Bug introduced in #386.

Current consensus is that reporting a non-updatable RPM package is better than ignoring it.
Both solutions (ignoring as well as reporting) would be better than failing whole inventory reporting.

[Reversion of #386 test override](https://github.com/GoogleCloudPlatform/osconfig/pull/386/files#diff-97c1483086e27560a7248457f5bf35daa0fb2eb8c42cdb512db692423b6c832bR399) makes this fix covered by existing test.

### DIFF
--- a/agentendpoint/inventory.go
+++ b/agentendpoint/inventory.go
@@ -159,13 +159,13 @@ func formatPackages(ctx context.Context, pkgs *packages.Packages, shortName stri
 	}
 	if pkgs.Rpm != nil {
 		temp := make([]*agentendpointpb.Inventory_SoftwarePackage, len(pkgs.Rpm))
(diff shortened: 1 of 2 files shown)


burov (approved):
Quite rare situation, so i think i'm good with such approach

PR #812, opened 2025-04-25
Snapshot ZypperPatches, ZypperUpdates (GetPackageUpdates) output for sles 12, 15 testdata

1pnq:


packages/zypper_test.go
```diff
@@ -90,28 +91,94 @@ this is junk data`
...
-	ret, err := ZypperUpdates(testCtx)
-	if err != nil {
-		t.Errorf("unexpected error: %v", err)
+		expectedCommandsChain []expectedCommand
+		expectedError         error
+		expectedResults       []*PkgInfo
+		expectedResultsFile   string
+	}{
+		{
+			name: "empty",
```
burov, line 104: NIT: could you name test cases in a form, input expected output? Just to make sure from the name i can fully understand the expected behavior

burov: I'm ready to approve right after that :)

1pnq: Good point, added descriptive names

burov (approved):
LGTM

PR #818, opened 2025-05-06
Improve test coverage for gem package manager

burov:


packages/gem.go
```diff
@@ -83,25 +83,44 @@ func InstalledGemPackages(ctx context.Context) ([]*PkgInfo, error) {
...
 	   foo (1.2.3, 1.2.4)
 	   bar (1.2.3)
+	   csv (default: 3.1.2)
+	   csv (default: 3.1.2, 1.5.8)
 	   ...
 	*/
 	lines := strings.Split(strings.TrimSpace(string(stdout)), "\n")
 
-	if len(lines) == 0 {
+	redundantLinesCount := 3
```
1pnq, line 92: Have you considered explicitly skipping empty line `""` and `"*** LOCAL GEMS ***"` lines from interpreting? (comparing to relying on magic offset `3`)

1pnq: (Or leaving existing logic as-is)

1pnq: Discussed in person

packages/gem.go
```diff
@@ -53,7 +53,7 @@ func GemUpdates(ctx context.Context) ([]*PkgInfo, error) {
 	*/
 
 	lines := strings.Split(strings.TrimSpace(string(stdout)), "\n")
-	if len(lines) == 0 {
+	if len(lines) == 0 || lines[0] == "" {
```
1pnq, line 56: if `len(lines) == 0` is never true, should we replace both with `strings.TrimSpace(string(stdout)) == ""`?

burov: good point, done

1pnq (approved):

PR #825, opened 2025-05-07
Refactor /inventory with dependency injection

burov:


inventory/inventory.go
```diff
@@ -41,6 +41,59 @@ type InstanceInventory struct {
...
+		OSConfigAgentVersion: agentconfig.Version(),
+		InstalledPackages:    installedPackages,
+		PackageUpdates:       packageUpdates,
+		LastUpdated:          time.Now().UTC().Format(time.RFC3339),
+	}
+}
+
+
 // Get generates inventory data.
 func Get(ctx context.Context) *InstanceInventory {
```
1pnq, line 98: unused

packages/packages.go
```diff
@@ -62,6 +62,26 @@ var (
 	ptyrunner = util.CommandRunner(&ptyRunner{})
 )
 
+type PackageUpdatesProvider interface {
+	GetPackageUpdates(context.Context) (*Packages, error)
```
1pnq, line 66: `Get`

packages/packages.go
```diff
@@ -62,6 +62,36 @@ var (
 	ptyrunner = util.CommandRunner(&ptyRunner{})
 )
 
+// PackageUpdatesProvider define contract to extract available updates from the VM.
```
1pnq, line 65: define -> defines

packages/packages.go
```diff
@@ -62,6 +62,36 @@ var (
...
+type PackageUpdatesProvider interface {
+	Get(context.Context) (*Packages, error)
+}
+
+// InstalledPackagesProvider define contract to extract installed packages from the VM.
```
1pnq, line 70: define -> defines

osinfo/osinfo.go
```diff
@@ -28,7 +28,24 @@ const (
...
+}
+
+// NewProvider returns fully function provider.
+func NewProvider() Provider {
+	return defaultProvider{}
+}
+
+type defaultProvider struct{}
+
+// Get extract return OSInfo for current platform.
```
1pnq, line 41: extract vs return

btw, can func body be replaced with `return Get()`?

1pnq (approved):

inventory/inventory.go
burov: done

packages/packages.go
burov: resuch approach to the naming will block single structure to implement all interfaces.

1pnq (approved):

PR #834, opened 2025-05-14
Introduce scalibr & extractor mappers for dpkg/rpm/cos os/filesystem extractors

1pnq:


packages/scalibr.go
```diff
@@ -0,0 +1,163 @@
...
+		} else if metadata, ok := pkg.Metadata.(*scalibrcos.Metadata); ok {
+			packages.COS = append(packages.COS, pkgInfoFromCosExtractorPackage(pkg, metadata, osinfo))
+		} else {
+			clog.Errorf(ctx, "Package type not implemented: %v", pkg)
+		}
+	}
+	return packages
+}
+
+func gatherScanRoots() ([]*scalibrfs.ScanRoot, error) {
```
burov, line 88: Speaking about scan roots, in the scenario of OS extractors only we don't need to do full file system scan, do you think that might be good idea to craft a set roots that will cover all of the currently supported extractors and does not trigger full file system scan.

packages/scalibr.go
```diff
@@ -0,0 +1,163 @@
...
+		return nil, err
+	}
+	var scanRoots []*scalibrfs.ScanRoot
+	for _, path := range scanRootPaths {
+		scanRoots = append(scanRoots, scalibrfs.RealFSScanRoot(path))
+	}
+	return scanRoots, nil
+}
+
+func gatherConfig() (*scalibr.ScanConfig, []error) {
```
burov, line 100: I think we can do better naming for this function, f.e. getDefaultConfig() might better represent the purpose of the function.

packages/scalibr.go
```diff
@@ -0,0 +1,163 @@
...
+		DirsToSkip:           dirsToSkip,
+	}, errs
+}
+
+type scalibrInstalledPackagesProvider struct {
+	osinfoProvider osinfo.Provider
+}
+
+// ScalibrInstalledPackagesProvider makes provider that uses osv-scalibr as its implementation.
+func ScalibrInstalledPackagesProvider(osinfoProvider osinfo.Provider) InstalledPackagesProvider {
```
burov, line 136: NewScalibrInstalledPackagesProvider would stick with current  naming

1pnq: Makes sense. Although I've sticked with `NewInstalledPackagesProvider` in packages_linux.go with config-based toggling for linux only as we discussed in person.

packages/scalibr.go
1pnq: Decided to go with `provider.getScanConfig()`

packages/scalibr.go
1pnq: Yes — roots mangling could be used as a workaround to avoid full filesystem scan. However, as we discussed, probably making these scanners standalone would be more appropriate fix if performance matters.

I've changed scalibr provider structure to allow roots modifications for integration test, see `TestScalibrIntegration`

packages/scalibr.go
```diff
@@ -0,0 +1,163 @@
...
+func gatherConfig() (*scalibr.ScanConfig, []error) {
+	var errs []error
+	extractors := []string{
+		"os/dpkg",
+		"os/rpm",
+		"os/cos",
+		// TODO: implement "os/zypper" for `zypper patches` — excluded from scan till then
+	}
+
+	filesystemExtractors, err := fslist.ExtractorsFromNames(extractors)
```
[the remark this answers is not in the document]
1pnq: Adjusted — now we fail fast

packages/packages_linux.go
```diff
@@ -177,3 +179,20 @@ func getInstalledPackages(ctx context.Context) (Packages, []string) {
 
 	return pkgs, errs
 }
+
+// NewInstalledPackagesProvider makes provider that uses osv-scalibr as its implementation if enabled by config, otherwise falls back to default legacy implementation.
+func NewInstalledPackagesProvider(osinfoProvider osinfo.Provider) InstalledPackagesProvider {
```
burov, line 184: There small problem that current scenario does not handle, my assumption that inventory extraction logic should not be changed after the scalibr enablement, but in this scenario if SCALIBR enabled i assume we will trigger only 3 currently supported SCALIBR extractors, so we will broke current logic.

packages/scalibr.go
```diff
@@ -0,0 +1,163 @@
+package packages
```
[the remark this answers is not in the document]
burov: Which cycles we have right now, could you be a bit more specific ?

packages/scalibr_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/google/go-cmp/cmp"
+	scalibr "github.com/google/osv-scalibr"
+	"github.com/google/osv-scalibr/extractor"
+	scalibrcos "github.com/google/osv-scalibr/extractor/filesystem/os/cos"
+	scalibrdpkg "github.com/google/osv-scalibr/extractor/filesystem/os/dpkg"
+	scalibrrpm "github.com/google/osv-scalibr/extractor/filesystem/os/rpm"
+	"github.com/google/osv-scalibr/inventory"
+)
+
+func TestExtractedPackageMappings(t *testing.T) {
```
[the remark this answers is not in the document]
1pnq: see `TestScalibrIntegration`

agentconfig/agentconfig.go
```diff
@@ -351,12 +353,24 @@ func createConfigFromMetadata(md metadataJSON) *config {
...
 	}
 
+	setInventoryExtractor(md, c)
 	setSVCEndpoint(md, c)
 	setTraceGetInventory(md, c)
 
 	return c
 }
 
+func setInventoryExtractor(md metadataJSON, c *config) {
```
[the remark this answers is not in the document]
1pnq: Updated (single flag scalibrLinux for now)

packages/scalibr.go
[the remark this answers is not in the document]
1pnq: Potential `scalibr` package depends on `packages[.Packages, .PkgInfo, .Source], InstalledPackagesProvider`, so we either move these structs to third package (doubtful), or move InstalledPackagesProvider instantiation to separate package (possibly inline near `inventory.NewProvider`) to avoid cycle like
```
package github.com/GoogleCloudPlatform/osconfig
        imports github.com/GoogleCloudPlatform/osconfig/agentendpoint from main.go
        imports github.com/GoogleCloudPlatform/osconfig/config from config_task.go
        imports github.com/GoogleCloudPlatform/osconfig/packages from package_resource.go
        imports github.com/GoogleCloudPlatform/osconfig/scalibr from packages_linux.go
        imports github.com/GoogleCloudPlatform/osconfig/packages from scalibr.go: import cycle not allowed
```

At this moment we use packages_linux vs packages_windows conditional compilation to turn on scalibr on linux only before we have windows tested, so that's quite handy to keep it in a single package for a while. We can fall back to  `runtime.GOOS == "linux"` as well. What I'm saying is that slicing scalibr parts from packages could be good next step and current PoC is good enough for probers run.

burov (approved):

PR #835, opened 2025-05-16
Trace GetInstalledPackages memory levels

1pnq:
when "trace-get-inventory" metadata flag is enabled

### DIFF
--- a/inventory/inventory.go
+++ b/inventory/inventory.go
@@ -73,7 +73,10 @@ type defaultInventoryProvider struct {
 func NewProvider() Provider {
 	installedPackagesProvider := packages.NewInstalledPackagesProvider()
 	if agentconfig.TraceGetInventory() {
-		installedPackagesProvider = packages.TracingInstalledPackagesProvider(installedPackagesProvider, osinfo.NewProvider())
+		installedPackagesProvider = packages.TracingInstalledPackagesProvider(
+			installedPackagesProvider,
+			osinfo.NewProvider(),
+		)
 	}
 	return &defaultInventoryProvider{
 		osInfoProvider:            osinfo.NewProvider(),
--- a/inventory/inventory.go
+++ b/inventory/inventory.go
@@ -73,7 +73,7 @@ type defaultInventoryProvider struct {
 func NewProvider() Provider {
 	installedPackagesProvider := packages.NewInstalledPackagesProvider()
 	if agentconfig.TraceGetInventory() {
-		installedPackagesProvider = packages.TracingInstalledPackagesProvider(context.Background(), installedPackagesProvider, osinfo.NewProvider())
+		installedPackagesProvider = packages.TracingInstalledPackagesProvider(installedPackagesProvider, osinfo.NewProvider())
 	}
 	return &defaultInventoryProvider{
 		osInfoProvider:            osinfo.NewProvider(),
--- a/inventory/inventory.go
+++ b/inventory/inventory.go
@@ -24,6 +24,7 @@ import (
 	"github.com/GoogleCloudPlatform/osconfig/clog"
 	"github.com/GoogleCloudPlatform/osconfig/osinfo"
 	"github.com/GoogleCloudPlatform/osconfig/packages"
+	"github.com/GoogleCloudPlatform/osconfig/util/utiltrace"
 )
 
 // InstanceInventory is an instances inventory data.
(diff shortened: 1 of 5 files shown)


inventory/inventory.go
```diff
@@ -71,10 +72,31 @@ type defaultInventoryProvider struct {
 
 // NewProvider returns ready to work default provider
 func NewProvider() Provider {
+	osInfoProvider := osinfo.NewProvider()
+	installedPackagesProvider := packages.NewInstalledPackagesProvider()
+	if agentconfig.TraceGetInventory() {
+		ctx := context.Background()
+		installedPackagesProvider = packages.TracingInstalledPackagesProvider(installedPackagesProvider, func(stats utiltrace.TraceMemoryResult, duration time.Duration) {
+			osinfo, _ := osInfoProvider.GetOSInfo(ctx)
```
burov: This function become overloaded, could you move log construction out from this function ? and Please use DEBUG level log

packages/trace.go
```diff
@@ -0,0 +1,38 @@
...
+type HandleTraceResult func(stats utiltrace.TraceMemoryResult, duration time.Duration)
+
+func TracingInstalledPackagesProvider(provider InstalledPackagesProvider, handleTraceResult HandleTraceResult) InstalledPackagesProvider {
+	return tracingInstalledPackagesProvider{
+		provider:          provider,
+		handleTraceResult: handleTraceResult,
+	}
+}
+
+func (p tracingInstalledPackagesProvider) GetInstalledPackages(ctx context.Context) (Packages, error) {
```
burov, line 22: Can we move logic related to the time measuring to the utiltrace.Trace or smth ? This function should be easy wrapper rather than

packages/trace.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/GoogleCloudPlatform/osconfig/util/utiltrace"
+)
+
+type tracingInstalledPackagesProvider struct {
+	provider          InstalledPackagesProvider
+	handleTraceResult HandleTraceResult
```
burov: Why do we need this as a property ? Are we expect to override this ?

util/utiltrace/utiltrace.go
```diff
@@ -0,0 +1,62 @@
...
+	MemBeforeMB float64
+	MemAfterMB  float64
+	MemPeakMB   float64
+	MemMeanMB   float64
+	SampleCount int
+}
+
+// TraceMemory collects memory usage with specified interval until done channel is closed
+func TraceMemory(ctx context.Context, interval time.Duration, resultChannel chan TraceMemoryResult) {
+	compactMemory()
```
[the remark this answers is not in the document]
1pnq: We call it twice when measuring beforeMem/afterMem to get somewhat representable/comparable numbers to see whether there's any growing trend. Flukes may arise, and that's what mean/peak memory consumption metrics is for, not affected by GC.

util/utiltrace/utiltrace.go
```diff
@@ -0,0 +1,62 @@
...
+type TraceMemoryResult struct {
+	MemBeforeMB float64
+	MemAfterMB  float64
+	MemPeakMB   float64
+	MemMeanMB   float64
+	SampleCount int
+}
+
+// TraceMemory collects memory usage with specified interval until done channel is closed
+func TraceMemory(ctx context.Context, interval time.Duration, resultChannel chan TraceMemoryResult) {
```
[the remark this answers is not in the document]
1pnq: That would be good idea to simplify consumer (yet single one). However that would require TraceMemory to call measured function itself, and I decided to avoid overgeneralization/using generics here.

packages/trace.go
1pnq: Time measuring logic consists of two lines. The rest is TraceMemory call/result gathering and logging. Do you see any meaningful interface for `utiltrace.Trace` that I don't?
We could do it in composable way only if Trace & TraceMemory would call measured function itself.

packages/trace.go
1pnq: (Outdated code fragment)

inventory/inventory.go
```diff
@@ -71,10 +71,15 @@ type defaultInventoryProvider struct {
 
 // NewProvider returns ready to work default provider
 func NewProvider() Provider {
+	osInfoProvider := osinfo.NewProvider()
```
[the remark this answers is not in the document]
1pnq: Done

inventory/inventory.go
1pnq: However (let's discuss in person) I'd still prefer to do it on Info level (it is toggled by metadata and we don't want to test debug logging perf/memory penalty)

packages/trace.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/GoogleCloudPlatform/osconfig/util/utiltrace"
+)
+
+type tracingInstalledPackagesProvider struct {
+	tracedProvider InstalledPackagesProvider
+	osInfoProvider osinfo.Provider
+}
+
+// TracingInstalledPackagesProvider creates an InstalledPackagesProvider decorator that traces the execution time, memory usage, and OS information for each call.
+func TracingInstalledPackagesProvider(ctx context.Context, tracedProvider InstalledPackagesProvider, osInfoProvider osinfo.Provider) InstalledPackagesProvider {
```
burov, line 18: ctx is redundant

packages/trace.go
```diff
@@ -0,0 +1,57 @@
...
+	return tracingInstalledPackagesProvider{tracedProvider: tracedProvider, osInfoProvider: osInfoProvider}
+}
+
+func (p tracingInstalledPackagesProvider) GetInstalledPackages(ctx context.Context) (Packages, error) {
+	traceCtx, cancel := context.WithCancel(ctx)
```
burov, line 23: call it just ctx

packages/trace.go
```diff
@@ -0,0 +1,57 @@
...
+		ctx,
+		"GetInstalledPackages: %.3fs, memory %+.2f MB (=%.2f-%.2f), peak %.2f MB, mean %.2f MB (%d samples), OS: %s@%s",
+		duration.Seconds(),
+		result.MemAfterMB-result.MemBeforeMB,
+		result.MemAfterMB,
+		result.MemBeforeMB,
+		result.MemPeakMB,
+		result.MemMeanMB,
+		result.SampleCount,
+		osinfo.ShortName,
```
burov, line 54: Maybe hostname might be useful

packages/trace.go
1pnq: Discussed in person — seems current separation of concerns is legit

burov (approved):

PR #846, opened 2025-07-17
Make OS Config agent TPC aware

petercieslak:
Changes include:

- Disabling Cloud Logging if agent works in a TPC
- Changing SvcEndpoint host according to the universe domain


agentconfig/agentconfig.go
```diff
@@ -82,6 +82,9 @@ const (
 
 	osConfigPollIntervalDefault = 10
 	osConfigMetadataPollTimeout = 60
+
+	//GDU domain
```
burov, line 86: please avoid abbreviations, // Default Google API domain will be slightly better

burov (commented):
Regarding clients, i see that we updated OS Config Agent Client creation, but still create some of the clients without explicitly specify domain. For example Cloud Storage client.

burov (approved):
LGTM

PR #925, opened 2026-03-12
Cover agentconfig functionality by unit tests

iliatsuprik:
This PR contains unit tests and minor changes in code for github.com/GoogleCloudPlatform/osconfig/agentconfig package.
Initial coverage: 58.1%, current coverage: 96.4%

Changes in the code of main app:
- move 5 seconds retry interval and runtime.GOOS into global variable. That's needed for mocking it in tests

### DIFF
--- a/agentconfig/agentconfig_test.go
+++ b/agentconfig/agentconfig_test.go
@@ -30,7 +30,6 @@ import (
 	"reflect"
 	"runtime"
 	"strings"
-	"sync"
 	"testing"
 	"time"
 
@@ -55,35 +54,90 @@ func TestWatchConfig(t *testing.T) {
 		t.Fatalf("Error running WatchConfig: %v", err)
 	}
 
-	testsBool := []struct {
+	tests := []struct {
 		name string
-		op   any
+		op   func() any
 		want any
 	}{
-		{name: "metadata endpoint is SvcEndpoint, returns configured service endpoint", op: SvcEndpoint, want: "SvcEndpoint"},
-		{name: "metadata zone and name are populated, returns instance resource path", op: Instance, want: "zone/instances/name"},
-		{name: "metadata instance id is 12345, returns instance id string", op: ID, want: "12345"},
-		{name: "metadata project id is projectId, returns project id", op: ProjectID, want: "projectId"},
-		{name: "metadata zone is zone, returns zone", op: Zone, want: "zone"},
-		{name: "metadata instance name is name, returns instance name", op: Name, want: "name"},
-		{name: "project disables inventory and instance enables it, returns inventory enabled", op: OSInventoryEnabled, want: true},
-		{name: "instance enables prerelease tasks, returns task notifications enabled", op: TaskNotificationEnabled, want: true},
-		{name: "instance enables prerelease ospatch, returns guest policies enabled", op: GuestPoliciesEnabled, want: true},
-		{name: "project disables debug and instance enables it, returns debug enabled", op: Debug, want: true},
-		{name: "instance enables scalibr linux, returns scalibr linux enabled", op: ScalibrLinuxEnabled, want: true},
-		{name: "instance enables trace get inventory, returns inventory tracing enabled", op: TraceGetInventory, want: true},
-		{name: "instance enables guest attributes, returns guest attributes enabled", op: GuestAttributesEnabled, want: true},
-		{name: "svc poll interval is 3 minutes, returns proper time", op: SvcPollInterval, want: 3 * time.Minute},
-		{name: "numeric project id is 12345, successfuly returned", op: NumericProjectID, want: int64(12345)},
+		{
+			name: "metadata endpoint is SvcEndpoint, returns configured service endpoint",
+			op:   asAny(SvcEndpoint),
+			want: "SvcEndpoint",
+		},
+		{
+			name: "metadata zone and name are populated, returns instance resource path",
+			op:   asAny(Instance),
+			want: "zone/instances/name",
+		},
+		{
+			name: "metadata instance id is 12345, returns instance id string",
+			op:   asAny(ID),
+			want: "12345",
+		},
+		{
+			name: "metadata project id is projectId, returns project id",
+			op:   asAny(ProjectID),
+			want: "projectId",
+		},
+		{
+			name: "metadata zone is zone, returns zone",
+			op:   asAny(Zone),
+			want: "zone",
+		},
+		{
+			name: "metadata instance name is name, returns instance name",
+			op:   asAny(Name),
+			want: "name",
+		},
+		{
+			name: "project disables inventory and instance enables it, returns inventory enabled",
+			op:   asAny(OSInventoryEnabled),
+			want: true,
+		},
+		{
+			name: "instance enables prerelease tasks, returns task notifications enabled",
+			op:   asAny(TaskNotificationEnabled),
+			want: true,
+		},
+		{
+			name: "instance enables prerelease ospatch, returns guest policies enabled",
+			op:   asAny(GuestPoliciesEnabled),
+			want: true,
+		},
+		{
+			name: "project disables debug and instance enables it, returns debug enabled",
+			op:   asAny(Debug),
+			want: true,
+		},
+		{
+			name: "instance enables scalibr linux, returns scalibr linux enabled",
+			op:   asAny(ScalibrLinuxEnabled),
+			want: true,
+		},
+		{
+			name: "instance enables trace get inventory, returns inventory tracing enabled",
+			op:   asAny(TraceGetInventory),
+			want: true,
+		},
+		{
+			name: "instance enables guest attributes, returns guest attributes enabled",
+			op:   asAny(GuestAttributesEnabled),
+			want: true,
+		},
+		{
+			name: "svc poll interval is 3 minutes, returns proper time",
+			op:   asAny(SvcPollInterval),
+			want: 3 * time.Minute,
+		},
+		{
+			name: "numeric project id is 12345, successfuly returned",
+			op:   asAny(NumericProjectID),
+			want: int64(12345),
+		},
 	}
-	for _, tt := range testsBool {
+	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			results := reflect.ValueOf(tt.op).Call(nil)
-			if len(results) == 0 {
-				t.Fatalf("Function %v did not return any value", tt.op)
-			}
-			got := results[0].Interface()
-			utiltest.AssertEquals(t, got, tt.want)
+			utiltest.AssertEquals(t, tt.op(), tt.want)
 		})
 	}
 }
@@ -107,44 +161,65 @@ func TestSetConfigEnabled(t *testing.T) {
 		}
 	})
 
-	for i, want := range []bool{false, true, false} {
-		request = i
-		if err := WatchConfig(context.Background()); err != nil {
-			t.Fatalf("Error running SetConfig: %v", err)
-		}
-
-		testsBool := []struct {
-			name string
-			op   func() bool
-		}{
-			{name: "enable-osconfig metadata is applied to inventory state, returns expected inventory flag", op: OSInventoryEnabled},
-			{name: "enable-osconfig metadata is applied to task state, returns expected task notification flag", op: TaskNotificationEnabled},
-			{name: "enable-osconfig metadata is applied to guest policy state, returns expected guest policy flag", op: GuestPoliciesEnabled},
-		}
-		for _, tt := range testsBool {
-			t.Run(fmt.Sprintf("request %d: %s", request, tt.name), func(t *testing.T) {
-				utiltest.AssertEquals(t, tt.op(), want)
-			})
-		}
-	}
-
-	request = 3
-	if err := WatchConfig(context.Background()); err != nil {
-		t.Fatalf("Error running SetConfig: %v", err)
-	}
-
-	testsBool := []struct {
+	type assertion struct {
 		name string
 		op   func() bool
 		want bool
-	}{
-		{name: "disabled features contains osinventory, returns inventory disabled", op: OSInventoryEnabled, want: false},
-		{name: "osconfig remains enabled for tasks, returns task notifications enabled", op: TaskNotificationEnabled, want: true},
-		{name: "osconfig remains enabled for guest policies, returns guest policies enabled", op: GuestPoliciesEnabled, want: true},
 	}
-	for _, tt := range testsBool {
-		t.Run(tt.name, func(t *testing.T) {
-			utiltest.AssertEquals(t, tt.op(), tt.want)
+	tests := []struct {
+		name       string
+		request    int
+		assertions []assertion
+	}{
+		{
+			name:    "project and instance disable osconfig, returns features disabled",
+			request: 0,
+			assertions: []assertion{
+				{name: "inventory is requested, returns disabled", op: OSInventoryEnabled, want: false},
+				{name: "task notifications are requested, returns disabled", op: TaskNotificationEnabled, want: false},
+				{name: "guest policies are requested, returns disabled", op: GuestPoliciesEnabled, want: false},
+			},
+		},
+		{
+			name:    "project disables osconfig and instance enables osconfig, returns features enabled",
+			request: 1,
+			assertions: []assertion{
+				{name: "inventory is requested, returns enabled", op: OSInventoryEnabled, want: true},
+				{name: "task notifications are requested, returns enabled", op: TaskNotificationEnabled, want: true},
+				{name: "guest policies are requested, returns enabled", op: GuestPoliciesEnabled, want: true},
+			},
+		},
+		{
+			name:    "project and instance disable osconfig again, returns features disabled",
+			request: 2,
+			assertions: []assertion{
+				{name: "inventory is requested, returns disabled", op: OSInventoryEnabled, want: false},
+				{name: "task notifications are requested, returns disabled", op: TaskNotificationEnabled, want: false},
+				{name: "guest policies are requested, returns disabled", op: GuestPoliciesEnabled, want: false},
+			},
+		},
+		{
+			name:    "osconfig enabled and disabled features contains osinventory, returns inventory disabled only",
+			request: 3,
+			assertions: []assertion{
+				{name: "inventory is requested, returns disabled", op: OSInventoryEnabled, want: false},
+				{name: "task notifications are requested, returns enabled", op: TaskNotificationEnabled, want: true},
+				{name: "guest policies are requested, returns enabled", op: GuestPoliciesEnabled, want: true},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(fmt.Sprintf("request %d: %s", tt.request, tt.name), func(t *testing.T) {
+			request = tt.request
+			if err := WatchConfig(context.Background()); err != nil {
+				t.Fatalf("Error running SetConfig: %v", err)
+			}
+
+			for _, assertion := range tt.assertions {
+				t.Run(assertion.name, func(t *testing.T) {
+					utiltest.AssertEquals(t, assertion.op(), assertion.want)
+				})
+			}
 		})
 	}
 }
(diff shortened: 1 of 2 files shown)


burov (commented):
I think we might benefit from using table driven testing more often in this CL, as well i see a lot of code duplication in test set-up, some of the code repetitions are low hanging fruits and can be just extracted to separate function as is.

agentconfig/agentconfig_test.go
```diff
@@ -212,6 +229,129 @@ func TestSetConfigDefaultValues(t *testing.T) {
...
+// TestWatchConfigUnchangedConfigTimeout tests how the agent behaves when it receives
+// updates from the metadata server, but the actual configuration data hasn't changed.
+//
+// The agent checks the SHA256 hash of the new data. If the hash is identical to
+// the current configuration, it knows the update is superficial. Instead of
+// applying the configuration and exiting, the agent should ignore the update and
+// keep polling for real changes. This test verifies that the agent correctly
+// continues to wait until its internal timeout runs out, and then exits normally.
+func TestWatchConfigUnchangedConfigTimeout(t *testing.T) {
+	defer OverrideWatchConfigTimeouts(1*time.Millisecond, 10*time.Millisecond)()
```
burov: not critical, but changing it 

from 
```
defer OverrideWatchConfigTimeouts(1*time.Millisecond, 10*time.Millisecond)()
```
to 
```
rollback := OverrideWatchConfigTimeouts(1*time.Millisecond, 10*time.Millisecond)
defer rollback()
```

Improve readability of the code

petercieslak (dismissed):

agentconfig/agentconfig_test.go
```diff
@@ -16,81 +16,90 @@ package agentconfig
...
-	if NumericProjectID() != 12345 {
-		t.Errorf("NumericProjectID: got(%v) != want(%d)", NumericProjectID(), 12345)
-	}
-
-	if Instance() != "zone/instances/name" {
-		t.Errorf("zone: got(%s) != want(%s)", Instance(), "zone/instances/name")
-	}
+	utiltest.AssertEquals(t, SvcPollInterval().Minutes(), float64(3))
+	utiltest.AssertEquals(t, NumericProjectID(), int64(12345))
+	utiltest.AssertEquals(t, Instance(), "zone/instances/name")
```
burov: Why don't we put it to the table tests ? I suppose if we change parameter type to `any` we can simply add types of the variables to table.

iliatsuprik: done

agentconfig/agentconfig_test.go
iliatsuprik: this seems to be outdated. Now we use utiltest.OverrideVariable here

agentconfig/agentconfig_test.go
```diff
@@ -16,81 +16,81 @@ package agentconfig
...
-			t.Errorf("%q: got(%q) != want(%q)", tt.desc, tt.op(), tt.want)
-		}
-	}
-
 	testsBool := []struct {
-		desc string
-		op   func() bool
-		want bool
+		name string
+		op   any
```
burov: We can avoid reflection usage to call function if w make it `func() any`

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +134,721 @@ func TestSetConfigEnabled(t *testing.T) {
...
-	if err := os.Setenv("GCE_METADATA_HOST", strings.Trim(ts.URL, "http://")); err != nil {
-		t.Fatalf("Error running os.Setenv: %v", err)
-	}
+	})
 
 	if err := WatchConfig(context.Background()); err != nil {
 		t.Fatalf("Error running SetConfig: %v", err)
 	}
 
 	testsString := []struct {
```
burov: As i mentioned if we change the structure of test to

struct{
 op func() any
 want any 
}

we can stop repeating ourselves by generilise test casees

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +134,721 @@ func TestSetConfigEnabled(t *testing.T) {
...
+	setupMockMetadataServer(t, func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Etag", fmt.Sprintf("unmarshal-error-etag-%d", time.Now().UnixNano()))
+		w.Header().Set("Metadata-Flavor", "Google")
+		w.Write(badJSON)
+	})
+
+	err := WatchConfig(context.Background())
+
+	var dummy metadataJSON
+	expectedErr := json.Unmarshal(badJSON, &dummy)
```
petercieslak: is it possible to hardcode expectedErr here? Similarly to how it's done in line 255

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +134,721 @@ func TestSetConfigEnabled(t *testing.T) {
...
+	utiltest.AssertErrorMatch(t, WatchConfig(ctx), nil)
+}
+
+func TestSetConfigError(t *testing.T) {
+	setupMockMetadataServer(t, func(w http.ResponseWriter, r *http.Request) {})
+	utiltest.OverrideVariable(t, &osConfigWatchConfigTimeout, 1*time.Millisecond)
+
+	err := WatchConfig(context.Background())
+	var dummy metadataJSON
+	expectedErr := json.Unmarshal([]byte{}, &dummy)
```
petercieslak: same as above

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +134,721 @@ func TestSetConfigEnabled(t *testing.T) {
...
+		fmt.Fprint(w, `{"bad json"`) // Trigger unmarshal error loop which checks context
+	})
+
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel() // Cancel the context immediately prior to passing it in
+
+	utiltest.AssertErrorMatch(t, WatchConfig(ctx), nil)
+}
+
+func TestSetConfigError(t *testing.T) {
```
petercieslak, line 438: Could you add a small comment explaining what does this test do?

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +134,721 @@ func TestSetConfigEnabled(t *testing.T) {
...
+	utiltest.AssertEquals(t, SvcPollInterval().Minutes(), float64(osConfigPollIntervalDefault))
 
 	expectedEndpoint := "fake-zone-osconfig.googleapis.com.:443"
-	if SvcEndpoint() != expectedEndpoint {
-		t.Errorf("Default endpoint: got(%s) != want(%s)", SvcEndpoint(), expectedEndpoint)
+	utiltest.AssertEquals(t, SvcEndpoint(), expectedEndpoint)
+}
+
+// TestWatchConfigUnchangedConfigTimeout ignores unchanged metadata until timeout.
+func TestWatchConfigUnchangedConfigTimeout(t *testing.T) {
```
petercieslak, line 347: Correct me if I'm wrong but my understanding is that we should test here that the agent config has not changed even though we are sending many requests until timeout. In my opinion this test lacks assertion that the config did not change - can you try and use getAgentConfig() to make this check? Also it would probably be good to check if the `count` variable is bigger than 1 to verify that we make couple of calls.

agentconfig/agentconfig_test.go
```diff
@@ -282,22 +882,315 @@ func TestDisableCloudLogging(t *testing.T) {
...
+	}{
+		{input: "true", want: true},
+		{input: "1", want: true},
+		{input: "false", want: false},
+		{input: "0", want: false},
+		{input: "invalid", want: false},
+	}
+
+	for _, tt := range tests {
+		if got := parseBool(tt.input); got != tt.want {
```
petercieslak: can we use utiltest.AssertEquals() for consistency here?

agentconfig/agentconfig_test.go
iliatsuprik: here we need an error of type json.SyntaxError, but it has unexported field msg and test code cannot set it. If you know how to handle it better, I can replace

agentconfig/agentconfig_test.go
iliatsuprik: the same issue, we cannot build json.SyntaxError with proper message here

agentconfig/agentconfig_test.go
iliatsuprik: done

agentconfig/agentconfig_test.go
iliatsuprik: I've added a count check and assertion that config is unchanged

agentconfig/agentconfig_test.go
iliatsuprik: done

agentconfig/agentconfig_test.go
```diff
@@ -16,81 +16,81 @@ package agentconfig
...
-	}
-
-	if Instance() != "zone/instances/name" {
-		t.Errorf("zone: got(%s) != want(%s)", Instance(), "zone/instances/name")
+		t.Run(tt.name, func(t *testing.T) {
+			results := reflect.ValueOf(tt.op).Call(nil)
+			if len(results) == 0 {
+				t.Fatalf("Function %v did not return any value", tt.op)
+			}
+			got := results[0].Interface()
```
burov: If we replace just any with func() any we can get rid of this logic.

agentconfig/agentconfig_test.go
```diff
@@ -119,17 +114,17 @@ func TestSetConfigEnabled(t *testing.T) {
...
 
 		testsBool := []struct {
-			desc string
+			name string
 			op   func() bool
 		}{
-			{"OSInventoryEnabled", OSInventoryEnabled},
-			{"TaskNotificationEnabled", TaskNotificationEnabled},
-			{"GuestPoliciesEnabled", GuestPoliciesEnabled},
+			{name: "enable-osconfig metadata is applied to inventory state, returns expected inventory flag", op: OSInventoryEnabled},
```
burov: As name is quite long, can we put each field on the new line?

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +134,730 @@ func TestSetConfigEnabled(t *testing.T) {
...
+		fmt.Fprint(w, `{}`)
+	})
+
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+
+	err := WatchConfig(ctx)
+	utiltest.AssertErrorMatch(t, err, nil)
+	utiltest.AssertErrorMatch(t, ctx.Err(), nil)
+	if got := getAgentConfig(); got != before {
```
burov, line 368: Can we stick with usage of utiltest functions ?

petercieslak (dismissed):

agentconfig/agentconfig_test.go
```diff
@@ -282,22 +891,313 @@ func TestDisableCloudLogging(t *testing.T) {
...
+// TestFlagsAndEnvVars parses environment-backed flags.
+func TestFlagsAndEnvVars(t *testing.T) {
+	tests := []struct {
+		name                  string
+		freeOSMemoryVal       string
+		disableInventoryWrite string
+		wantFreeOS            bool
+		wantDisableInv        bool
+	}{
+		{name: "environment enables both flags, returns both flags enabled", freeOSMemoryVal: "true", disableInventoryWrite: "1", wantFreeOS: true, wantDisableInv: true},
```
burov: Names of the testcase is too long, let put each field on the new line in such scenarios

agentconfig/agentconfig_test.go
```diff
@@ -282,22 +891,313 @@ func TestDisableCloudLogging(t *testing.T) {
...
+		name  string
+		setup func(t *testing.T)
+		want  func(t *testing.T) string
+	}{
+		{
+			name: "xdg cache home is set, returns cache path under xdg cache home",
+			setup: func(t *testing.T) {
+				t.Setenv("XDG_CACHE_HOME", filepath.Join(t.TempDir(), "xdg-cache"))
+			},
+			want: func(t *testing.T) string {
```
burov, line 1233: What the reason here to have function here instead of string? i suppose in all of the scenarios we can just compute the value.

agentconfig/agentconfig_test.go
```diff
@@ -282,22 +891,313 @@ func TestDisableCloudLogging(t *testing.T) {
...
+
+// TestSetTraceGetInventory applies metadata precedence for inventory tracing.
+func TestSetTraceGetInventory(t *testing.T) {
+	tests := []struct {
+		name    string
+		projVal string
+		instVal string
+		want    bool
+	}{
+		{name: "project and instance values are empty, returns trace get inventory disabled", projVal: "", instVal: "", want: false},
```
burov: For the long testcase names, please put each field on the separate line

agentconfig/agentconfig_test.go
iliatsuprik: we cannot use it have as agent config has unexported fields (so calling cmp.Diff will lead to panic)

agentconfig/agentconfig_test.go
iliatsuprik: done

agentconfig/agentconfig_test.go
iliatsuprik: fixed

agentconfig/agentconfig_test.go
iliatsuprik: fixed

agentconfig/agentconfig_test.go
iliatsuprik: fixed

agentconfig/agentconfig_test.go
iliatsuprik: fixed

agentconfig/agentconfig_test.go
iliatsuprik: fixed with function wrapper

agentconfig/agentconfig_test.go
iliatsuprik: fixed with function wrapper

agentconfig/agentconfig_test.go
```diff
@@ -16,81 +16,135 @@ package agentconfig
...
-		{"Zone", Zone, "zone"},
-		{"Name", Name, "name"},
-	}
-	for _, tt := range testsString {
-		if tt.op() != tt.want {
-			t.Errorf("%q: got(%q) != want(%q)", tt.desc, tt.op(), tt.want)
-		}
-	}
-
 	testsBool := []struct {
```
burov: Please rename the variable as it is now handle not only bool cases

agentconfig/agentconfig_test.go
```diff
@@ -105,12 +159,7 @@ func TestSetConfigEnabled(t *testing.T) {
...
 		}
-	}))
-	defer ts.Close()
-
-	if err := os.Setenv("GCE_METADATA_HOST", strings.Trim(ts.URL, "http://")); err != nil {
-		t.Fatalf("Error running os.Setenv: %v", err)
-	}
+	})
 
 	for i, want := range []bool{false, true, false} {
```
burov: I suppose we can avoid this loop by just moving `want` and `request` to the tests structure.

agentconfig/agentconfig_test.go
```diff
@@ -119,17 +168,26 @@ func TestSetConfigEnabled(t *testing.T) {
 		}
 
 		testsBool := []struct {
```
burov: I suppose we can rename it to just tests

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +197,802 @@ func TestSetConfigEnabled(t *testing.T) {
 	}
 
 	testsBool := []struct {
```
burov: I suppose two types of the tests can be combined if we stick with the previous comments

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +197,802 @@ func TestSetConfigEnabled(t *testing.T) {
...
-	}
+	})
 
 	if err := WatchConfig(context.Background()); err != nil {
 		t.Fatalf("Error running SetConfig: %v", err)
 	}
 
 	testsString := []struct {
+		name string
 		op   func() string
```
burov: I don't see we actually fix the issue here, did you forget to push changes ?

burov: If not please refactor the tests according to the pattern used in `TestWatchConfig`

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +197,802 @@ func TestSetConfigEnabled(t *testing.T) {
...
+		{op: TaskNotificationEnabled, want: taskNotificationEnabledDefault},
+		{op: GuestPoliciesEnabled, want: guestPoliciesEnabledDefault},
+		{op: Debug, want: debugEnabledDefault},
 	}
 	for _, tt := range testsBool {
-		if tt.op() != tt.want {
-			f := filepath.Base(runtime.FuncForPC(reflect.ValueOf(tt.op).Pointer()).Name())
-			t.Errorf("%q: got(%t) != want(%t)", f, tt.op(), tt.want)
-		}
+		f := filepath.Base(runtime.FuncForPC(reflect.ValueOf(tt.op).Pointer()).Name())
```
burov: Do we still need this here ?

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +197,802 @@ func TestSetConfigEnabled(t *testing.T) {
...
+		t.Run(fmt.Sprintf("%s is requested, returns default boolean", f), func(t *testing.T) {
+			utiltest.AssertEquals(t, tt.op(), tt.want)
+		})
 	}
 
-	if SvcPollInterval().Minutes() != float64(osConfigPollIntervalDefault) {
-		t.Errorf("Default poll interval: got(%f) != want(%d)", SvcPollInterval().Minutes(), osConfigPollIntervalDefault)
-	}
+	utiltest.AssertEquals(t, SvcPollInterval().Minutes(), float64(osConfigPollIntervalDefault))
```
burov: Can we move it to the general tests cases

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +197,802 @@ func TestSetConfigEnabled(t *testing.T) {
...
 
 	expectedEndpoint := "fake-zone-osconfig.googleapis.com.:443"
```
burov: The same

agentconfig/agentconfig_test.go
burov: Get it, i think long term we can change the names to public, but i don't want to bloat this PR

agentconfig/agentconfig_test.go
burov: NIT: we can move it to global variable down the file to keep noice out of actual tests.

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +197,802 @@ func TestSetConfigEnabled(t *testing.T) {
...
+}
+
+// TestSetConfigError returns an unmarshal error when metadata is empty.
+func TestSetConfigError(t *testing.T) {
+	setupMockMetadataServer(t, func(w http.ResponseWriter, r *http.Request) {})
+	utiltest.OverrideVariable(t, &osConfigWatchConfigTimeout, 1*time.Millisecond)
+
+	err := WatchConfig(context.Background())
+	var dummy metadataJSON
+	expectedErr := json.Unmarshal([]byte{}, &dummy)
```
burov: As we repeat the json error i kind of suggest to move it to global variable or create a function that generate this error

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +197,802 @@ func TestSetConfigEnabled(t *testing.T) {
...
+
+	utiltest.OverrideVariable(t, stdout, false)
+	utiltest.OverrideVariable(t, disableLocalLogging, false)
+	utiltest.AssertEquals(t, Stdout(), false)
+	utiltest.AssertEquals(t, DisableLocalLogging(), false)
+}
+
+// TestLogFeatures logs feature status without panicking.
+func TestLogFeatures(t *testing.T) {
+	LogFeatures(context.Background())
```
burov: What we're testing here ? I don't see we actually verify any output or side behavior of the LogFeature, that means this test case just to syntactically bump the code coverage without any check, i'm strongly against such scenarios .

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +197,802 @@ func TestSetConfigEnabled(t *testing.T) {
...
+	})})
+
+	err := WatchConfig(context.Background())
+
+	expectedBaseErr := &url.Error{
+		Op:  "Get",
+		URL: "http://mock-host/computeMetadata/v1/?recursive=true&alt=json&wait_for_change=true&last_etag=0&timeout_sec=60",
+		Err: mockNetErr,
 	}
+	expectedErr := fmt.Errorf("network error when requesting metadata, make sure your instance has an active network and can reach the metadata server: %w", expectedBaseErr)
```
burov: Please stick with one pattern `expected` vs `want`,  we're slowly refactoring to `want` as it more aligned with Golang standards

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +197,802 @@ func TestSetConfigEnabled(t *testing.T) {
...
+	LogFeatures(context.Background())
+}
+
+// TestIDToken validates token caching and token parsing errors.
+func TestIDToken(t *testing.T) {
+	// Create a valid dummy JWS token
+	// Header: {"alg":"RS256","typ":"JWT"} -> eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9
+	// Payload: {"exp": 4102444800} (January 1, 2100 00:00:00 UTC) -> eyJleHAiOiA0MTAyNDQ0ODAwfQ
+	// Signature: dummy -> ZHVtbXk
+	validToken := "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOiA0MTAyNDQ0ODAwfQ.ZHVtbXk"
```
burov: Is there a possibility to move the generation of token to some function to avoid bloating test case with unnecessary comments ?

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +197,802 @@ func TestSetConfigEnabled(t *testing.T) {
...
+			name: "token expires within ten minutes, fetches a fresh token on each call",
+			handler: func(w http.ResponseWriter, r *http.Request) {
+				if strings.HasPrefix(r.URL.Path, "/computeMetadata/v1/instance/service-accounts/default/identity") {
+					w.Header().Set("Metadata-Flavor", "Google")
+					fmt.Fprint(w, expiringToken)
+					return
+				}
+				http.NotFound(w, r)
+			},
+			numCalls:     2,
```
burov: See the comment above, we don't need the two calls actually, and as there the only one case that have value other than 1, i think we should get rid of this.

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +197,802 @@ func TestSetConfigEnabled(t *testing.T) {
...
+			wantToken:    validToken,
+			wantErr:      nil,
+			wantRequests: 1, // Only 1 request should be made due to caching
+		},
+		{
+			name: "token expires within ten minutes, fetches a fresh token on each call",
```
burov: I think the logic of the test is not actually correct, what i expect to be tested in this scenario that if existing token is token that have less than 10 minutes we request the server and get a valid token. 

So my perspective the test should looks like that, we manually set token to expired, server return valid and we verify that the actual token we get is valid not expired and the server called at least one time.

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +197,802 @@ func TestSetConfigEnabled(t *testing.T) {
...
+			suffix:  "test-500",
+			wantNil: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			body, etag, err := getMetadata(tt.suffix)
+			utiltest.AssertErrorMatch(t, err, nil)
+			if tt.wantNil {
```
burov: I think we can stick with just checking that we get empty string in the scenarios we expect, string([]byte(nil)) is just an empty string

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +197,802 @@ func TestSetConfigEnabled(t *testing.T) {
...
+	var requestedURL string
+	utiltest.OverrideVariable(t, &defaultClient, &http.Client{Transport: roundTripperFunc(func(req *http.Request) (*http.Response, error) {
+		requestedURL = req.URL.String()
+		return &http.Response{StatusCode: 200, Body: io.NopCloser(strings.NewReader("mock response"))}, nil
+	})})
+
+	_, _, err := getMetadata("test-suffix")
+	utiltest.AssertErrorMatch(t, err, nil)
+
+	expected := "http://" + metadataIP + "/computeMetadata/v1/test-suffix"
```
burov: reminder on expected vs want

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +197,802 @@ func TestSetConfigEnabled(t *testing.T) {
...
+			defer wg.Done()
+			le.set(fmt.Sprintf("etag-%d", val))
+			_ = le.get()
+		}(i)
+	}
+	wg.Wait()
+
+	if le.get() == "" {
+		t.Errorf("Expected non-empty etag")
+	}
```
burov: It will never be empty string, i'm not sure this test is testing anything.  i'm not sure it is actually possible to test this part in an easy way.

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +197,802 @@ func TestSetConfigEnabled(t *testing.T) {
...
+}
+
+// TestSystemPaths returns OS-specific system paths.
+func TestSystemPaths(t *testing.T) {
+	utiltest.OverrideVariable(t, &goos, runtime.GOOS)
+	// t.Setenv("XDG_CACHE_HOME", filepath.Join(t.TempDir(), "system-cache"))
```
burov: Remove commented code

agentconfig/agentconfig_test.go
```diff
@@ -139,93 +197,802 @@ func TestSetConfigEnabled(t *testing.T) {
...
+}
+
+// TestMiscGetters returns static getter values.
+func TestMiscGetters(t *testing.T) {
+	SetVersion("1.2.3")
+
+	tests := []struct {
+		name string
+		got  interface{}
+		want interface{}
```
burov: switch to any

agentconfig/agentconfig_test.go
iliatsuprik: fixed

agentconfig/agentconfig_test.go
iliatsuprik: fixed

agentconfig/agentconfig_test.go
iliatsuprik: fixed

agentconfig/agentconfig_test.go
iliatsuprik: fixed

agentconfig/agentconfig_test.go
iliatsuprik: refactored

agentconfig/agentconfig_test.go
iliatsuprik: no, removed

agentconfig/agentconfig_test.go
iliatsuprik: done

agentconfig/agentconfig_test.go
iliatsuprik: done

agentconfig/agentconfig_test.go
iliatsuprik: done

agentconfig/agentconfig_test.go
iliatsuprik: removed (it just checked there are no crashes on this code)

agentconfig/agentconfig_test.go
iliatsuprik: fixed

agentconfig/agentconfig_test.go
iliatsuprik: done

agentconfig/agentconfig_test.go
iliatsuprik: removed this

agentconfig/agentconfig_test.go
iliatsuprik: fixed

agentconfig/agentconfig_test.go
iliatsuprik: done

agentconfig/agentconfig_test.go
iliatsuprik: done

agentconfig/agentconfig_test.go
iliatsuprik: removed this test

agentconfig/agentconfig_test.go
iliatsuprik: done

agentconfig/agentconfig_test.go
iliatsuprik: done

petercieslak (dismissed):

agentconfig/agentconfig_test.go
```diff
@@ -282,22 +977,424 @@ func TestDisableCloudLogging(t *testing.T) {
...
+						OSConfigEndpointOld: tt.projOld,
+					},
+				},
+				Instance: instanceJSON{
+					Attributes: attributesJSON{
+						OSConfigEndpoint:    tt.instNew,
+						OSConfigEndpointOld: tt.instOld,
+					},
+				},
+			}
```
burov: What the reason to bloat test function here, test function should be as short as possible, initialization of the metadata and config must be moved to the initialization of the test case or to separate function, putting it here decrease readability and reduce focus on what is important the actual verification.

agentconfig/agentconfig_test.go
```diff
@@ -282,22 +977,424 @@ func TestDisableCloudLogging(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := &config{}
+			md := metadataJSON{
+				Project:  projectJSON{Attributes: attributesJSON{TraceGetInventory: tt.projVal}},
+				Instance: instanceJSON{Attributes: attributesJSON{TraceGetInventory: tt.instVal}},
+			}
```
burov: Same comment as below.

burov: In the example of TestSetSVCEndpoint, i can understand that amount of initialization is big, in this scenario i don't see why metadata should be initialized inside of the test function.

agentconfig/agentconfig_test.go
```diff
@@ -282,22 +977,424 @@ func TestDisableCloudLogging(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := &config{}
+			md := metadataJSON{
+				Project:  projectJSON{Attributes: attributesJSON{ScalibrLinuxEnabled: tt.projVal}},
+				Instance: instanceJSON{Attributes: attributesJSON{ScalibrLinuxEnabled: tt.instVal}},
+			}
```
burov: Same comment as below

agentconfig/agentconfig_test.go
```diff
@@ -105,127 +159,791 @@ func TestSetConfigEnabled(t *testing.T) {
...
-			f := filepath.Base(runtime.FuncForPC(reflect.ValueOf(tt.op).Pointer()).Name())
-			t.Errorf("%q: got(%q) != want(%q)", f, tt.op(), tt.want)
+		{
+			name:         "token stays valid across two calls, reuses cached token",
+			handler:      metadataIdentityHandler(validToken),
+			numCalls:     2,
+			wantToken:    validToken,
+			wantErr:      nil,
+			wantRequests: 1,
+		},
```
burov, line 492: Again, this test case can be verified without numCalls parameter, i strongly recommend get rid of this parameter as it complicate test function and reduce it readability dramatically

iliatsuprik: removed

agentconfig/agentconfig_test.go
iliatsuprik: moved this to test setup

agentconfig/agentconfig_test.go
iliatsuprik: moved this to test setup

agentconfig/agentconfig_test.go
iliatsuprik: moved this to test setup

burov (approved):

petercieslak (approved):

PR #932, opened 2026-03-19
Cover patch_linux.go by unit tests

iliatsuprik:
This PR contains unit tests and minor changes in code for github.com/GoogleCloudPlatform/osconfig/agentendpoint/patch_linux.go.
Initial coverage: 23.4%, current coverage: 100%

The following variables were moved to be global (needed to mock them):
```
runAptGetUpgrade = ospatch.RunAptGetUpgrade
runYumUpdate     = ospatch.RunYumUpdate
runZypperPatch   = ospatch.RunZypperPatch
retryPeriod      = 3 * time.Minute
```

### DIFF
--- a/agentendpoint/patch_linux_test.go
+++ b/agentendpoint/patch_linux_test.go
@@ -32,7 +32,7 @@ import (
 func TestExcludeConversion(t *testing.T) {
 	regex, _ := regexp.Compile("PackageName")
 	emptyRegex, _ := regexp.Compile("")
-	regexpErr := func() error { _, err := regexp.Compile("[a-z"); return err }()
+	_, regexpErr := regexp.Compile("[a-z")
 
 	tests := []struct {
 		name    string
@@ -40,20 +40,20 @@ func TestExcludeConversion(t *testing.T) {
 		want    []*ospatch.Exclude
 		wantErr error
 	}{
-		{name: "StrictStringConversion", input: []string{"PackageName"}, want: CreateStringExcludes("PackageName")},
-		{name: "MultipleStringConversion", input: []string{"PackageName1", "PackageName2"}, want: CreateStringExcludes("PackageName1", "PackageName2")},
-		{name: "RegexConversion", input: []string{"/PackageName/"}, want: []*ospatch.Exclude{ospatch.CreateRegexExclude(regex)}},
-		{name: "CornerCaseRegex", input: []string{"//"}, want: []*ospatch.Exclude{ospatch.CreateRegexExclude(emptyRegex)}},
-		{name: "CornerCaseStrictString", input: []string{"/"}, want: CreateStringExcludes("/")},
-		{name: "CornerCaseEmptyString", input: []string{""}, want: CreateStringExcludes("")},
-		{name: "ErrorInvalidRegex", input: []string{"/[a-z/"}, wantErr: regexpErr},
+		{name: "Single package name, want one string exclude", input: []string{"PackageName"}, want: createStringExcludes("PackageName")},
+		{name: "Multiple package names, want multiple string excludes", input: []string{"PackageName1", "PackageName2"}, want: createStringExcludes("PackageName1", "PackageName2")},
+		{name: "Slash-wrapped value, want regex exclude", input: []string{"/PackageName/"}, want: []*ospatch.Exclude{ospatch.CreateRegexExclude(regex)}},
+		{name: "Empty regex //, want empty regex exclude", input: []string{"//"}, want: []*ospatch.Exclude{ospatch.CreateRegexExclude(emptyRegex)}},
+		{name: "Single slash, want string exclude", input: []string{"/"}, want: createStringExcludes("/")},
+		{name: "Empty string, want string exclude", input: []string{""}, want: createStringExcludes("")},
+		{name: "Invalid regex, want regex compile error", input: []string{"/[a-z/"}, wantErr: regexpErr},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			excludes, err := convertInputToExcludes(tt.input)
 
-			if tt.wantErr == nil && !reflect.DeepEqual(excludes, tt.want) {
+			if !reflect.DeepEqual(excludes, tt.want) {
 				t.Errorf("convertInputToExcludes() = %s, want = %s", toString(excludes), toString(tt.want))
 			}
 			utiltest.AssertErrorMatch(t, err, tt.wantErr)
@@ -62,8 +62,7 @@ func TestExcludeConversion(t *testing.T) {
 }
 
 func TestRunUpdates(t *testing.T) {
-	stubGlobalState(t)
-	retryPeriod = 1 * time.Millisecond
+	utiltest.OverrideVariable(t, &retryPeriod, 1*time.Millisecond)
 
 	mockAptSuccess := func(ctx context.Context, opts ...ospatch.AptGetUpgradeOption) error { return nil }
 	mockYumSuccess := func(ctx context.Context, opts ...ospatch.YumUpdateOption) error { return nil }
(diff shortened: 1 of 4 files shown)


agentendpoint/patch_linux_test.go
```diff
@@ -15,44 +15,257 @@
...
 	emptyRegex, _ := regexp.Compile("")
 
 	tests := []struct {
-		name  string
-		input []string
-		want  []*ospatch.Exclude
+		name    string
+		input   []string
+		want    []*ospatch.Exclude
+		wantErr bool
```
burov: I don't see the name of the variable make sense

agentendpoint/patch_linux_test.go
```diff
@@ -15,44 +15,246 @@
...
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			excludes, err := convertInputToExcludes(tt.input)
-			if err != nil {
-				t.Errorf("err = %v, want %v", err, nil)
-			}
-			if !reflect.DeepEqual(excludes, tt.want) {
+
+			if tt.wantErr == nil && !reflect.DeepEqual(excludes, tt.want) {
```
burov, line 56: https://github.com/GoogleCloudPlatform/osconfig/blob/master/util/utiltest/utiltest.go#L88 should be useful in this case, as well i suppose we don't need to check error is nil to ensure want is equal to got

agentendpoint/patch_linux_test.go
```diff
@@ -15,44 +15,246 @@
...
 				t.Errorf("convertInputToExcludes() = %s, want = %s", toString(excludes), toString(tt.want))
 			}
+			utiltest.AssertErrorMatch(t, err, tt.wantErr)
 		})
 	}
 }
 
+func TestRunUpdates(t *testing.T) {
+	stubGlobalState(t)
```
burov, line 65: I am a big fan of the principle explicit is better than explicit, i'd prefer each test case set-up the state it require rather than some global function change state before the test is running.

agentendpoint/patch_linux_test.go
```diff
@@ -15,44 +15,246 @@
...
+		setupMocks func()
+		taskConfig *agentendpointpb.PatchConfig
+		wantErr    error
+	}{
+		{
+			name: "No package managers found",
+			setupMocks: func() {
+				packages.AptExists = false
+				packages.YumExists = false
+				packages.ZypperExists = false
```
burov, line 87: You can simplify the set-up by creating one function that set everything to false and after that enable only what is required.

For example disableAllPackageManagers() can set all package managers to false. https://github.com/GoogleCloudPlatform/osconfig/blob/master/util/utiltest/utiltest.go#L140 might be useful in both cases as the base function used in disableAllPackageManagers and set it to expected value in the test.

agentendpoint/patch_linux_test.go
```diff
@@ -15,44 +15,246 @@
...
+	regexpErr := func() error { _, err := regexp.Compile("[a-z"); return err }()
+
+	tests := []struct {
+		name       string
+		setupMocks func()
+		taskConfig *agentendpointpb.PatchConfig
+		wantErr    error
+	}{
+		{
+			name: "No package managers found",
```
burov, line 83: Please keep the names in the format, "input, expected results" For example "No package managers found, want empty Patch Config and no errors"

agentendpoint/patch_linux_test.go
```diff
@@ -15,44 +15,246 @@
...
+		{
+			name: "Apt conversion error bubbles up",
+			setupMocks: func() {
+				packages.AptExists = true
+				packages.DpkgQueryExists = true
+			},
+			taskConfig: &agentendpointpb.PatchConfig{
+				Apt: &agentendpointpb.AptSettings{Excludes: []string{"/[a-z/"}}, // Invalid regex
+			},
+			wantErr: func() error { _, err := regexp.Compile("[a-z"); return err }(),
```
burov, line 115: Why don't we use regexpErr here ?

agentendpoint/patch_linux_test.go
```diff
@@ -15,44 +15,246 @@
...
+	stubGlobalState(t)
+	retryPeriod = 1 * time.Millisecond
+
+	mockAptSuccess := func(ctx context.Context, opts ...ospatch.AptGetUpgradeOption) error { return nil }
+	mockYumSuccess := func(ctx context.Context, opts ...ospatch.YumUpdateOption) error { return nil }
+	mockZypperSuccess := func(ctx context.Context, opts ...ospatch.ZypperPatchOption) error { return nil }
+	mockAptErr := func(ctx context.Context, opts ...ospatch.AptGetUpgradeOption) error { return errors.New("apt err") }
+	mockYumErr := func(ctx context.Context, opts ...ospatch.YumUpdateOption) error { return errors.New("yum err") }
+	mockZypperErr := func(ctx context.Context, opts ...ospatch.ZypperPatchOption) error { return errors.New("zypper err") }
+	regexpErr := func() error { _, err := regexp.Compile("[a-z"); return err }()
```
burov, line 74: What the reason to use the function here ?

I suppose we can replace it with
_, regexpErr := regexp.Compile("[a-z")

agentendpoint/patch_linux_test.go
```diff
@@ -15,44 +15,246 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			tt.setupMocks()
+
+			task := &patchTask{
+				Task: &applyPatchesTask{
+					ApplyPatchesTask: &agentendpointpb.ApplyPatchesTask{
+						PatchConfig: tt.taskConfig,
+					},
+				},
+			}
```
burov, line 223: Can be extracted to the function to improve readability

task := initializePatchTask(tt.taskConfig)
...

agentendpoint/patch_linux_test.go
iliatsuprik: names are pretty standard. I suppose you want to make wantErr to be error, that's done

agentendpoint/patch_linux_test.go
iliatsuprik: done

agentendpoint/patch_linux_test.go
iliatsuprik: done

agentendpoint/patch_linux_test.go
iliatsuprik: done

agentendpoint/patch_linux_test.go
iliatsuprik: done

agentendpoint/patch_linux_test.go
iliatsuprik: done

agentendpoint/patch_linux_test.go
iliatsuprik: done

agentendpoint/patch_linux_test.go
iliatsuprik: we have to use reflect.DeepEqual as excludes contain non-public fields, so AssertEquals won't work here. `tt.wantErr == nil` is removed

util/utiltest/utiltest.go
```diff
@@ -137,10 +137,8 @@ func AssertFileContents(t *testing.T, filePath string, wantContents string) {
 }
 
 // OverrideVariable overrides the value of a variable and returns a function to restore it.
-func OverrideVariable[T any](ptr *T, val T) func() {
+func OverrideVariable[T any](t *testing.T, ptr *T, val T) {
```
burov, line 140: Good change, i love the usage of t.Cleanup in this case as well as parameter of t save us from using this outside of testing package, well done

agentendpoint/patch_linux_test.go
```diff
@@ -15,44 +15,233 @@
...
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			excludes, err := convertInputToExcludes(tt.input)
-			if err != nil {
-				t.Errorf("err = %v, want %v", err, nil)
-			}
+
 			if !reflect.DeepEqual(excludes, tt.want) {
 				t.Errorf("convertInputToExcludes() = %s, want = %s", toString(excludes), toString(tt.want))
 			}
```
burov, line 58: I suppose we still can play around with options to introduce hook function to properly equal objects, but i don't want to insists on it.

burov (approved):

szerlak (approved):

PR #935, opened 2026-03-23
Cover config/file.go by unit tests

iliatsuprik:
This PR contains unit tests for github.com/GoogleCloudPlatform/osconfig/config/file.go.
Initial coverage: 14.3%, current coverage: 100%

### DIFF
--- a/config/file_test.go
+++ b/config/file_test.go
@@ -1,0 +1,229 @@
+//  Copyright 2026 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package config
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"cloud.google.com/go/osconfig/agentendpoint/apiv1/agentendpointpb"
+)
+
+// setupTempDir creates a temporary directory for file downloads and returns its path.
+func setupTempDir(t *testing.T) string {
+	t.Helper()
+	tmpDir, err := os.MkdirTemp("", "osconfig_file_test_")
+	if err != nil {
+		t.Fatalf("failed to create temp dir: %v", err)
+	}
+	t.Cleanup(func() { os.RemoveAll(tmpDir) })
+	return tmpDir
+}
+
+// setEnv temporarily sets an environment variable, restoring its previous value
+// (or unsetting it) when the test finishes.
+func setEnv(t *testing.T, key, val string) {
+	t.Helper()
+	oldVal, exists := os.LookupEnv(key)
+	os.Setenv(key, val)
+	t.Cleanup(func() {
+		if exists {
+			os.Setenv(key, oldVal)
+		} else {
+			os.Unsetenv(key)
+		}
+	})
+}
+
+// setupMockServer creates a mock HTTP server to simulate remote file downloads.
+func setupMockServer(t *testing.T) *httptest.Server {
+	t.Helper()
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/success" {
+			w.Write([]byte("test content"))
+			return
+		}
+		http.NotFound(w, r)
+	}))
+	t.Cleanup(ts.Close)
+	return ts
+}
+
+// TestChecksum verifies the SHA256 checksum calculation for given io.Reader inputs.
+func TestChecksum(t *testing.T) {
+	tests := []struct {
+		name  string
+		input string
+		want  string
+	}{
+		{
+			name:  "empty string",
+			input: "",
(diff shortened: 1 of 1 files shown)


config/file_test.go
```diff
@@ -0,0 +1,229 @@
...
+		},
+		{
+			name: "Unknown file type",
+			file: &agentendpointpb.OSPolicy_Resource_File{
+				Type: nil,
+			},
+			wantErr: "unknown remote File type: <nil>",
+		},
+		{
+			name: "GCS remote file fetch error",
```
burov, line 155: Is there a success case for GCS ?

config/file_test.go
```diff
@@ -0,0 +1,229 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"cloud.google.com/go/osconfig/agentendpoint/apiv1/agentendpointpb"
+)
+
+// setupTempDir creates a temporary directory for file downloads and returns its path.
+func setupTempDir(t *testing.T) string {
```
burov, line 31: Might be a good candidate for utiltest

config/file_test.go
```diff
@@ -0,0 +1,229 @@
...
+	if err != nil {
+		t.Fatalf("failed to create temp dir: %v", err)
+	}
+	t.Cleanup(func() { os.RemoveAll(tmpDir) })
+	return tmpDir
+}
+
+// setEnv temporarily sets an environment variable, restoring its previous value
+// (or unsetting it) when the test finishes.
+func setEnv(t *testing.T, key, val string) {
```
burov, line 43: Might be a good candidate for utiltest

config/file_test.go
```diff
@@ -0,0 +1,229 @@
...
+			name:  "simple string",
+			input: "test content",
+			want:  "6ae8a75555209fd6c44157c0aed8016e763ff435a19cf186f76863140143ff72",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			r := strings.NewReader(tt.input)
+			if got := checksum(r); got != tt.want {
```
burov, line 92: Might be good to use for consistency https://github.com/GoogleCloudPlatform/osconfig/blob/master/util/utiltest/utiltest.go#L88

config/file_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+// TestChecksum verifies the SHA256 checksum calculation for given io.Reader inputs.
+func TestChecksum(t *testing.T) {
+	tests := []struct {
+		name  string
+		input string
+		want  string
+	}{
+		{
+			name:  "empty string",
```
burov, line 78: Not a big concern in this scenario, but the name in the form of "input, expected result" would be a bit better

config/file_test.go
```diff
@@ -0,0 +1,229 @@
...
+
+			_, err := downloadFile(ctx, destPath, 0644, tt.file)
+
+			if err == nil {
+				if tt.wantErr != "" {
+					t.Errorf("downloadFile() expected error %q, got nil", tt.wantErr)
+				}
+			} else if err.Error() != tt.wantErr {
+				t.Errorf("downloadFile() error = %q, wantErr %q", err.Error(), tt.wantErr)
+			}
```
burov, line 226: My opinion that the tests logic should be simple, branching to verify an error can be avoided with support of utiltest.AssertErrorMatch

config/file_test.go
```diff
@@ -0,0 +1,229 @@
...
+			env: map[string]string{
+				"STORAGE_EMULATOR_HOST":          "",
+				"GOOGLE_APPLICATION_CREDENTIALS": "/non/existent/path/to/creds.json",
+			},
+		},
+	}
+
+	for i, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			for k, v := range tt.env {
```
burov, line 211: In this particular scenario i suppose imperative approach will be beneficial rather than declarative, if we replace envs variable in the test case with the simple setup () string

we can do set-up 

```
func (i int) string {
     setEnv("GOOGLE_APPLICATION_CREDENTIALS", "/non/existent/path/to/creds.json")
      ...

    return filepath.Join(tmpDir, fmt.Sprintf("test_file_%d", i))
}
```

In this case we will significantly clean-up the test initialization logic without any harm to the readability of the test case.

config/file_test.go
iliatsuprik: added

config/file_test.go
iliatsuprik: I found t.Setenv, so this method was removed

config/file_test.go
iliatsuprik: done

config/file_test.go
iliatsuprik: done

config/file_test.go
iliatsuprik: done

config/file_test.go
iliatsuprik: done

config/file_test.go
iliatsuprik: I found t.TempDir, so removed this function

config/file_test.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	for i, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var destPath string
+			if tt.setup != nil {
+				destPath = tt.setup(t, i)
+			} else {
+				destPath = filepath.Join(tmpDir, fmt.Sprintf("test_file_%d", i))
+			}
```
burov: Just to make it consistent and more readable i suppose we can stick with always have a set-up function that return destination, in this sceanrio we can replace it to

`destPath := destPath = tt.setup(t, i)`

That is more readable from my perspective

burov (dismissed):

config/file_test.go
iliatsuprik: done

burov (approved):

szerlak (approved):

PR #940, opened 2026-03-30
Bump timeouts for all build workflows

burov:
I bump timeouts as i see some occasional failures due to the timeouts error. When i look under the hood it seems like builds finishing without error after some time.


petercieslak (approved):

paulinakania (approved):
LGTM

PR #941, opened 2026-03-31
Cover google_update.go by unit tests

iliatsuprik:
Coverage increased from 0 to 100%


ospatch/google_update_test.go
```diff
@@ -0,0 +1,135 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.desc, func(t *testing.T) {
+			mockCtrl := gomock.NewController(t)
+			defer mockCtrl.Finish()
+
+			mockCommandRunner := utilmocks.NewMockCommandRunner(mockCtrl)
+			packages.SetCommandRunner(mockCommandRunner)
```
burov: Do we really need to recreate mock each time or we can do it just once?

ospatch/google_update_test.go
```diff
@@ -0,0 +1,135 @@
...
+				Times(1)
+
+			if tt.expectInstall {
+				installArgs := append([]string{"-noconfirm", "install"}, tt.installPkgNames...)
+				installCmd := utilmocks.EqCmd(exec.Command(googet, installArgs...))
+				mockCommandRunner.EXPECT().
+					Run(ctx, installCmd).
+					After(updatesCall).
+					Return(tt.installOut, []byte("stderr"), tt.installErr).
+					Times(1)
```
burov: I think declarative way of expected command improve readability of the code and reduce cognitive load of the actual test case, see https://github.com/GoogleCloudPlatform/osconfig/blob/master/packages/pip_test.go#L25

ospatch/google_update_test.go
```diff
@@ -0,0 +1,135 @@
...
+		updatesOut      []byte
+		updatesErr      error
+		installOut      []byte
+		installErr      error
+		expectInstall   bool
+		installPkgNames []string
+		wantErr         error
+	}{
+		{
+			desc:            "success",
```
burov: Please be more specific in name of the test, list what is expected result of the test, success is to generic from my perspective, i suggest to have something like "want install command list packages foo and bar"

burov: Same on other cases "updates error" -> "googet update error is propogated"

ospatch/google_update_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	"github.com/GoogleCloudPlatform/osconfig/packages"
+	utilmocks "github.com/GoogleCloudPlatform/osconfig/util/mocks"
+	"github.com/GoogleCloudPlatform/osconfig/util/utiltest"
+	"github.com/golang/mock/gomock"
+)
+
+func TestRunGooGetUpdate(t *testing.T) {
+	ctx := context.Background()
+	googet := filepath.Join(os.Getenv("GooGetRoot"), "googet.exe")
```
burov: Can we avoid reliance on the environment variable here ?

iliatsuprik: done

ospatch/google_update_test.go
iliatsuprik: updated

ospatch/google_update_test.go
iliatsuprik: fixed

ospatch/google_update_test.go
iliatsuprik: moved this out of for loop

ospatch/google_update_test.go
```diff
@@ -0,0 +1,179 @@
...
+				cmd := utilmocks.EqCmd(ec.Cmd)
+				call := mockCommandRunner.EXPECT().
+					Run(ctx, cmd).
+					Return(ec.Stdout, ec.Stderr, ec.Err).
+					Times(1)
+				if prev != nil {
+					call.After(prev)
+				}
+				prev = call
+			}
```
burov: Can we create a function that accept set of expectations and set-up the mock, we have many places where we do set-up mocks in the way

ospatch/google_update_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+	excludeName := "bar"
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+
+	mockCommandRunner := utilmocks.NewMockCommandRunner(mockCtrl)
+	packages.SetCommandRunner(mockCommandRunner)
+
+	tests := []struct {
+		desc     string
```
burov: stick with `name` as it is consistent with other tests

ospatch/google_update_test.go
```diff
@@ -0,0 +1,179 @@
...
+		opts     []GooGetUpdateOption
+		commands []utiltest.ExpectedCommand
```
burov: Consider rename the parameter to expectedCommands to make name  more consistent with the purpose of the variable

iliatsuprik: done

ospatch/google_update_test.go
iliatsuprik: done

ospatch/google_update_test.go
iliatsuprik: done

burov (approved):

zoltak-g (approved):

PR #942, opened 2026-03-31
Cover ospatch/system_linux.go by unit tests

iliatsuprik:
Split `DisableAutoUpdates` function to a few sub-functions to make this code more testable. One note: I see `DisableAutoUpdates` is not used anywhere, so probably it can be removed.

Coverage increased from 0 to 76%


burov:
We will remove the logic https://github.com/GoogleCloudPlatform/osconfig/pull/970, so this PR can be closed

PR #943, opened 2026-04-01
Add unit tests for packages/pty_linux.go

ganochenkodg:
New test coverage for this file:
github.com/GoogleCloudPlatform/osconfig/packages/pty_linux.go:33:	ioctl 0.0% -> 100.0%
github.com/GoogleCloudPlatform/osconfig/packages/pty_linux.go:45:	runWithPty 0.0% -> 82.0%

### DIFF
--- a/packages/pty_linux_test.go
+++ b/packages/pty_linux_test.go
@@ -1,0 +1,130 @@
+//  Copyright 2026 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+package packages
+
+import (
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+	"unsafe"
+
+	"github.com/GoogleCloudPlatform/osconfig/util/utiltest"
+	"golang.org/x/sys/unix"
+)
+
+// CheckPtmxAvailability skips the test if /dev/ptmx is not available
+func CheckPtmxAvailability(t *testing.T) {
+	if _, err := os.Stat("/dev/ptmx"); os.IsNotExist(err) {
+		t.Skip("/dev/ptmx not found, skipping PTY tests")
+	}
+}
+
+// TestIoctl verifies that the ioctl wrapper correctly performs a system call on a valid file descriptor
+func TestIoctl(t *testing.T) {
+	CheckPtmxAvailability(t)
+
+	f, err := os.OpenFile("/dev/ptmx", os.O_RDWR, 0)
+	if err != nil {
+		t.Fatalf("unable to open /dev/ptmx: %v", err)
+	}
+	defer f.Close()
+
+	// TIOCSPTLCK (unlockpt)
+	var i int
+	if err := ioctl(f.Fd(), unix.TIOCSPTLCK, uintptr(unsafe.Pointer(&i))); err != nil {
+		t.Errorf("ioctl() unexpected error: %v", err)
+	}
+}
+
+// TestIoctlError ensures that the ioctl function correctly handles and returns errors
+func TestIoctlError(t *testing.T) {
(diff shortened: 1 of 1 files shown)


packages/pty_linux_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+// CheckPtmxAvailability skips the test if /dev/ptmx is not available
+func CheckPtmxAvailability(t *testing.T) {
+	if _, err := os.Stat("/dev/ptmx"); os.IsNotExist(err) {
+		t.Skip("/dev/ptmx not found, skipping PTY tests")
+	}
+}
+
+// TestIoctl verifies that the ioctl wrapper correctly performs a system call on a valid file descriptor
+func TestIoctl(t *testing.T) {
```
burov: Correct me if i'm wrong, but TestRunWithPty should cover this use case?

packages/pty_linux_test.go
```diff
@@ -0,0 +1,130 @@
...
+			stdout, stderr, err := runWithPty(tt.cmd)
+
+			utiltest.AssertErrorMatch(t, err, tt.wantErr)
+			if tt.wantErr != nil {
+				return
+			}
+
+			if err != nil {
+				t.Errorf("runWithPty() unexpected error: %v", err)
+			}
```
burov: I'm not sure why do we need this ? i suppose in the scenario when we get an expected error we can just compare that stdout is empty, is it ?

packages/pty_linux_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+			// PTY often converts \n to \r\n, normalize for comparison.
+			gotOut := strings.ReplaceAll(string(stdout), "\r\n", "\n")
```
burov: Can we just use the same stdout variable ?

```
stdout := strings.ReplaceAll(string(stdout), "\r\n", "\n")
```

 i don't see a reason to create another variable if we're not using stdout variable after defining gotOut ?

packages/pty_linux_test.go
```diff
@@ -0,0 +1,130 @@
...
+			if !strings.Contains(gotOut, tt.wantOut) {
```
burov: Why do we use Contains and not equals ? Can we rely on the commands that return the stable output all of the time ?

packages/pty_linux_test.go
```diff
@@ -0,0 +1,130 @@
...
+				t.Errorf("runWithPty() stdout = %q, wantOut %q", gotOut, tt.wantOut)
+			}
+			if !strings.Contains(string(stderr), tt.wantStderr) {
```
burov: utiltest.AssertEquals might become handy here

packages/pty_linux_test.go
ganochenkodg: Sorry, you're absolutely right, probably I started in the beginning with this function and forgot to remove. Removing it now...

ganochenkodg: Done

packages/pty_linux_test.go
ganochenkodg: Yes, I used it before moving to AssertErrorMatch, cleaned now

packages/pty_linux_test.go
ganochenkodg: Honestly there is no reason to use this replace here, results are stable and predictable. So i reworked this test block completely and removed useless code

packages/pty_linux_test.go
ganochenkodg: Yes, in this test test results are stable. Reworked completely. Funcs from utiltest are brilliant, if I find such direct string comparisons in other PRs I will also replace them them with AssertErrorMatch and AssertEquals functions...

packages/pty_linux_test.go
ganochenkodg: Yes, updated

packages/pty_linux_test.go
```diff
@@ -0,0 +1,100 @@
...
+			wantErr: &exec.Error{Name: "this-command-does-not-exist", Err: exec.ErrNotFound},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			stdout, stderr, err := runWithPty(tt.cmd)
+
+			utiltest.AssertErrorMatch(t, err, tt.wantErr)
+			if tt.wantErr != nil {
```
burov: I suppose we should not fail if we AsserEquals that stdout and stderr content is empty?

burov (dismissed):

packages/pty_linux_test.go
ganochenkodg: Hmm, yes, fixed

burov (approved):

petercieslak (approved):

PR #949, opened 2026-04-13
Add unit tests for policies/policies.go PART 1

ganochenkodg:
First part of unit tests for policies.go:

github.com/GoogleCloudPlatform/osconfig/policies/policies.go:39:	run			85.7%
github.com/GoogleCloudPlatform/osconfig/policies/policies.go:66:	Run			100.0%
github.com/GoogleCloudPlatform/osconfig/policies/policies.go:70:	installRecipes		100.0%
github.com/GoogleCloudPlatform/osconfig/policies/policies.go:214:	checksum		100.0%
github.com/GoogleCloudPlatform/osconfig/policies/policies.go:220:	writeIfChanged		91.7%


ganochenkodg:
Moved expectedCommand and setExpectations from PR's part2, as it's a dependency for another 4 PRs and would be better to merge it asap

policies/policies_test.go
```diff
@@ -0,0 +1,210 @@
...
+		},
+		{
+			name: "empty data, want hash of empty data",
+			data: []byte(""),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			r := bytes.NewReader(tt.data)
```
petercieslak: Please change the variable names so that it reflects more what is their purpose

policies/policies_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+			err := writeIfChanged(ctx, tt.newContent, path)
+			utiltest.AssertErrorMatch(t, err, tt.wantErr)
+			utiltest.AssertFileContents(t, path, string(tt.newContent))
+		})
+	}
+}
+
+// TestInstallRecipes verifies that installRecipes correctly iterates over and installs recipes.
+func TestInstallRecipes(t *testing.T) {
```
petercieslak: This test does not do what it should do according to the description - it does not verify if recipes are installed. I suggest to either change the test description or add verification if recipes are installed correctly (API for recipeDB https://github.com/GoogleCloudPlatform/osconfig/blob/master/policies/recipes/recipedb.go)

policies/policies_test.go
```diff
@@ -0,0 +1,210 @@
...
+			egp: &agentendpointpb.EffectiveGuestPolicy{
+				SoftwareRecipes: []*agentendpointpb.EffectiveGuestPolicy_SourcedSoftwareRecipe{
+					{
+						SoftwareRecipe: nil,
+					},
+				},
+			},
+		},
+		{
+			name: "policy with invalid recipe",
```
petercieslak: This test case is misleading. It suggests that there should be some error since the recipe is invalid, but later there is an assert to nil (`wantErr` is nil for all test cases). I suggest to change the description of this test case.

petercieslak (commented):
I see that this PR is a part of some larger chain, however I don't think that it should introduce code that is not used in this specific PR. It is hard to review such code without the whole context in head, it requires jumping between PR's to verify if every line of code is valid and because of that, the review process becomes really hard and complicated.

policies/policies_test.go
ganochenkodg: That's because installRecipes always returns nil and continues to install what possible. All errors go in clog. 
https://github.com/GoogleCloudPlatform/osconfig/blob/f0fbd304a30893a56fa73251bc0fc27b54a41d82/policies/policies.go#L70-L78
I can try to catch them and check.

policies/policies_test.go
ganochenkodg: Yes, renamed it

policies/policies_test.go
ganochenkodg: Updated

policies/policies_test.go
ganochenkodg: I have reworked this test, now we can check errors and this testcase returns one

policies/policies_test.go
```diff
@@ -0,0 +1,218 @@
...
+				Run(gomock.Any(), utilmocks.EqCmd(cmd)).
+				Return(expectedCmd.stdout, expectedCmd.stderr, expectedCmd.err).Times(1)
+		} else {
+			prev = mockCommandRunner.EXPECT().
+				Run(gomock.Any(), utilmocks.EqCmd(cmd)).
+				After(prev).
+				Return(expectedCmd.stdout, expectedCmd.stderr, expectedCmd.err).Times(1)
+		}
+	}
+}
```
burov: Should be a good candidate for the utiltest

ganochenkodg: I will update it then

ganochenkodg: @burov Done

util/utiltest/utiltest.go
```diff
@@ -3,14 +3,65 @@ package utiltest
...
+			prev = mockCommandRunner.EXPECT().
+				Run(gomock.Any(), utilmocks.EqCmd(cmd)).
+				After(prev).
+				Return(expectedCmd.Stdout, expectedCmd.Stderr, expectedCmd.Err).Times(1)
+		}
+	}
+}
+
+// AssertFormatMatch verifies that the gotString matches the wantFormat regular expression.
+func AssertFormatMatch(t *testing.T, gotString string, wantFormat string) {
```
burov: Specifying the type in the name of the parameter is not a good pattern

policies/policies_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+// TestRunSmoke is a smoke test for the Run function.
+func TestRunSmoke(t *testing.T) {
+	Run(context.Background())
+}
+
+// TestRunCancelledContext verifies that run returns immediately when context is cancelled.
+func TestRunCancelledContext(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
```
burov: What are we testing here ? basically test is running something and didn't check any side behavior, that means the test is mostly for the coverage, not the actual testing.

policies/policies_test.go
```diff
@@ -0,0 +1,183 @@
...
+			installRecipes(ctx, tt.egp)
+
+			utiltest.AssertFormatMatch(t, buf.String(), tt.wantErrFormat)
+		})
+	}
+}
+
+// TestRunSmoke is a smoke test for the Run function.
+func TestRunSmoke(t *testing.T) {
+	Run(context.Background())
```
burov: What are we testing here ? Test without checking any side behavior is just bumping coverage rather than test something.

util/utiltest/utiltest.go
```diff
@@ -3,14 +3,65 @@ package utiltest
...
 	"testing"
 
+	utilmocks "github.com/GoogleCloudPlatform/osconfig/util/mocks"
+	"github.com/golang/mock/gomock"
 	"github.com/google/go-cmp/cmp"
 	"github.com/kr/pretty"
 )
 
+// ExpectedCommand represents a command that is expected to be run.
+type ExpectedCommand struct {
```
burov: I love the idea to move this common code blocks to utiltest

policies/policies_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+// TestInstallRecipesHandlesInputs verifies that installRecipes correctly handles different EffectiveGuestPolicy inputs.
+func TestInstallRecipesHandlesInputs(t *testing.T) {
+	uniqueSuffix := fmt.Sprintf("-%d", time.Now().UnixNano())
+
+	tests := []struct {
+		name       string
+		egp        *agentendpointpb.EffectiveGuestPolicy
+		wantErrFormat string
+	}{
```
burov, line 127: Why don't we have any positive test cases ?

policies/policies_test.go
```diff
@@ -0,0 +1,183 @@
...
+											Destination: "/tmp/dest",
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			wantErrFormat: `(?s).*Error installing recipe: error running step 0 \(CopyFile\): could not find location for artifact "non-existent".*`,
```
burov: Even that i agree the errors returned by the function is not ideal, i don't like an idea of having regexp match for error, can we refactor errors raised by `installRecipe` to make it still information, but easier to verify ?

I suppose in this scanrio we can rely on exact match and avoid creating yet another function that AssertError is actually expected

policies/policies_test.go
ganochenkodg: Reworked these both tests, now it makes two checks - client is created and error during connection to metadata server.

policies/policies_test.go
ganochenkodg: The problem is that installRecipe returns nil and all errors go in clog -> they contain a lot of additional information like timestamps. Currently I can only check clog errors with regexp func. So the only way is to return errors in installRecipe and send them to clog.Errorf with if err!=nil block in run runc. Can I apply these changes?

policies/policies_test.go
ganochenkodg: Added two positive cases - with one and few recipes. Recipes without real steps, in this test I only check how tested function handles inputs and throws errors.

util/utiltest/utiltest.go
ganochenkodg: fixed

policies/policies_test.go
ganochenkodg: reworked

util/utiltest/utiltest.go
burov: https://github.com/GoogleCloudPlatform/osconfig/pull/941 Already added ExpectedCommand to the utiltest package, please update CL to  align with changes in https://github.com/GoogleCloudPlatform/osconfig/pull/941

policies/policies_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+// TestWriteIfChanged verifies that writeIfChanged only writes to the file if the content has changed.
+func TestWriteIfChanged(t *testing.T) {
+	ctx := context.Background()
+
+	tests := []struct {
+		name           string
+		initialContent []byte
+		newContent     []byte
+		wantErr        error
```
burov, line 69: There is zero cases where we have wantErr not nil, are we missing some test cases ?

policies/policies_test.go
```diff
@@ -0,0 +1,221 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			td := t.TempDir()
+			path := filepath.Join(td, "test_file")
+
+			if tt.initialContent != nil {
+				if err := os.WriteFile(path, tt.initialContent, 0644); err != nil {
+					t.Fatalf("failed to setup initial file: %v", err)
+				}
+			}
```
burov: I think that code can be generalized to the utiltest function. 
```path, err := utiltest.WriteToTempFile(t, content)```

File: policies/policies_test.go (hunk 10 above)
Comment (burov) [reply policies/policies_test.go]: From what i see installRecipe is a private function with only one case of usage, i suggest a bit of a refactoring here.

Let's create a support error type

```
type installRecipesError struct {
     errors []error
}

func (e installRecipesError) Error() string {
    results := make([]string, len(errors))
    for i, err := range errors {
        results[i] = err.Error()
    }

    return fmt.Sprintf("unable to install recipes:\n%s", strings.Join(results, ",\n"))
}
```

In this scenario we can properly aggregate multiple errors into single.

Refactor function installRecipes

```
func installRecipes(ctx context.Context, egp *agentendpointpb.EffectiveGuestPolicy) error {
    var errors []error
	for _, recipe := range egp.GetSoftwareRecipes() {
		if r := recipe.GetSoftwareRecipe(); r != nil {
			if err := recipes.InstallRecipe(ctx, r); err != nil {
                 errors = append(errors, fmt.Errorf("Error installing recipe: %v", err))
			}
		}
	}
    if len(errors) > 0 {
        return installRecipesError{errors: errors}
    }

	return nil
}
```

After that we can move logging of the error in places where installRecipes used.

File: util/utiltest/utiltest.go (hunk 8 above)
Comment (ganochenkodg) [reply util/utiltest/utiltest.go]: Hmm, you're right, merged. From first look I see they are pretty different, the merged one for example doesn't support envs, need to investigate it

Comment (ganochenkodg) [reply util/utiltest/utiltest.go]: Removed expectedCommand parts from this PR, so all dependent PRs - 951, 952, 953, 061 - can be reviewed and merged any time.

File: policies/policies_test.go (hunk 12 above)
Comment (ganochenkodg) [reply policies/policies_test.go]: Good idea, done

File: policies/policies_test.go (hunk 10 above)
Comment (ganochenkodg) [reply policies/policies_test.go]: Done

File: policies/policies_test.go (hunk 11 above)
Comment (ganochenkodg) [reply policies/policies_test.go:69]: I can see two negative cases for testing:
1. File exists before calling WriteIfChanged but unreadable. Unfortunately "chmod 000" doestn't work as tests are running under root
2. Instead of file name (existing or non-existing) we pass some real directory name and get an error with many random elements, like this "rename /tmp/TestWriteIfChangedDirectory2252093707/00159385.tmp /tmp/TestWriteIfChangedDirectory2252093707/001: file exists" -> we have to use regexp to parse it

Comment (burov) [reply policies/policies_test.go:69]: My point is why we have parameter that never used ? There 0 test cases in the test that check error.

The easiest way to test negative scenario is to pass directory as a path, it is expect file rather than directory, i suppose the reader will fail attempting to read content. 

As well if we think deeper function called in the writeIfChanged is wrong, we should use os.OpenFile instead of os.Open as directory in this scenario is actually unexpected input. In this case test with the directory will be the easiest.

In summary we should not test AtomicWrite it self as a part of this test and rather focus on the logic of the test.

File: policies/policies_test.go [hunk 13]
```diff
@@ -0,0 +1,212 @@
...
+			wantErr:        nil,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			path, err := utiltest.WriteToTempFile(t, "test_file", tt.initialContent)
+			if err != nil {
+				t.Fatalf("failed to setup initial file: %v", err)
+			}
```
burov: Why we're not using utiltest.AssertErrorMatchAndFail(t, err, nil) here ?

policies/policies_test.go
ganochenkodg: Appllied suggested changes and added one negative testcase with directory

policies/policies_test.go
ganochenkodg: Done

policies/policies_test.go
```diff
@@ -0,0 +1,226 @@
...
+			data: []byte(""),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			reader := bytes.NewReader(tt.data)
+			hasher := checksum(reader)
+
+			expected := sha256.Sum256(tt.data)
```
burov: Nit: stick with single pattern got/want,

policies/policies_test.go
```diff
@@ -0,0 +1,226 @@
...
+			utiltest.AssertErrorMatchAndFail(t, err, nil)
+
+			if tt.pathPredefined != "" {
+				path = tt.pathPredefined
+			}
+
+			err = writeIfChanged(ctx, tt.newContent, path)
+			utiltest.AssertErrorMatchAndFail(t, err, tt.wantErr)
+
+			if tt.wantErr == nil {
```
burov: Again, if we use `utiltest.AssertErrorMatchAndFail` it exit at the point, we don't need to check if `wantErr == nil`

policies/policies_test.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			path, err := utiltest.WriteToTempFile(t, "test_file", tt.initialContent)
+			utiltest.AssertErrorMatchAndFail(t, err, nil)
+
+			if tt.pathPredefined != "" {
+				path = tt.pathPredefined
+			}
```
burov: Please avoid bloating test functions, what the reason to set path variable and replace it later if we can just replace pathPredefined with path variable, there is no reason to return an error from  utiltest.WriteToTempFile we can refactor this function to WriteToTempFileMust and just return path.

policies/policies_test.go
ganochenkodg: Nope, it has another logic. With:
```golang
utiltest.AssertErrorMatchAndFail(t, err, tt.wantErr)
utiltest.AssertFileContents(t, path, string(tt.newContent))
```

I get:
--- FAIL: TestWriteIfChanged (0.01s)
    --- FAIL: TestWriteIfChanged/path_is_a_directory,_want_error (0.00s)
        policies_test.go:110: Failed to read file "/tmp": read /tmp: is a directory
        
 It goes through with this path = /tmp and fails on content checking. Maybe lets modify AssertErrorMatch next way:
 ```golang
 func assertErrorMatch(t *testing.T, gotErr, wantErr error, failNow bool, skipNow bool) {
	t.Helper()
	if gotErr == nil && wantErr == nil {
		return
	}
	if gotErr == nil || wantErr == nil || reflect.TypeOf(gotErr) != reflect.TypeOf(wantErr) || gotErr.Error() != wantErr.Error() {
		t.Errorf("Errors mismatch, want %v, got %v", wantErr, gotErr)
		if failNow {
			t.FailNow()
		}
	}
	if skipNow{
	  t.SkipNow()
	}
}
```

So If we run it with skipNow = true and errors are not nil and they are same - we stop current step (and don't run utiltest.AssertFileContents(t, path, string(tt.newContent)) or any other code after) and simply go to the next testcase?

Comment (burov) [reply policies/policies_test.go]: No i get it, i suppose we can skip it for now to not delay the submition of the PR

File: policies/policies_test.go (hunk 16 above)
Comment (ganochenkodg) [reply policies/policies_test.go]: refactored the code

File: policies/policies_test.go (hunk 15 above)
Comment (ganochenkodg) [reply policies/policies_test.go]: Already realised this idea, now this test looks clean, finally:
```golang
			path := tt.pathFunc(t, tt.initialContent)

			err := writeIfChanged(ctx, tt.newContent, path)
			utiltest.AssertErrorMatchAndSkip(t, err, tt.wantErr)
			utiltest.AssertFileContents(t, path, string(tt.newContent))
```

ganochenkodg: This additional MatchAndSkip function could be useful for another tests with same logic

policies/policies_test.go
ganochenkodg: Done

burov (dismissed):

burov (dismissed):

petercieslak (approved):

burov (approved):

PR #952, opened 2026-04-16
Add unit tests for policies/yum.go

ganochenkodg:
New test coverage:
github.com/GoogleCloudPlatform/osconfig/policies/yum.go:68:		yumChanges		100.0%


policies/yum_test.go
```diff
@@ -80,3 +86,124 @@ func TestYumRepositories(t *testing.T) {
 		}
 	}
 }
+
+// TestYumChanges tests the yumChanges function, ensuring it correctly handles package installations, removals, and updates.
+func TestYumChanges(t *testing.T) {
+	rpmQueryArgs := []string{"--queryformat", `\{"architecture":"%{ARCH}","package":"%{NAME}","source_name":"%{SOURCERPM}","version":"%|EPOCH?{%{EPOCH}:}:{}|%{VERSION}-%{RELEASE}"\}` + "\n", "-a"}
+	yumCheckUpdateArgs := []string{"check-update", "--assumeyes"}
+	yumListUpdatesArgs := []string{"update", "--assumeno", "--color=never"}
```
burov, line 94: Why don't we use existing variables defined in yum.go ?

ganochenkodg: because we try to avoid changing the original code. they are placed in different package ("packages", not "policies") and start from small letters -> can't be imported without renaming. Should we update yum.go?

policies/yum_test.go
petercieslak, line 1: Please apply changes from comments for PR 949 and 951 if possible here and in other PR's in the chain as well

ganochenkodg: done

petercieslak (approved):

burov (approved):

PR #953, opened 2026-04-16
Add unit tests for policies/zypper.go

ganochenkodg:
New test coverage:
github.com/GoogleCloudPlatform/osconfig/policies/zypper.go:68:		zypperChanges		100.0%


policies/zypper_test.go
```diff
@@ -80,3 +86,119 @@ func TestZypperRepositories(t *testing.T) {
 		}
 	}
 }
+
+// TestZypperChanges tests the zypperChanges function, ensuring it correctly handles package installations, removals, and updates.
+func TestZypperChanges(t *testing.T) {
+	rpmQueryArgs := []string{"--queryformat", `\{"architecture":"%{ARCH}","package":"%{NAME}","source_name":"%{SOURCERPM}","version":"%|EPOCH?{%{EPOCH}:}:{}|%{VERSION}-%{RELEASE}"\}` + "\n", "-a"}
```
burov, line 92: Why don't we use https://github.com/GoogleCloudPlatform/osconfig/blob/master/packages/rpm.go#L41 ?

policies/zypper_test.go
```diff
@@ -80,3 +86,119 @@ func TestZypperRepositories(t *testing.T) {
...
+	zypperListUpdatesArgs := []string{"--gpg-auto-import-keys", "-q", "list-updates"}
```
burov, line 93: same question as above

policies/zypper_test.go
ganochenkodg: because we try to avoid changing the original code. they are placed in different package ("packages", not "policies") and start from small letters -> can't be imported without renaming. Should we update rpm.go?

burov (approved):

petercieslak (approved):

PR #954, opened 2026-04-16
Change disk size for deb11

petercieslak:
Deb11 presubmits are failing because of no disk space left. Let's try bigger disk size


burov (approved):

paulinakania (approved):
LGTM

PR #957, opened 2026-04-17
Add unit tests for policies/apt.go PART 2

ganochenkodg:
New test coverage:

github.com/GoogleCloudPlatform/osconfig/policies/apt.go:59:		getAptGPGKey		91.7%
github.com/GoogleCloudPlatform/osconfig/policies/apt.go:136:	aptRepositories		86.7%


policies/apt_test.go
```diff
@@ -31,120 +31,158 @@ import (
...
-		got, err := runAptRepositories(context.Background(), tt.repos)
-		if err != nil {
-			t.Fatal(err)
-		}
+			td, err := ioutil.TempDir(os.TempDir(), "")
+			if err != nil {
+				t.Fatal(err)
+			}
+			defer os.RemoveAll(td)
+			testRepo := filepath.Join(td, "testRepo")
```
burov: I think that might be a good idea to extract this logic to separate function to improve readability.

policies/apt_test.go
```diff
@@ -31,120 +31,158 @@ import (
...
 
-		if got != tt.want {
-			t.Errorf("%s: got:\n%q\nwant:\n%q", tt.name, got, tt.want)
-		}
+			err = aptRepositories(ctx, tt.repos, testRepo)
+			utiltest.AssertErrorMatch(t, err, nil)
```
burov, line 136: Do we plan to test negative case scenarios ?

policies/apt_test.go
ganochenkodg: Added helper TempDir in utiltest.go

policies/apt_test.go
ganochenkodg: Not yet, coverage for aptRepositories already reached 86%, maybe i will investigate negative cases later

policies/apt_test.go
```diff
@@ -31,120 +30,174 @@ import (
...
-		osInfoStub := stubOsInfoProvider{nameVersionProvider: tt.nameAndVersionProvider}
-		osInfoProvider = osInfoStub
+		t.Run(tt.name, func(t *testing.T) {
+			defer utiltest.OverrideVariable(&osInfoProvider, tt.provider)()
 
-		got, err := runAptRepositories(context.Background(), tt.repos)
-		if err != nil {
-			t.Fatal(err)
-		}
+			td, cleanup := utiltest.TempDir(t)
```
burov: We recently found https://pkg.go.dev/testing#T.TempDir

policies/apt_test.go
burov: Coverage is not ideal measure of how effective tests are, i suggest to cover negative cases as well.

policies/apt_test.go
```diff
@@ -31,120 +30,174 @@ import (
...
 
-	entityList, err := getAptGPGKey(key)
-	if err != nil {
-		t.Fatal(err)
+	tests := []struct {
+		name    string
+		url     string
+		wantErr error
+	}{
+		{
```
burov, line 265: Do we want to case where the actually valid key doesn't return an error ?

policies/apt_test.go
ganochenkodg: And removed it, switched now to t.TempDir

policies/apt_test.go
ganochenkodg: Yes, will use it

policies/apt_test.go
ganochenkodg: This case affects the coverage of getAptGPGKey, lets keep it

policies/apt_test.go
ganochenkodg: Added negative cases. Now we have changes in utiltest - helper utiltest.AssertFormatMatch. This helper is presented in sevral PR s and required to pass the tests. After any of those PRs with AssertFormatMatch will be merged I will fix all others tp avoid conflicts. And yes, clog errors contain timestamps, I can't directly compare such logs with expected string, unfortunately.

util/utiltest/utiltest.go
```diff
@@ -224,3 +224,11 @@ func WriteToTempFileMust(t *testing.T, filename string, content []byte) string {
 	}
 	return path
 }
+
+// NormalizeErrorType normalizes any error to a standard errors.New(err.Error()) or returns nil.
+func NormalizeErrorType(err error) error {
```
petercieslak: I would prefer to compare errors with the same type instead of flattening them to a generic `error`. If this is not possible, maybe it's better to compare error messages instead of making this transformation?

ganochenkodg: It's a way to remove if err != nil from test block. Let me try if I can find all possible error types to use them in testcases

ganochenkodg: done

petercieslak (dismissed):

policies/apt_test.go
```diff
@@ -15,140 +15,262 @@
...
-		name                   string
-		repos                  []*agentendpointpb.AptRepository
-		nameAndVersionProvider func() (string, string, string)
-		want                   string
+		name     string
+		repos    []*agentendpointpb.AptRepository
+		provider osinfo.Provider
+		repoFile string
+		wantRepo string
+		wantErr  string
```
burov: name does not represent the actual purpose of the field, wantErr usually used to show the actual error rather than regexp as it is right now, please use reasonable name

policies/apt_test.go
burov: I'm not sure you got my question, i don't see an actual case where we test that we provide valid case and it is actually behave as expected

ganochenkodg: added

policies/apt_test.go
ganochenkodg: fixed, wantErr -> wantLogsRegexp

burov (approved):

petercieslak (approved):

PR #958, opened 2026-04-20
Cover zypper_patch by unit tests

iliatsuprik:
Add more unit tests for ospatch/zypper_patch.go.

Coverage increased from 21.4% to 97.1%


ospatch/zypper_patch_test.go
```diff
@@ -78,6 +85,208 @@ func TestRunFilter(t *testing.T) {
...
+	tests := []struct {
+		name        string
+		description string
+		opts        []ZypperPatchOption
+		setupMock   func(ctx context.Context, mock *utilmocks.MockCommandRunner)
+		wantErr     error
+	}{
+		{
+			name:        "ZypperPatchesError",
+			description: "When listing available patches fails, RunZypperPatch should surface the wrapped command error and not attempt to install anything.",
```
burov: I don't think it is reasonable to have both `name` and `description`, let's rename `description` to `name` and keep only it.

burov (dismissed):

ospatch/zypper_patch_test.go
iliatsuprik: done

burov (approved):

petercieslak (approved):

PR #973, opened 2026-05-06
Align format of dates and timestamp collected across Windows packages

petercieslak:


agentendpoint/inventory.go
```diff
@@ -333,17 +333,23 @@ func wuaToInventoryItem(packages []*packages.WUAPackage) []*agentendpointpb.VmIn
 				"KbArticleId":              structpb.NewListValue(kbArticleIdsList),
 				"MoreInfoUrls":             structpb.NewListValue(moreInfoUrls),
 				"RevisionNumber":           structpb.NewNumberValue(float64(pkg.RevisionNumber)),
-				"LastDeploymentChangeTime": structpb.NewStringValue(pkg.LastDeploymentChangeTime.String()),
+				"LastDeploymentChangeTime": structpb.NewStringValue(pkg.LastDeploymentChangeTime.UTC().Format("2006-01-02 15:04:05 +0000 GMT")),
```
burov: Please move the format of the date into separate variable

agentendpoint/inventory.go
```diff
@@ -352,7 +358,7 @@ func qfeToInventoryItem(packages []*packages.QFEPackage) []*agentendpointpb.VmIn
 			Location: []string{},
 			Metadata: &structpb.Struct{Fields: map[string]*structpb.Value{
 				"Description": structpb.NewStringValue(pkg.Description),
```
burov, line 360: Login to format date might be extracted to separate functions.

agentendpoint/inventory_test.go
```diff
@@ -832,3 +832,84 @@ func Benchmark_computeStableFingerprint(b *testing.B) {
...
+	fields := got[0].Metadata.Fields
+	if fields["Publisher"].GetStringValue() != "TestPublisher" {
+		t.Errorf("Expected Publisher 'TestPublisher', got '%s'", fields["Publisher"].GetStringValue())
+	}
+	if fields["InstallDate"].GetStringValue() != "2026-04-28 14:30:00 +0000 GMT" {
+		t.Errorf("Expected InstallDate '2026-04-28 14:30:00 +0000 GMT', got '%s'", fields["InstallDate"].GetStringValue())
+	}
+	if fields["HelpLink"].GetStringValue() != "TestLink" {
+		t.Errorf("Expected HelpLink 'TestLink', got '%s'", fields["HelpLink"].GetStringValue())
+	}
```
burov: utiltest.AssertEquals should be helpful here.

agentendpoint/inventory.go
```diff
@@ -767,3 +774,22 @@ func fingerprintForPackage(pkg *agentendpointpb.Inventory_SoftwarePackage) strin
...
+		clog.Infof(ctx, "installedOn is empty, returning EPOCH")
+		return time.Time{}, nil
+	}
+	for _, layout := range layouts {
+		if t, err := time.Parse(layout, installedOn); err == nil {
+			return t, nil
+		}
+	}
+	return time.Time{}, fmt.Errorf("unable to parse date %q with any known layout", installedOn)
+}
```
burov, line 790: This function is handle some of the errors and propagate the others, what i suggest is to propagate all of the errors up to the stack and handle error and default value there.

agentendpoint/inventory.go
```diff
@@ -333,17 +334,23 @@ func wuaToInventoryItem(packages []*packages.WUAPackage) []*agentendpointpb.VmIn
...
-func qfeToInventoryItem(packages []*packages.QFEPackage) []*agentendpointpb.VmInventory_InventoryItem {
+func qfeToInventoryItem(ctx context.Context, packages []*packages.QFEPackage) []*agentendpointpb.VmInventory_InventoryItem {
 	qfeFormattedPackages := make([]*agentendpointpb.VmInventory_InventoryItem, len(packages))
 	for i, pkg := range packages {
+		installedOn := pkg.InstalledOn
+		t, err := parseQFEDate(ctx, pkg.InstalledOn)
+		if err != nil {
+			clog.Warningf(ctx, "Error parsing QFE InstalledOn date: %v", err)
+		}
+		installedOn = t.UTC().Format(dateTimeFormat)
```
burov: This code can be refactored to be more focused on what is actually doing, fist of all `installedOn := pkg.InstalledOn` is doing nothing, it is not default value, we will always override it, what i suggest is to either set default value to it on the start and override it only if error is equals to nil, or set installedOn to empty and override it when needed.

agentendpoint/inventory.go
petercieslak: The function does not handle any errors, it only handles a special case when the `installedOn` string is empty, then I return Golang default time Epoch. Putting all errors that occured during parsing can be misleading, for example if there is a string date in format "02-Jan-2006", the function will parse the string correctly to `time.Time`, but error stack will not be empty. It will be hard to handle such cases in a straightforward way if I was to move this logic somewhere else.

agentendpoint/inventory.go
petercieslak: Good catch, I've changed the behavior of `parseQFEDate` and forgot to change this codeblock. Thanks!

agentendpoint/inventory.go
petercieslak: I've corrected the function to only parse the date, it also returns Epoch date on error

agentendpoint/inventory.go
petercieslak: I've refactored the code

burov (approved):

paulinakania (approved):
LGTM

PR #974, opened 2026-05-10
Add unit tests for policies/recipes/steps.go PART 1

ganochenkodg:
First part of unit tests for steps.go:

github.com/GoogleCloudPlatform/osconfig/policies/recipes/steps.go:51:		stepCopyFile 0.0% -> 87.0%
github.com/GoogleCloudPlatform/osconfig/policies/recipes/steps.go:93:		parsePermissions 0.0 % -> 100.0%
github.com/GoogleCloudPlatform/osconfig/policies/recipes/steps.go:132:	ensureSymlinkBelongsToDir 0.0% -> 73.3%


policies/recipes/steps_test.go
```diff
@@ -216,3 +218,207 @@ func ensureTar(t *testing.T, dst string, entries []fileEntry) {
...
+			artifacts: map[string]string{"art2": filepath.Join(tmpDir, "missing")},
+			wantErr:   &os.PathError{Op: "open", Path: filepath.Join(tmpDir, "missing"), Err: fmt.Errorf("no such file or directory")},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Cleanup(func() {
+				os.Remove(destPath)
+			})
```
iliatsuprik, line 424: can this cleanup be moved out of the test cycle?

policies/recipes/steps_test.go
```diff
@@ -216,3 +218,207 @@ func ensureTar(t *testing.T, dst string, entries []fileEntry) {
...
+			if tt.setupFunc != nil {
+				tt.setupFunc()
+			}
```
iliatsuprik: nit: consider using setupFunc in each test case and call it without if

iliatsuprik (approved):

policies/recipes/steps_test.go
ganochenkodg: No, in each test iteration we do some operations with file destPath (creation/overwrite/changing perms) and this t.Cleanup removes our file after each test iteration

policies/recipes/steps_test.go
ganochenkodg: Good idea, will rework it

policies/recipes/steps_test.go
iliatsuprik: I mean this func is operating with variables created out of cycle, so in theory we should be able to move it there

policies/recipes/steps_test.go
ganochenkodg: fixed

policies/recipes/steps_test.go
```diff
@@ -216,3 +218,209 @@ func ensureTar(t *testing.T, dst string, entries []fileEntry) {
 		t.Errorf("unable to close file: %s", err)
 	}
 }
+
+// Test_parsePermissions verifies that octal permission strings are correctly parsed into os.FileMode.
+func Test_parsePermissions(t *testing.T) {
```
petercieslak, line 223: Let's stick to pascal case for test naming

policies/recipes/steps_test.go
```diff
@@ -216,3 +218,209 @@ func ensureTar(t *testing.T, dst string, entries []fileEntry) {
...
+	tests := []struct {
+		name      string
+		step      *agentendpointpb.SoftwareRecipe_Step_CopyFile
+		artifacts map[string]string
+		setupFunc func()
+		wantErr   error
+		wantPerm  os.FileMode
+	}{
+		{
+			name: "successful copy, want nil error and 0644",
```
petercieslak: please add "... and 0644 **permissions**"

policies/recipes/steps_test.go
```diff
@@ -216,3 +218,209 @@ func ensureTar(t *testing.T, dst string, entries []fileEntry) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Cleanup(func() {
+				os.Remove(destPath)
+			})
+			tt.setupFunc()
+
+			gotErr := stepCopyFile(tt.step, tt.artifacts, nil, "")
+			utiltest.AssertErrorMatchAndSkip(t, gotErr, tt.wantErr)
+			utiltest.AssertFileContents(t, destPath, content)
```
petercieslak: I think it would be also beneficial to check if the old file contents did not change in case of any error, changing the assert to `AssertErrorMatch` and adding another field `wantContent` to the test table should be enough

policies/recipes/steps_test.go
burov: That is actually common pattern to call Tests for private struct/functions in this way

policies/recipes/steps_test.go
burov: Yeah, i suppose we can call it when we create dest path, i don't think we actually need to remove it few times :) once is enough

policies/recipes/steps_test.go
burov: +1

ganochenkodg: Done, added an additionel helper in utiltest for this content checking and chenged test's logic aa you requested

policies/recipes/steps_test.go
ganochenkodg: fixed

policies/recipes/steps_test.go
ganochenkodg: So i don't have to change the name, right? We used such naming for privates in other places

policies/recipes/steps_test.go
ganochenkodg: Like this?
```
	destPath := filepath.Join(tmpDir, "dest")
	t.Cleanup(func() {
		os.Remove(destPath)
	})
```
It crashes the test's logic. On negative cases this file must not exist, but it could left there from previous positive cases when it was created and not cleaned. And then I get such errors:
```
    --- FAIL: Test_stepCopyFile/artifact_not_found,_want_find_error (0.00s)
        steps_test.go:428: Errors mismatch, want could not find location for artifact "unknown", got file already exists at path "/tmp/Test_stepCopyFile4111799713/001/dest" and Overwrite = false
        steps_test.go:429: File contents mismatch (-want +got):
              string(
            - 	"",
            + 	"artifact content",
              )
        steps_test.go:434: got != want (-want +got):
              fs.FileMode(
            - 	s"----------",
            + 	s"-rwxr-xr-x",
              )
    --- FAIL: Test_stepCopyFile/artifact_file_missing,_want_no_file_error (0.00s)
        steps_test.go:428: Errors mismatch, want open /tmp/Test_stepCopyFile4111799713/001/missing: no such file or directory, got file already exists at path "/tmp/Test_stepCopyFile4111799713/001/dest" and Overwrite = false
        steps_test.go:429: File contents mismatch (-want +got):
              string(
            - 	"",
            + 	"artifact content",
              )
        steps_test.go:434: got != want (-want +got):
              fs.FileMode(
            - 	s"----------",
            + 	s"-rwxr-xr-x",
              )
 ```

burov: I get it, what make it not obvious is that we create file once at the top, that why it force me to think that it can be just deleted once.

burov (approved):

petercieslak (approved):

PR #979, opened 2026-05-12
Add unit tests for config/config,go

bromivipo:
This PR contains unit tests for config.go
Initial coverage: 69%, current coverage 96,6%

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -17,7 +17,6 @@ package config
 import (
 	"context"
 	"errors"
-	"fmt"
 	"testing"
 
 	"cloud.google.com/go/osconfig/agentendpoint/apiv1/agentendpointpb"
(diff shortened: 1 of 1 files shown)


config/config_test.go
```diff
@@ -27,3 +33,123 @@ func init() {
...
+						State: agentendpointpb.OSPolicy_Resource_FileResource_ABSENT,
+					},
+				},
+			},
+		}
+	}
+
+	tests := []struct {
+		name      string
+		setupFunc func(t *testing.T) error
```
burov: Avoid putting type in the name of the variable

config/config_test.go
```diff
@@ -27,3 +33,123 @@ func init() {
 	packages.ZypperExists = true
 	packages.MSIExists = true
 }
+
+func TestOSPolicyResourceMethods(t *testing.T) {
+	ctx := context.Background()
+
+	// Helper to create a fresh valid OSPolicyResource
+	newPR := func(path string) *OSPolicyResource {
```
burov: newPR does not give enough information about the function, please find the name that give context.

config/config_test.go
```diff
@@ -27,3 +33,123 @@ func init() {
...
+					},
+				}
+				return pr.Validate(ctx)
+			},
+			wantErr: errors.New("ResourceType field not set"),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := tt.setupFunc(t)
```
burov: Set-up and execution should not be mixed in the single function.

config/config_test.go
```diff
@@ -27,3 +33,123 @@ func init() {
 	packages.ZypperExists = true
 	packages.MSIExists = true
 }
+
+func TestOSPolicyResourceMethods(t *testing.T) {
+	ctx := context.Background()
```
burov: t.Context() can be used here

config/config_test.go
burov: I don't think setup function is useful here, basically we can just store, the OSPolicy_Resource and have a separate field that incapsulate operations that need to be executed.

config/config_test.go
petercieslak: Please move this helper to a separate function instead of a lambda

config/config_test.go
```diff
@@ -27,3 +33,123 @@ func init() {
...
+	}
+
+	tests := []struct {
+		name      string
+		setupFunc func(t *testing.T) error
+		wantErr   error
+	}{
+		{
+			name: "CheckState call before Validate, expect run before Validate error",
+			setupFunc: func(t *testing.T) error {
```
petercieslak: I'm not sure if I like the idea that setup function contains test logic. Setup functions should only prepare test preconditions, this setup functions are also running the function that we want to test. We can easily separate those cases into separate test functions and move code that we want to test from `setupFunc` to the `t.Run(...){...}` clause so it is more visible what exactly are we testing

petercieslak: You can add `pr` variable to the test table:
`tests := []struct {
		name      string
		setupFunc func(t *testing.T) error
         pr            &OSPolicyResource
		wantErr   error
	}`

 and perform any setup actions in the existing `setupFunc`, but move the return function calls from `setupFunc` to test body

config/config_test.go
bromivipo: Fixed

config/config_test.go
bromivipo: Renamed `setupFunc -> setup`. Also added a separate field for policy resources and refactored setup so it now only consists of the preparations and the validation call itself.

config/config_test.go
bromivipo: Renamed `newPR -> newTestOSPolicyFileResource`

config/config_test.go
bromivipo: Refactored the tests as suggested by Piotr: split the test into separate functions that target only one method each so it is easy to remove the actual method execution from setup.

config/config_test.go
bromivipo: Replaced all occurrences of `context.Background()` with `t.Context()`

config/config_test.go
bromivipo: Refactored the tests as suggested: split the test into separate functions that target only one method each.

petercieslak (dismissed):

config/config_test.go
```diff
@@ -27,3 +32,188 @@ func init() {
...
+				pr.OSPolicy_Resource.GetFile().Path = tmpPath
+				return pr.Validate(ctx)
+			},
+			pr:      newTestOSPolicyFileResource(),
+			wantErr: nil,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.setup != nil {
```
burov: I slightly prefer in tests just set a blank setup function rather than do nil check

config/config_test.go
```diff
@@ -27,3 +32,188 @@ func init() {
...
+			setup: func(t *testing.T, pr *OSPolicyResource) error {
+				return pr.Validate(ctx)
+			},
+			pr:      newTestOSPolicyFileResource(),
+			wantErr: nil,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.setup != nil {
```
burov: Same comment as above

config/config_test.go
```diff
@@ -27,3 +32,188 @@ func init() {
...
```
burov: Same comment as below

burov (commented):
Looks good, small comment on the setup function, other than that looks good

config/config_test.go
bromivipo: fixed it in all cases

burov (approved):

petercieslak (approved):

PR #983, opened 2026-05-14
Add unit tests for agentendpoint/agentendpoint_beta.go

bromivipo:
This PR contains unit tests for agentendpoint_beta.go
Initial coverage: 0%, current coverage 69%


agentendpoint/agentendpoint_beta_test.go
```diff
@@ -0,0 +1,146 @@
...
+			defer tc.s.Stop()
+
+			policy, err := tc.client.LookupEffectiveGuestPolicies(ctx)
+			if (err != nil) != (tt.wantErr != nil) {
+				t.Fatalf("LookupEffectiveGuestPolicies() unexpected error state: got err = %v, wantErr = %v", err, tt.wantErr)
+			}
+
+			if tt.wantErr == nil && policy == nil {
+				t.Fatal("LookupEffectiveGuestPolicies() returned nil policy, want non-nil")
+			}
```
ganochenkodg: Too complicated, tests should be more straightforward. I would suggest to use 2 params - wantErr and wantPolicy - and compare them with AssertErrorMatch and AssertEquals, something like:
```golang
gotPolicy, gotErr := tc.client.LookupEffectiveGuestPolicies(ctx)
utiltest.AssertErrorMatch(t, gotErr, tt.wantErr)
utiltest.AssertEquals(t, gotPolicy, tt.wantPolicy)
```

agentendpoint/agentendpoint_beta_test.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+type testBetaClient struct {
+	client *BetaClient
+	s      *grpc.Server
+}
+
+func newBetaTestClient(ctx context.Context, srv agentendpointpb.AgentEndpointServiceServer) (*testBetaClient, error) {
+	lis := bufconn.Listen(bufSize)
+	s := grpc.NewServer()
```
ganochenkodg: Lets use more readable variable names

bromivipo: renamed to listener and server

agentendpoint/agentendpoint_beta_test.go
bromivipo: Slightly changed the test case to call AssertErrorMatch. But calling AssertEquals won't work as expected because policies are Protobuf messages. I can either compare it in the test case manually (we can use the same func as in AssertEquals, but with an extra option) or just check whether the policy is non-nil. I believe the second option is more suitable.

agentendpoint/agentendpoint_beta_test.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func (c *testBetaClient) close() {
+	c.client.Close()
+	c.server.Stop()
+}
+
+func newBetaTestClient(ctx context.Context, srv agentendpointpb.AgentEndpointServiceServer) (*testBetaClient, error) {
+	listener := bufconn.Listen(bufSize)
+	server := grpc.NewServer()
```
petercieslak: Using mocks instead of setting up the grpc server would simplify this code. Take a look at https://github.com/GoogleCloudPlatform/osconfig/blob/master/util/mocks/mock_agent_endpoint_client.go
and
https://github.com/GoogleCloudPlatform/osconfig/blob/master/agentendpoint/inventory_test.go#L564

bromivipo: thanks for the links. I've switched to mocks

agentendpoint/agentendpoint_beta_test.go
```diff
@@ -0,0 +1,91 @@
...
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockClient := utilmocks.NewMockAgentEndpointBetaClient(ctrl)
+	client := &BetaClient{raw: mockClient, noti: make(chan struct{}, 1)}
+
+	tests := []struct {
+		name       string
+		mockPolicy *agentendpointpb.EffectiveGuestPolicy
+		mockErr    error
```
petercieslak: Hm, naming suggests that policy and error are mocks, I would rather call it just `policy` and  `err`

petercieslak: On a second though, lets change policy to `wantPolicy` and leave `mockErr`, sorry

bromivipo: Changed

petercieslak (dismissed):

agentendpoint/agentendpoint_beta_test.go
```diff
@@ -0,0 +1,91 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockClient.EXPECT().LookupEffectiveGuestPolicy(gomock.Any(), gomock.Any()).Return(tt.wantPolicy, tt.mockErr)
+
+			gotPolicy, gotErr := client.LookupEffectiveGuestPolicies(ctx)
+			utiltest.AssertErrorMatch(t, gotErr, tt.wantErr)
+			if diff := cmp.Diff(tt.wantPolicy, gotPolicy, protocmp.Transform()); diff != "" {
```
burov: Why don't we use existing functions from the utitest? If there some problems, can we try to adopt existing solution first ?

agentendpoint/agentendpoint_beta_test.go
```diff
@@ -0,0 +1,91 @@
...
+			name:       "server returns error, expect error",
+			wantPolicy: nil,
+			mockErr:    errors.New("mock error"),
+			wantErr:    fmt.Errorf("error calling LookupEffectiveGuestPolicies: %w", errors.New("mock error")),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockClient.EXPECT().LookupEffectiveGuestPolicy(gomock.Any(), gomock.Any()).Return(tt.wantPolicy, tt.mockErr)
```
burov: I think usage of setup function will simplify things here

agentendpoint/agentendpoint_test.go
```diff
@@ -334,8 +335,6 @@ func TestLoadPatchTaskFromState(t *testing.T) {
 	defer os.RemoveAll(td)
 	taskStateFile = filepath.Join(td, "testState")
 
-	srv.streamSend <- struct{}{}
```
bromivipo, line 337: It looks like this line is redundant and causes a race condition. It triggers a flow that is later expected not to run, making the test flaky. Correct me if I'm wrong.

petercieslak: What flow are you exactly talking about here? Can you provide some logs? 

From what I see, this code sends an empty state to the channel, which is later checked in line 339.

bromivipo: As far as I understand, this test is meant to verify that an interrupted task can be successfully resumed from the local state file, so it shouldn't actually need a server notification. It seems that on line 339, it's only checking whether it can read from the local state file.

If we send to streamSend, it forces a server notification [on line 180](https://github.com/GoogleCloudPlatform/osconfig/blob/da0fe81d6c72041d7c4d7a5df759211ab3e03856/agentendpoint/agentendpoint_test.go#L180). This makes the agent call `StartNextTask` to check for new tasks. In the mock server `StartNextTask` [sets srv.TaskStart flag to true](https://github.com/GoogleCloudPlatform/osconfig/blob/master/agentendpoint/agentendpoint_test.go#L192) and later in the test we check, that it is false on [line 343](https://github.com/GoogleCloudPlatform/osconfig/blob/master/agentendpoint/agentendpoint_test.go#L343) and [line 371](https://github.com/GoogleCloudPlatform/osconfig/blob/master/agentendpoint/agentendpoint_test.go#L371). 

That's why this line seems redundant. Does that make sense, or am I missing something?

agentendpoint/agentendpoint_beta_test.go
bromivipo: utiltest.AssertEquals didn't work correctly with proto structures. I've updated the function by adding the protocmp.Transform() option to it, and then used it here.

agentendpoint/agentendpoint_beta_test.go
bromivipo: moved mock.Expect into setup func and removed mockError field from the table

agentendpoint/agentendpoint_test.go
petercieslak: Yep, makes sense to limit this test only to reading from state. I've doubled checked the behavior and looks fine to me to not send the server notification. Thank you :)

petercieslak (dismissed):

petercieslak (approved):

burov (approved):

PR #984, opened 2026-05-15
Add test cases for agentendpoint/task_state.go

bromivipo:
This PR contains unit tests for task_state.go
Initial coverage: 77.8%, current coverage 88.9%

### DIFF
--- a/agentendpoint/task_state_test.go
+++ b/agentendpoint/task_state_test.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


agentendpoint/task_state_test.go
```diff
@@ -151,41 +157,91 @@ func TestStateSave(t *testing.T) {
...
-		state *taskState
-		want  string
+		desc    string
+		state   *taskState
+		path    string
+		want    string
+		wantErr bool
 	}{
 		{
 			"NilState",
```
ganochenkodg: Use field names please, it's hard to read tests.

agentendpoint/task_state_test.go
```diff
@@ -151,41 +157,91 @@ func TestStateSave(t *testing.T) {
...
+			true,
 		},
 	}
 	for _, tt := range tests {
-		err := tt.state.save(testState)
+		err := tt.state.save(tt.path)
+		if (err != nil) != tt.wantErr {
+			t.Errorf("%s: unexpected save error state: %v, wantErr: %v", tt.desc, err, tt.wantErr)
+			continue
+		}
```
ganochenkodg: Could you rewrite this test next way: 
1. wantErr must be Error type, not bool
2. use AssertErrorMatch to compare returned errors with predictable error messages and tt.wantErr

agentendpoint/task_state_test.go
bromivipo: done

agentendpoint/task_state_test.go
bromivipo: Rewrote it like that, but in one case the path of a temp file is part of the error, so I update it dynamically during the test execution.

agentendpoint/task_state_test.go
```diff
@@ -145,55 +155,71 @@ func TestLoadOldState(t *testing.T) {
...
-		if err != nil {
-			t.Errorf("%s: error reading state: %v", tt.desc, err)
-			continue
-		}
-
-		if string(got) != tt.want {
-			t.Errorf("%s:\ngot:\n%q\nwant:\n%q", tt.desc, got, tt.want)
-		}
+		t.Run(tt.desc, func(t *testing.T) {
+			err := tt.state.save(tt.path)
```
petercieslak: lets rename `err` to `gotErr` to indicate in test that this is the error used for assertions

agentendpoint/task_state_test.go
```diff
@@ -55,6 +61,10 @@ func TestLoadState(t *testing.T) {
 	defer os.RemoveAll(td)
 	testState := filepath.Join(td, "testState")
 
+	// test read error by attempting to read a directory
+	_, err = loadState(td)
+	utiltest.AssertErrorMatch(t, err, &fs.PathError{Op: "read", Path: td, Err: errors.New("is a directory")})
```
petercieslak: This test can be moved to the test table below, so we avoid doing assertions outside of the test "body" (encapsulated with `t.Run(...){...}`) and ensure code cleanness. It would require adding a `setup` field in the test table, which would be a setup function for each test case. Then you could move `ioutil.WriteFile(testState, tt.state, 0600)` to the setup function.  I don't think that we need to check the exact error in this case, so the rest of the test table and test body does not have to change.

bromivipo: Moved this case to the test table. Added tt.setup and removed tt.state since it is used only in setup. Also named all the parameters in the table so it is more readable now.

petercieslak (dismissed):

agentendpoint/task_state_test.go
```diff
@@ -63,33 +69,53 @@ func TestLoadState(t *testing.T) {
...
+		setup   func(t *testing.T) string
 		wantErr bool
 		want    *taskState
 	}{
 		{
-			"BlankState",
-			[]byte("{}"),
-			false,
-			&taskState{},
+			name: "BlankState",
```
burov: Please name test cases in the format, "input, want state/output"

agentendpoint/task_state_test.go
```diff
@@ -99,14 +125,19 @@ func TestLoadState(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "ReadDirectoryError",
+			setup: func(t *testing.T) string {
+				return td
+			},
+			wantErr: true,
```
burov: Please check an actual error rather than just it presence.

petercieslak (approved):

agentendpoint/task_state_test.go
```diff
@@ -99,20 +126,20 @@ func TestLoadState(t *testing.T) {
...
-				t.Fatalf("unexpected error: %v", err)
-			}
-			if err == nil && tt.wantErr {
-				t.Fatalf("expected error")
-			}
+			path := tt.setup(t)
+			st, err := loadState(path)
+			utiltest.AssertErrorMatch(t, err, tt.wantErr)
 
 			if diff := cmp.Diff(tt.want, st, cmpopts.IgnoreUnexported(patchTask{}), protocmp.Transform()); diff != "" {
```
burov, line 144: I suppose it might be reasonable to contribute new function to utiltest

burov (approved):

PR #986, opened 2026-05-19
Add test cases for clog/clog.go

bromivipo:
This PR contains unit tests for task_state.go
Initial coverage: 51.9%, current coverage 96.3%


clog/clog_test.go
```diff
@@ -44,3 +50,159 @@ func TestWithLabels(t *testing.T) {
...
+			req:      nil,
+			resp:     nil,
+			expected: "",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tw := &testWriter{}
+			ctx := context.Background()
```
ganochenkodg: You can move it outside of the loop block. Lets keep in loops only specific preparations that require testcase values and tested functions + result checking. Common preparations should be placed somewhere outside of the loop, in the beginning of Test function.

clog/clog_test.go
```diff
@@ -44,3 +50,159 @@ func TestWithLabels(t *testing.T) {
...
+						break
+					}
+				}
+				if !found {
+					t.Errorf("expected to find %q in logs, but did not. Logs: %v", tt.expected, tw.logs)
+				}
+			} else {
+				if len(tw.logs) > 0 {
+					t.Errorf("expected no logs, but got: %v", tw.logs)
+				}
```
ganochenkodg: AssertEquals(t, len_tw.logs), 0) would be better

clog/clog_test.go
```diff
@@ -44,3 +50,159 @@ func TestWithLabels(t *testing.T) {
...
+				if strings.Contains(logLine, tt.expected) {
+					found = true
+					break
+				}
+			}
+			if !found {
+				t.Errorf("expected to find %q in logs, but did not. Logs: %v", tt.expected, tw.logs)
+			}
+		})
+	}
```
ganochenkodg: Just join all logs in one string and use AssertFormatMatch (regexp comparison helper) to check if logs include expected text. And the same for TestDebugRPC

clog/clog_test.go
bromivipo: Fixed it and added helper function for test-logger initialization to avoid duplications

clog/clog_test.go
bromivipo: Done

clog/clog_test.go
bromivipo: If we join all logs into a single string, it will be harder to validate the test case where nothing should be logged. But I agree that the slice was unnecessary. I've replaced it with a string and I'm clearing it at the beginning of each test case, so assertions are easier.

clog/clog_test.go
```diff
@@ -44,3 +51,132 @@ func TestWithLabels(t *testing.T) {
...
+			expected: "",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tw.logs = ""
+			DebugRPC(ctx, tt.name, tt.req, tt.resp)
+
+			if tt.expected != "" {
```
petercieslak: Using `if{} else{}` clauses in unit tests introduces some additional logic which is an antipattern. Can assertions be simplified here to avoid using conditional statements?

bromivipo: Sure. I changed `expected` to a regex matching an empty string for that test case, allowing all cases to use the same assertion function.

petercieslak (dismissed):

clog/clog_test.go
```diff
@@ -44,3 +50,127 @@ func TestWithLabels(t *testing.T) {
...
+	logs string
+}
+
+func (w *testWriter) Write(p []byte) (n int, err error) {
+	w.logs = string(p)
+	return len(p), nil
+}
+
+// Initializes logger and returns testWriter
+func initTestLogger(t *testing.T, ctx context.Context) *testWriter {
```
petercieslak: from golint:

"clog/clog_test.go:64:1: context.Context should be the first parameter of a function
Found 1 lint suggestions; failing."

bromivipo: fixed

clog/clog_test.go
```diff
@@ -44,3 +50,127 @@ func TestWithLabels(t *testing.T) {
...
+	})
+	if err != nil {
+		t.Fatalf("logger.Init error: %v", err)
+	}
+	return tw
+}
+
+func TestDebugRPC(t *testing.T) {
+	DebugEnabled = true
+	defer func() { DebugEnabled = false }()
```
burov: [If i understand it correctlyt ](https://github.com/GoogleCloudPlatform/osconfig/blob/master/util/utiltest/utiltest.go#L187) can be helpful for the purpose of override of the global variable

petercieslak (dismissed):

clog/clog_test.go
```diff
@@ -44,3 +50,127 @@ func TestWithLabels(t *testing.T) {
...
+	ctx := context.Background()
+	tw := initTestLogger(ctx, t)
+	req := wrapperspb.String("request")
+	resp := wrapperspb.String("response")
+
+	tests := []struct {
+		name     string
+		req      proto.Message
+		resp     proto.Message
+		expected string
```
burov: We stick to the pattern got/want rather than expected, please stick to the consistent pattern

clog/clog_test.go
```diff
@@ -44,3 +50,127 @@ func TestWithLabels(t *testing.T) {
...
+	}{
+		{
+			name:     "ReqAndRespMethod",
+			req:      req,
+			resp:     resp,
+			expected: `^\[Debug\] Called: ReqAndRespMethod with request:`,
```
burov: I don't think regexp here cover an actual log, if we will not log response here test still will path

clog/clog_test.go
bromivipo: fixed, thank you

clog/clog_test.go
bromivipo: fixed

clog/clog_test.go
bromivipo: modified the regexps to check both request and response messages

petercieslak (approved):

burov (approved):

PR #988, opened 2026-05-20
Add test cases for config/file_resource.go

bromivipo:
This PR contains unit tests for file_resource.go
Initial coverage: 67.1%, current coverage 86.6%

### DIFF
--- a/config/file_resource_test.go
+++ b/config/file_resource_test.go
@@ -17,9 +17,12 @@ package config
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
 	"testing"
 
 	"github.com/GoogleCloudPlatform/osconfig/util"
@@ -382,9 +385,11 @@ func TestFileResourceEnforceStatePresent(t *testing.T) {
 
 func TestFileResourceValidateErrors(t *testing.T) {
 	ctx := context.Background()
+	_, errLocalPath := os.Open("does_not_exist")
 	tests := []struct {
-		name string
-		fr   *fileResource
+		name    string
+		fr      *fileResource
+		wantErr error
 	}{
 		{
 			name: "InvalidState",
@@ -393,6 +398,7 @@ func TestFileResourceValidateErrors(t *testing.T) {
 					State: agentendpointpb.OSPolicy_Resource_FileResource_DESIRED_STATE_UNSPECIFIED,
 				},
 			},
+			wantErr: fmt.Errorf("unrecognized DesiredState for FileResource: %q", agentendpointpb.OSPolicy_Resource_FileResource_DESIRED_STATE_UNSPECIFIED),
 		},
 		{
 			name: "InvalidPermissions",
@@ -402,6 +408,7 @@ func TestFileResourceValidateErrors(t *testing.T) {
 					Permissions: "invalid",
 				},
 			},
+			wantErr: fmt.Errorf("can't parse permissions %q: %v", "invalid", &strconv.NumError{Func: "ParseUint", Num: "invalid", Err: strconv.ErrSyntax}),
 		},
 		{
 			name: "LocalPathDoesNotExist",
@@ -418,6 +425,7 @@ func TestFileResourceValidateErrors(t *testing.T) {
 					},
 				},
 			},
+			wantErr: errLocalPath,
 		},
 		{
 			name: "UnrecognizedSource",
@@ -428,6 +436,7 @@ func TestFileResourceValidateErrors(t *testing.T) {
 					Source: nil,
 				},
 			},
+			wantErr: errors.New("unrecognized Source type for FileResource: %!q(<nil>)"),
 		},
 		{
 			name: "ContentDownloadError",
@@ -440,6 +449,7 @@ func TestFileResourceValidateErrors(t *testing.T) {
 					},
 				},
 			},
+			wantErr: nil, // overridden dynamically below
 		},
 		{
 			name: "RemoteDownloadError",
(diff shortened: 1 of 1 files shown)


config/file_resource_test.go
```diff
@@ -377,3 +382,200 @@ func TestFileResourceEnforceStatePresent(t *testing.T) {
...
+			fr: &fileResource{
+				OSPolicy_Resource_FileResource: &agentendpointpb.OSPolicy_Resource_FileResource{
+					State:       agentendpointpb.OSPolicy_Resource_FileResource_PRESENT,
+					Permissions: "invalid",
+				},
+			},
+			wantErr: fmt.Errorf("can't parse permissions %q: %v", "invalid", &strconv.NumError{Func: "ParseUint", Num: "invalid", Err: strconv.ErrSyntax}),
+		},
+		{
+			name: "LocalPathDoesNotExist",
```
iliatsuprik: could you use names in format '<what is set in setup>, <what is expected by the test>', for example 'local file does not exist, failed to open file'.
Could you use this format for all tests in this PR and other your PRs as well

config/file_resource_test.go
```diff
@@ -377,3 +382,200 @@ func TestFileResourceEnforceStatePresent(t *testing.T) {
...
+			wantErr: &url.Error{Op: "Get", URL: "doesnot/exist", Err: errors.New("unsupported protocol scheme \"\"")},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := tt.fr.validate(ctx)
+			if tt.name == "ContentDownloadError" && err != nil {
+				tt.wantErr = err
+			}
```
iliatsuprik: why not specify wantErr in ContentDownloadError test setup? otherwise we compare err with err
could you remove this if statement

config/file_resource_test.go
```diff
@@ -377,3 +382,200 @@ func TestFileResourceEnforceStatePresent(t *testing.T) {
...
+func TestFileResourceEnforceStateErrors(t *testing.T) {
+	ctx := context.Background()
+	errRemove := os.Remove("/path/does/not/exist")
+	tests := []struct {
+		name    string
+		fr      *fileResource
+		wantErr error
+	}{
+		{
+			name: "InvalidState",
```
iliatsuprik: please use names in format like <what is set>, <what is expected>

config/file_resource_test.go
```diff
@@ -377,3 +382,200 @@ func TestFileResourceEnforceStatePresent(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			inDesiredState, err := tt.fr.enforceState(ctx)
+			utiltest.AssertErrorMatch(t, err, tt.wantErr)
+			if tt.wantErr != nil {
+				return
+			}
```
iliatsuprik: this if statement seems to be not needed, assertions should compare empty strings/booleans correctly

iliatsuprik (approved):

config/file_resource_test.go
bromivipo: fixed

config/file_resource_test.go
bromivipo: this error has several fields containing a dynamic temp file name, so I cannot specify the exact expected error before test execution. I agree, that this overwrite does not make sense though. 
To fix it I've specified a type of  wantErr and one field, but other fields should still be overwritten dynamically to use AssertErrorMatch. I could use some advice on how to deal with this type of errors. Is there a cleaner pattern to assert these errors?

config/file_resource_test.go
```diff
@@ -377,3 +382,216 @@ func TestFileResourceEnforceStatePresent(t *testing.T) {
 		t.Fatal("Repo file contents do not match after enforcement")
 	}
 }
+
+func TestFileResourceValidateErrors(t *testing.T) {
```
burov: Why do we test here only negative cases and there is no positive scenarios ?

config/file_resource_test.go
```diff
@@ -377,3 +382,216 @@ func TestFileResourceEnforceStatePresent(t *testing.T) {
...
+					ResourceType: &agentendpointpb.OSPolicy_Resource_File_{File: tt.frpb},
+				},
+			}
+			gotErr := pr.Validate(ctx)
+			utiltest.AssertErrorMatch(t, gotErr, tt.wantErr)
+		})
+	}
+}
+
+func TestFileResourceEnforceStateErrors(t *testing.T) {
```
burov, line 449: The same question, why don't we have any positive case scenario ?

config/file_resource_test.go
```diff
@@ -377,3 +382,216 @@ func TestFileResourceEnforceStatePresent(t *testing.T) {
...
+						ResourceType: &agentendpointpb.OSPolicy_Resource_File_{
+							File: &agentendpointpb.OSPolicy_Resource_FileResource{
+								Path:  missingPath,
+								State: agentendpointpb.OSPolicy_Resource_FileResource_ABSENT,
+							},
+						},
+					},
+				}
+				t.Cleanup(func() { pr.Cleanup(ctx) })
+				_ = pr.Validate(ctx)
```
burov: Did you thought about setting up resources in the way you don't have to call Validate ?

config/file_resource_test.go
bromivipo: Positive cases were covered in one of the tests above. I've merged my new test function with the existing one so everything is grouped together now and there's no misunderstanding.

config/file_resource_test.go
bromivipo: Positive cases are already covered in the two tests above (TestFileResourceEnforceStateAbsent and TestFileResourceEnforceStatePresent). I don't think merging them into a table-driven test is a good idea, as they have different verification flows and readability would suffer.

config/file_resource_test.go
bromivipo: I've now refactored it to set up the resources directly. I used Validate initially to match the pattern in the existing tests for this file.

config/file_resource_test.go
```diff
@@ -39,37 +44,42 @@ func TestFileResourceValidate(t *testing.T) {
...
 	}
 
+	_, errLocalPath := os.Open("does_not_exist")
+
 	var tests = []struct {
-		name   string
-		frpb   *agentendpointpb.OSPolicy_Resource_FileResource
-		wantMR ManagedFile
+		name    string
+		frpb    *agentendpointpb.OSPolicy_Resource_FileResource
```
burov: Longer name should increase readability of the code, frpb does not give any context, as you're touching the code can you make it fileResource, the same for wantMR, wantManagedFile. It should improve readability.

burov (dismissed):
LGTM to me, one small nit comment and we can merge it

config/file_resource_test.go
bromivipo: Done. I've renamed `wantMR` to `wantManagedFile`. For `frpb`, I used `fileResourcePB` since `fileResource` conflicts with the existing type name.

burov (approved):

paulinakania (approved):
LGTM

PR #989, opened 2026-05-21
Add test cases for policies/recipes/recipedb.go

bromivipo:
This PR contains unit tests for file_resource.go
Initial coverage: 0%, current coverage 79.1%

### DIFF
--- a/policies/recipes/recipedb_test.go
+++ b/policies/recipes/recipedb_test.go
@@ -37,42 +37,44 @@ func setupTestDB(t *testing.T) string {
 
 func TestNewRecipeDB(t *testing.T) {
 	tests := []struct {
-		name        string
-		fileContent string
-		wantErr     error
-		wantSize    int
+		name     string
+		setup    func(t *testing.T)
+		wantErr  error
+		wantSize int
 	}{
 		{
-			name:        "no file exists, expect no error and empty db",
-			fileContent: "",
-			wantErr:     nil,
-			wantSize:    0,
+			name:     "no file exists, expect no error and empty db",
+			setup:    func(t *testing.T) { setupTestDB(t) },
+			wantErr:  nil,
+			wantSize: 0,
 		},
 		{
-			name:        "valid JSON, expect no error and db size 2",
-			fileContent: `[{"Name":"recipe1","Version":[1],"InstallTime":1,"Success":true},{"Name":"recipe2","Version":[2],"InstallTime":2,"Success":false}]`,
-			wantErr:     nil,
-			wantSize:    2,
+			name: "valid JSON, expect no error and db size 2",
+			setup: func(t *testing.T) {
+				tmpDir := setupTestDB(t)
+				if err := os.WriteFile(filepath.Join(tmpDir, dbFileName), []byte(`[{"Name":"recipe1","Version":[1],"InstallTime":1,"Success":true},{"Name":"recipe2","Version":[2],"InstallTime":2,"Success":false}]`), 0644); err != nil {
+					t.Fatalf("failed to write test file: %v", err)
+				}
+			},
+			wantErr:  nil,
+			wantSize: 2,
 		},
 		{
-			name:        "invalid JSON, expect syntax error",
-			fileContent: `invalid json`,
-			wantErr:     func() error { var dummy interface{}; return json.Unmarshal([]byte(`invalid json`), &dummy) }(),
-			wantSize:    0,
+			name: "invalid JSON, expect syntax error",
+			setup: func(t *testing.T) {
+				tmpDir := setupTestDB(t)
+				if err := os.WriteFile(filepath.Join(tmpDir, dbFileName), []byte(`invalid json`), 0644); err != nil {
+					t.Fatalf("failed to write test file: %v", err)
+				}
+			},
+			wantErr:  func() error { var dummy interface{}; return json.Unmarshal([]byte(`invalid json`), &dummy) }(),
+			wantSize: 0,
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			tmpDir := setupTestDB(t)
-			dbFilePath := filepath.Join(tmpDir, dbFileName)
-
-			if tt.fileContent != "" {
-				if err := os.WriteFile(dbFilePath, []byte(tt.fileContent), 0644); err != nil {
-					t.Fatalf("failed to write test file: %v", err)
-				}
-			}
-
+			tt.setup(t)
 			db, err := newRecipeDB()
 			utiltest.AssertErrorMatch(t, err, tt.wantErr)
 			utiltest.AssertEquals(t, len(db), tt.wantSize)
@@ -80,7 +82,7 @@ func TestNewRecipeDB(t *testing.T) {
 	}
 }
 
-func TestAddRecipe(t *testing.T) {
+func TestAddRecipe_InMemory(t *testing.T) {
 	tests := []struct {
 		name       string
 		recipeName string
@@ -96,7 +98,7 @@ func TestAddRecipe(t *testing.T) {
 			wantErr:    nil,
 		},
 		{
-			name:       "invalid version, expect error",
+			name:       "invalid version, expect invalid Version string error",
 			recipeName: "bad-recipe",
 			version:    "invalid.version",
 			success:    false,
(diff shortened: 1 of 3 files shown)


policies/recipes/recipedb_test.go
```diff
@@ -0,0 +1,133 @@
...
+			name:        "invalid JSON, expect syntax error",
+			fileContent: `invalid json`,
+			wantErr:     func() error { var dummy interface{}; return json.Unmarshal([]byte(`invalid json`), &dummy) }(),
+			wantSize:    0,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tmpDir := setupTestDB(t)
```
iliatsuprik: can you extract this into a function or move to test setup. We need to make the test case shorter

policies/recipes/recipedb_test.go
```diff
@@ -0,0 +1,133 @@
...
+			utiltest.AssertEquals(t, r.Success, tt.success)
+
+			readDB, err := newRecipeDB()
+			if err != nil {
+				t.Fatalf("failed to read db file: %v", err)
+			}
+
+			readRecipe, ok := readDB.getRecipe(tt.recipeName)
+			utiltest.AssertEquals(t, ok, true)
+			utiltest.AssertEquals(t, readRecipe, r)
```
iliatsuprik: Consider to make this test validation shorter, maybe split into a few tests

iliatsuprik (approved):

policies/recipes/recipedb_test.go
bromivipo: I can't use a single assertion for the whole recipe since it has an installation time field. I've split into two separate tests, but it still is not really short. Please, check If this version looks better to you.

policies/recipes/recipedb_test.go
```diff
@@ -0,0 +1,137 @@
...
+	t.Helper()
+	tmpDir := t.TempDir()
+
+	utiltest.OverrideVariable(t, &dbDirWindows, tmpDir)
+	utiltest.OverrideVariable(t, &dbDirUnix, tmpDir)
+
+	return tmpDir
+}
+
+func TestNewRecipeDB(t *testing.T) {
```
burov: I see that in this test we're not testing actual content but rather the length, i don't think it is a good pattern as we're getting coverage without actually checking correctness of the code, if somebody will change the code in the way that entries will be corrupted tests still will path, please avoid writing tests to only bump the coverage, tests should verify that behavior correct rather than code is called.

policies/recipes/recipedb_test.go
```diff
@@ -0,0 +1,137 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			tt.setup(t)
+			db, gotErr := newRecipeDB()
+			utiltest.AssertErrorMatch(t, gotErr, tt.wantErr)
+			utiltest.AssertEquals(t, len(db), tt.wantSize)
+		})
+	}
+}
+
+func TestAddRecipe_InMemory(t *testing.T) {
```
burov: Test is not actually testing non-empty db, that means if we broke the code in the way to only keep the latest entry we add, the test will path but behavior is broken.

policies/recipes/recipedb_test.go
```diff
@@ -0,0 +1,137 @@
...
+	db := make(RecipeDB)
+	err := db.addRecipe("test-recipe", "1.2.3", true)
+	utiltest.AssertErrorMatch(t, err, nil)
+	memRecipe, ok := db.getRecipe("test-recipe")
+
+	readDB, err := newRecipeDB()
+	utiltest.AssertErrorMatch(t, err, nil)
+	fileRecipe, ok := readDB.getRecipe("test-recipe")
+	utiltest.AssertEquals(t, ok, true)
+	utiltest.AssertEquals(t, memRecipe, fileRecipe)
```
burov: Again, we're testing only single entry, it is not covering full purpose of the data structure.

burov (changes requested):
I see that most of the tests is actually not covering big portion of expected behavior, from my perspective tests should cover the actual usage and check proper side effects rather than just call code

policies/recipes/recipedb_test.go
bromivipo: I've updated the verification approach in this test to check the actual content instead of just the length, and added an extra test case for existing empty file.

policies/recipes/recipedb_test.go
bromivipo: Added more test cases to cover extra scenarios with stricter verification.

policies/recipes/recipedb_test.go
bromivipo: merged this test with the previous one

policies/recipes/recipedb_test.go
```diff
@@ -0,0 +1,241 @@
...
+				}
+			},
+			wantErr: nil,
+			wantDB: RecipeDB{
+				"recipe1": Recipe{Name: "recipe1", Version: recipeVersion{1}, InstallTime: 1, Success: true},
+				"recipe2": Recipe{Name: "recipe2", Version: recipeVersion{2}, InstallTime: 2, Success: false},
+			},
+		},
+		{
+			name: "invalid JSON, expect syntax error",
```
burov: NIT: group error cases together

policies/recipes/recipedb_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			db := tt.setup(t)
+
+			gotErr := db.addRecipe(tt.recipeName, tt.version, tt.success)
+			utiltest.AssertErrorMatch(t, gotErr, tt.wantErr)
+
+			// Verify initial entries are still there and unchanged (except the one being updated).
+			for _, wantRecipe := range tt.initialRecipes {
```
petercieslak: I was thinking if we could move this for loop to a separate helper function, something like `assertOtherDBEntriesUnchanged` to make the test a little bit more readable

bromivipo: Sure, moved these assertions to a helper function

petercieslak (dismissed):

policies/recipes/recipedb_test.go
```diff
@@ -0,0 +1,233 @@
...
+				return
+			}
+			utiltest.AssertEquals(t, ok, true)
+			utiltest.AssertEquals(t, recipe.Name, tt.recipeName)
+			utiltest.AssertEquals(t, recipe.Version.String(), tt.version)
+			utiltest.AssertEquals(t, recipe.Success, tt.success)
+
+			// Verify file persistence.
+			readDB := getTestDB(t)
+			utiltest.AssertEquals(t, db, readDB)
```
burov: Can it be done in a less chaty way ? 

My understanding of what we need to test:
1) setup - preload some initial state if any 
2) act - trigger method
3) verify - confirm that the end state is correct

What i see:
1) setup
2) act
3) verify that initial state is actually exists
4) verify that what we added is actually what we added
5) verify that end state is actually end state

I see here redundancy that bloat the test body and reduce readability of the code. I think if we get rid of the redundant check and rather check that end state is correct we should be good.

policies/recipes/recipedb_test.go
```diff
@@ -0,0 +1,233 @@
...
+			t.Fatalf("failed to write recipes to file: %v", err)
+		}
+	}
+
+	if len(contentBytes) > 0 {
+		if err := os.WriteFile(filepath.Join(tmpDir, dbFileName), contentBytes, 0644); err != nil {
+			t.Fatalf("failed to write recipes to file: %v", err)
+		}
+	}
+}
```
burov: This function is overcomplicated, it consists of the logic that add cognitive load that i don't think is actually contribute to stability or quality of the test.

My main questions is why we are not allowing writting empty contentRecipes to the file, as well why we're override contentRecipes with contentBytes if it is not empty, function is not suits single responsibility principle from my perspective, it should be simple and straightforward.

policies/recipes/recipedb_test.go
```diff
@@ -0,0 +1,233 @@
...
+func getTestDB(t *testing.T) RecipeDB {
+	t.Helper()
+	db, err := newRecipeDB()
+	if err != nil {
+		t.Fatalf("failed to load recipe DB: %v", err)
+	}
+	return db
+}
+
+func assertOtherDBEntriesUnchanged(t *testing.T, initialRecipes []Recipe, db RecipeDB, recipeName string) {
```
burov: I think this function is redundant, see comments below.

policies/recipes/recipedb_test.go
```diff
@@ -0,0 +1,233 @@
...
+			wantErr: func() error { var dummy interface{}; return json.Unmarshal([]byte(`invalid json`), &dummy) }(),
+			wantDB:  nil,
+		},
+		{
+			name: "one valid recipe and malformed entry, expect syntax error",
+			setup: func(t *testing.T) {
+				setupTestDB(t, nil, []byte(`[{"Name":"recipe1","Version":[1],"InstallTime":1,"Success":true}, invalid json]`))
+			},
+			wantErr: func() error {
+				var dummy interface{}
```
burov: NIT: empty, as well empty interface replaced with any in recent versions.

bromivipo: fixed

policies/recipes/recipedb_test.go
bromivipo: removed

policies/recipes/recipedb_test.go
bromivipo: Refactored it. Now `setupTestDB` only takes a content as a slice of bytes, sets up a db directory and creates a db file if content is not nil. Moved initialization from recipes to another function that marshals recipes and calls `setupTestDB`.

policies/recipes/recipedb_test.go
bromivipo: Fixed. The test body is shorter now. However, to verify the end state, I overwrite InstallTime in wantDB (because it is initialized as time.Now()). Is this approach okay, or would you prefer refactoring the source code to extract time.Now into a variable and overwrite it in the test?

policies/recipes/recipedb_test.go
bromivipo: fixed

policies/recipes/recipedb_test.go
```diff
@@ -0,0 +1,227 @@
...
+func getTestDB(t *testing.T) RecipeDB {
+	t.Helper()
+	db, err := newRecipeDB()
+	if err != nil {
+		t.Fatalf("failed to load recipe DB: %v", err)
+	}
+	return db
+}
+
+func TestNewRecipeDB(t *testing.T) {
```
burov: NIT: According to the standards, name should be Test_newRecipeDB

policies/recipes/recipedb_test.go
```diff
@@ -0,0 +1,227 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			db := tt.setup(t)
+
+			gotErr := db.addRecipe(tt.recipeName, tt.recipeVersion, tt.recipeSuccess)
+			utiltest.AssertErrorMatch(t, gotErr, tt.wantErr)
+
+			// Sync dynamic InstallTime timestamp before comparing DB states.
+			tt.wantDB = setInstallTime(tt.wantDB, tt.recipeName, db[tt.recipeName].InstallTime)
```
burov: Under the hood AssertEquals use go-cmp library

https://github.com/GoogleCloudPlatform/osconfig/blob/master/util/utiltest/utiltest.go#L117

go-cmp library can take an options, one of the options that can be useful here is https://pkg.go.dev/github.com/google/go-cmp@v0.7.0/cmp/cmpopts#IgnoreTypes.

You probably need to extend AssertEquals to take options as a variadic parameter.

bromivipo: Thanks for the advice!

I've added options as a parameter to AssertEquals and removed the hardcoded protocmp option. I also fixed its usage in agentendpoint_beta_test, since that's the only place that requires the protocmp option in AssertEquals as far as I can see.

policies/recipes/recipedb_test.go
bromivipo: Fixed. Does this naming convention apply to every unexported function? 
E.g., Test_newRecipeDB for newRecipeDB (unexported) vs. TestNewRecipeDB for NewRecipeDB (exported)?

burov (approved):

alexman111 (approved):

PR #991, opened 2026-05-21
Add unit tests for policies/policies.go PART 2

ganochenkodg:
Second part of unit tests for policies.go: TestSetConfigApt implementation.
Final test coverage: github.com/GoogleCloudPlatform/osconfig/policies/policies.go:125:	setConfig		90.9%

### DIFF
--- a/policies/policies.go
+++ b/policies/policies.go
@@ -38,6 +38,15 @@ import (
 	"cloud.google.com/go/osconfig/agentendpoint/apiv1beta/agentendpointpb"
 )
 
+var (
+	googetRepoFilePath = agentconfig.GooGetRepoFilePath
+	aptRepoFilePath    = agentconfig.AptRepoFilePath
+	yumRepoFilePath    = agentconfig.YumRepoFilePath
+	zypperRepoFilePath = agentconfig.ZypperRepoFilePath
+
+	retry = retryutil.RetryFunc
+)
+
 func run(ctx context.Context) {
 	var resp *agentendpointpb.EffectiveGuestPolicy
 
@@ -59,8 +68,9 @@ func run(ctx context.Context) {
 
 	effective := mergeConfigs(local, resp)
 
-	// We don't check the error from setConfig as all errors are already logged.
-	setConfig(ctx, effective)
+	if err := setConfig(ctx, effective); err != nil {
+		clog.Errorf(ctx, "%v", err)
+	}
 	if err := installRecipes(ctx, effective); err != nil {
 		clog.Errorf(ctx, "%v", err)
 	}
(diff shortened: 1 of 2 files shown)


policies/policies_test.go
```diff
@@ -207,3 +213,584 @@ func TestInstallRecipesHandlesInputs(t *testing.T) {
...
+	rpmQueryArgs := []string{"--queryformat", "\\{\"architecture\":\"%{ARCH}\",\"package\":\"%{NAME}\",\"source_name\":\"%{SOURCERPM}\",\"version\":\"%|EPOCH?{%{EPOCH}:}:{}|%{VERSION}-%{RELEASE}\"\\}\n", "-a"}
+	aptUpgradableArgs := []string{"--just-print", "-qq", "dist-upgrade"}
+	yumCheckUpdateArgs := []string{"check-update", "--assumeyes"}
+	yumListUpdatesArgs := []string{"update", "--assumeno", "--color=never"}
+	zypperListUpdatesArgs := []string{"--gpg-auto-import-keys", "-q", "list-updates"}
+	aptEnv := []string{"DEBIAN_FRONTEND=noninteractive"}
+
+	yumCheckUpdateErr := exec.Command("/bin/bash", "-c", "exit 100").Run()
+
+	tests := []struct {
```
petercieslak, line 237: It looks like we can separate cases for apt, yum, zypper and googet to separate tests (`TestSetConfig_apt`, `TestSetConfig_yum`, ...) and separate them into smaller PR's or at least commits so it is easier to review. I would appreciate it if you could do that :)

ganochenkodg: Yes, sure, I can do this

ganochenkodg: @petercieslak @burov  I have splitted this big PR into 4 smaller. This pr contains TestSetConfigApt.
Please review and merge this part at first, as it contains required helpers and policies.go refactoring (return errors instead of always nil)
And next targets:
TestSetConfigYum - https://github.com/GoogleCloudPlatform/osconfig/pull/996
TestSetConfigZypper - https://github.com/GoogleCloudPlatform/osconfig/pull/997
TestSetConfigGooget - https://github.com/GoogleCloudPlatform/osconfig/pull/998

policies/policies_test.go
```diff
@@ -207,3 +213,200 @@ func TestInstallRecipesHandlesInputs(t *testing.T) {
...
+type policiesStubOsInfoProvider struct {
+	osinfo osinfo.OSInfo
+}
+
+func (s policiesStubOsInfoProvider) GetOSInfo(ctx context.Context) (osinfo.OSInfo, error) {
+	return s.osinfo, nil
+}
+
+// setupSetConfigTest sets up the environment for SetConfig tests by mocking the command runner.
+func setupSetConfigTest(t *testing.T, apt, yum, zypper, googet bool, runner *utilmocks.MockCommandRunner) {
```
petercieslak: If we have to setup all those variables then let's keep this function, but if we don't then I would propose to replace it with an inline setup function in test arrays, which will contain overrides only for the package manager that we are testing.

petercieslak: Just to add a little bit to that - it is incredibly unlikely to have, for example, yum and googet packages enabled at the same time. That's why I don't think that it's necessary to have such a broad setup function.

ganochenkodg: I will rework it and ping you then

ganochenkodg: @petercieslak done, i've updated this and all next PRs

petercieslak (approved):

policies/policies.go
```diff
@@ -59,8 +68,9 @@ func run(ctx context.Context) {
 
 	effective := mergeConfigs(local, resp)
 
-	// We don't check the error from setConfig as all errors are already logged.
-	setConfig(ctx, effective)
+	if err := setConfig(ctx, effective); err != nil {
+		clog.Errorf(ctx, "%v", err)
```
burov, line 72: I assume it better to do a reasonable message here, like clog.Errorf("unable to setConfig: %s", err)

burov (approved):

PR #995, opened 2026-05-26
Add unit tests for util/util.go

bromivipo:
This PR contains unit tests for util.go
Initial coverage: 3.3%, current coverage 78.7%


util/util_test.go
```diff
@@ -41,6 +51,210 @@ func TestSanitizePath(t *testing.T) {
...
+			if tt.wantErr != nil {
+				if err == nil || !strings.HasPrefix(err.Error(), tt.wantErr.Error()) {
+					t.Errorf("expected error starting with %q, got %v", tt.wantErr.Error(), err)
+				}
+			} else {
+				utiltest.AssertErrorMatch(t, err, nil)
+			}
+			if err == nil {
+				utiltest.AssertFileContents(t, tt.input, string(tt.content))
+			}
```
iliatsuprik: can we use here 
```
utiltest.AssertErrorMatch(t, err, tt.wantErr)
utiltest.AssertFileContents(t, tt.input, string(tt.content))
```

util/util_test.go
```diff
@@ -41,6 +51,210 @@ func TestSanitizePath(t *testing.T) {
...
+					t.Errorf("expected error starting with %q, got %v", tt.wantErr.Error(), err)
+				}
+			} else {
+				utiltest.AssertErrorMatch(t, err, nil)
+			}
+			utiltest.AssertEquals(t, got, tt.want)
+
+			if err == nil {
+				utiltest.AssertFileContents(t, tt.input, tt.content)
+			}
```
iliatsuprik: can we use here AssertErrorMatch, AssertEquals and AssertFileContents without conditions? these functions should handle empty input

util/util_test.go
```diff
@@ -41,6 +51,210 @@ func TestSanitizePath(t *testing.T) {
...
+func TestHelperProcess(t *testing.T) {
+	if os.Getenv("GO_WANT_HELPER_PROCESS") != "1" {
+		return
+	}
+	if os.Getenv("GO_HELPER_FAIL") == "1" {
+		fmt.Fprint(os.Stderr, "error msg")
+		os.Exit(1)
+	}
+	fmt.Fprint(os.Stdout, "success msg")
+	os.Exit(0)
```
iliatsuprik: not sure if this test validates anything useful, probably we don't need it

iliatsuprik (approved):

util/util_test.go
bromivipo: it does not validate anything itself. But it is used as a separate process to validate DefaultRunner.

util/util_test.go
bromivipo: AssertErrorMatch could be used without conditions, but AssertFileContents fails if a file does not exist. So I moved error cases to separate tests.

util/util_test.go
burov: I agree with Ilia, there is no reason to keep the test that is not clear what it test to

util/util_test.go
```diff
@@ -41,6 +50,229 @@ func TestSanitizePath(t *testing.T) {
...
+		})
+	}
+}
+
+func TestAtomicWrite_Error(t *testing.T) {
+	tmpDir := t.TempDir()
+
+	tests := []struct {
+		name            string
+		input           string
```
burov: Please refactor TestAtomicWrite to the single tests, there is no reason to have two table tests with only one case,  testing the same function. It is complete against the purpose of the approach.

util/util_test.go
```diff
@@ -41,6 +50,229 @@ func TestSanitizePath(t *testing.T) {
...
+			gotChecksum, err := AtomicWriteFileStream(r, tt.checksum, tt.input, tt.mode)
+
+			utiltest.AssertErrorMatch(t, err, nil)
+			utiltest.AssertEquals(t, gotChecksum, tt.wantChecksum)
+			utiltest.AssertFileContents(t, tt.input, tt.content)
+		})
+	}
+}
+
+func TestAtomicWriteFileStream_Error(t *testing.T) {
```
burov: Same comment as below, please refactor tests to have one test and table testing should cover all success and failure cases

util/util_test.go
```diff
@@ -41,6 +50,229 @@ func TestSanitizePath(t *testing.T) {
...
+	}
+	if os.Getenv("GO_HELPER_FAIL") == "1" {
+		fmt.Fprint(os.Stderr, "error msg")
+		os.Exit(1)
+	}
+	fmt.Fprint(os.Stdout, "success msg")
+	os.Exit(0)
+}
+
+func TestDefaultRunnerRun(t *testing.T) {
```
burov, line 110: What this test is actually testing?

burov (commented):
I see there differen

util/util_test.go
bromivipo: Fixed

util/util_test.go
bromivipo: fixed

util/util_test.go
bromivipo: It verifies that DefaultRunner.Run captures stdout and stderr from the executed command and propagates execution errors. I've replaced the custom helper process with standard system commands to make it clearer.
Is there anything else you think this test should be checking?

util/util_test.go
```diff
@@ -41,6 +51,192 @@ func TestSanitizePath(t *testing.T) {
...
+		wantStdout string
+		wantStderr string
+		wantErr    string
+	}{
+		{
+			name:       "successful command execution, expect stdout output",
+			cmd:        testSuccessCmd(),
+			wantStdout: "success msg",
+			wantStderr: "",
+			wantErr:    "^<nil>$",
```
burov: wantErr by convention should be an error rather then regexp, we should avoid doing regexp comparison for errors as much as possible.

util/util_test.go
burov: AssertErrorMatch can be customized to accept options than custom comparer can help you to archive the thing you want, you can create custom comparer to compare errors by return value of Error() only. Keep in mind that you might need to craft custom filter as well to ensure it does not use exact type, but compatibility with error interface.

burov (changes requested):

bromivipo:
AssertErrorMatch now accepts custom cmp.Option parameters. Added `utiltest.EquateErrorMessage` for exact string equality and `utiltest.EquateErrorMessagePrefix` for matching errors with dynamic suffixes (such as temporary filenames or path errors), so we no longer use regexps for this.

Hopefully, I understood the intent correctly.

util/utiltest/utiltest.go
```diff
@@ -17,6 +18,23 @@ import (
...
+
+// EquateErrorMessage is a cmp.Option that compares errors by exact match of their Error() string representation.
+var EquateErrorMessage = cmp.FilterValues(filterErrors, cmp.Comparer(func(got, want any) bool {
+	return got.(error).Error() == want.(error).Error()
+}))
+
+// EquateErrorMessagePrefix is a cmp.Option that compares errors by checking if either error string is a prefix of the other.
+var EquateErrorMessagePrefix = cmp.FilterValues(filterErrors, cmp.Comparer(func(got, want any) bool {
+	gotErr, wantErr := got.(error).Error(), want.(error).Error()
+	return strings.HasPrefix(gotErr, wantErr) || strings.HasPrefix(wantErr, gotErr)
```
bromivipo: Both directions are checked to satisfy cmp.Diff's symmetry requirement.

util/utiltest/utiltest.go
```diff
@@ -16,6 +17,23 @@ import (
...
+	return gotOk && wantOk
+}
+
+// EquateErrorMessage is a cmp.Option that compares errors by exact match of their Error() string representation.
+var EquateErrorMessage = cmp.FilterValues(filterErrors, cmp.Comparer(func(got, want any) bool {
+	return got.(error).Error() == want.(error).Error()
+}))
+
+// EquateErrorMessagePrefix is a cmp.Option that compares errors by checking if either error string is a prefix of the other.
+var EquateErrorMessagePrefix = cmp.FilterValues(filterErrors, cmp.Comparer(func(got, want any) bool {
```
burov: I suggest avoiding using Prefix matching for errors, that bring us a risk of hiding problems rather than showing the exact failure reason.

Could you explain why we cannot check exact error here ?

util/utiltest/utiltest.go
burov: I guess we need to refactor usage of AssertError.* to avoid duplicating check, just add required options where it is required.

util/util_test.go
```diff
@@ -41,6 +52,188 @@ func TestSanitizePath(t *testing.T) {
...
+			wantErr:      errors.New("unable to create temp file"),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			reader := strings.NewReader(tt.content)
+			gotChecksum, gotErr := AtomicWriteFileStream(reader, tt.checksum, tt.input, tt.mode)
+
+			utiltest.AssertErrorMatchAndSkip(t, gotErr, tt.wantErr, utiltest.EquateErrorMessagePrefix)
```
burov: I guess that can be refactored to use exact error match, if not please explain why

burov (changes requested):

util/utiltest/utiltest.go
bromivipo: AtomicWrite functions append a randomized integer suffix based on time.Now() to the tmp file path and the errors contain those paths in .Error(), so the error messages are not predetermined. I could not find a way to compare exact errors without using regexp or prefix matching, so I think it's better to remove those two cases to avoid breaking the testing guidelines.  I don't think we'll lose much coverage here, as those test cases were essentially just checking error propagation from syscall failures. Additionally, I've added a positive edge case to both tests instead.

util/utiltest/utiltest.go
bromivipo: I've moved base assertion logic into an option which is now used by default when no other options are provided. Hope I understood your point correctly

util/util_test.go
bromivipo: replied above

util/utiltest/utiltest.go
burov: Sure

petercieslak (approved):

PR #1002, opened 2026-06-11
Add unit tests for repository_resource.go

bromivipo:
This PR contains unit tests for repository_resource.go


config/repository_resource_test.go
```diff
@@ -54,97 +58,189 @@ var (
...
+	utiltest.AssertErrorMatch(t, gotErr, nil)
+
+	mr := pr.ManagedResources()
+	utiltest.AssertEquals(t, len(mr.Repositories), 1)
+
+	aptRepo := mr.Repositories[0].Apt
+	utiltest.AssertEquals(t, aptRepo != nil, true)
+	utiltest.AssertEquals(t, aptRepo.GpgFilePath != "", true)
+	utiltest.AssertEquals(t, aptRepo.GpgChecksum != "", true)
+	utiltest.AssertEquals(t, len(aptRepo.GpgFileContents) > 0, true)
```
iliatsuprik: Is it possible to compare this with some expected values?

config/repository_resource_test.go
```diff
@@ -220,6 +316,70 @@ func TestRepositoryResourceCheckState(t *testing.T) {
...
+		setup              func(t *testing.T) ManagedRepository
+		wantInDesiredState bool
+	}{
+		{
+			name: "Apt GPG matches, expect InDesiredState=true",
+			setup: func(t *testing.T) ManagedRepository {
+				dir := t.TempDir()
+				repoPath := filepath.Join(dir, "repo")
+				gpgPath := filepath.Join(dir, "key.gpg")
+				os.WriteFile(repoPath, []byte("repo contents"), 0644)
```
iliatsuprik, line 271: you can probably move these 4 lines out of setup

iliatsuprik (approved):

config/repository_resource_test.go
bromivipo: done

config/repository_resource_test.go
bromivipo: I'm not sure if hardcoding an expected value for FileContents is the best approach here:
-  Hardcoding it will cause the test to fail if the GPG key changes externally. I'm not sure that it is static.
- Fetching the key during the test just to compare it with what Validate downloaded feels redundant, and the same applies to creating a direct mock of that behavior.

Instead, I've added a check to verify the dependency between FileContents, Checksum, and Path.

config/repository_resource_test.go
```diff
@@ -54,168 +56,267 @@ var (
...
+			utiltest.AssertEquals(t, pr.resource.(*repositoryResource).managedRepository, *tt.wantManagedRepository)
 		})
 	}
 }
 
 func TestRepositoryResourceCheckState(t *testing.T) {
-	ctx := context.Background()
-	var tests = []struct {
+	ctx := t.Context()
+	dir := t.TempDir()
```
burov: NIT: dir -> tmp

burov (dismissed):

config/repository_resource_test.go
bromivipo: Changed

burov (approved):

paulinakania (approved):

burov (approved):

PR #1004, opened 2026-06-15
Add unit tests for policies/recipes/steps_linux.go

ganochenkodg:
github.com/GoogleCloudPlatform/osconfig/policies/recipes/steps_linux.go:24:	mkCharDevice			100.0%
github.com/GoogleCloudPlatform/osconfig/policies/recipes/steps_linux.go:28:	mkBlockDevice			100.0%
github.com/GoogleCloudPlatform/osconfig/policies/recipes/steps_linux.go:32:	mkFifo				100.0%


policies/recipes/steps_linux_test.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+}
+
+// Test_mkFifo tests the mkFifo function.
+func Test_mkFifo(t *testing.T) {
+	tmpDir := t.TempDir()
+	var mode uint32 = 0666
+
+	alreadyExistsPath := utiltest.WriteToTempFileMust(t, "already_exists_fifo", []byte("test"))
+	t.Cleanup(func() { os.Remove(alreadyExistsPath) })
```
iliatsuprik: that is used in each test, so can be extracted into a function

iliatsuprik (approved):

policies/recipes/steps_linux_test.go
ganochenkodg: done

policies/recipes/steps_linux_test.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// setupPrepareTempFile writes content to a temporary file and schedules its removal.
+func setupPrepareTempFile(t *testing.T, name string) (string, string) {
+	t.Helper()
+	tmpDir := t.TempDir()
+	path := utiltest.WriteToTempFileMust(t, name, []byte("test"))
+	t.Cleanup(func() { os.Remove(path) })
+	return tmpDir, path
+}
```
burov, line 155: I see that effectively this function have only two lines, i think that should be inlined rather.

policies/recipes/steps_linux_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/GoogleCloudPlatform/osconfig/util/utiltest"
+	"golang.org/x/sys/unix"
+)
+
+// Test_mkCharDevice tests the mkCharDevice function.
+func Test_mkCharDevice(t *testing.T) {
```
burov, line 30: I don't think this test is actually required, there 0 logic in the function it tests, we just write tests to bump codecoverage, we should avoid that.

policies/recipes/steps_linux_test.go
```diff
@@ -0,0 +1,155 @@
...
+			gotErr := mkCharDevice(tt.path, major, minor)
+
+			utiltest.AssertErrorMatchAndSkip(t, gotErr, tt.wantErr)
+			checkOsStat(t, tt.path)
+		})
+	}
+}
+
+// Test_mkBlockDevice tests the mkBlockDevice function.
+func Test_mkBlockDevice(t *testing.T) {
```
burov, line 67: I don't think this test is actually required, there 0 logic in the function it tests, we just write tests to bump codecoverage, we should avoid that.

burov (changes requested):
That PR add tests for one line functions that have zero logic, we should avoid adding tests just to bump coverage as we discussed before

burov:
I think this should be closed

PR #1014, opened 2026-06-23
Add unit tests for zypper.go

bromivipo:
This PR contains unit tests for packages/zypper.go
Initial coverage 73.8%. Current coverage: 100%


packages/zypper_test.go
```diff
@@ -598,11 +598,238 @@ Conflicts   : [5]
...
-	if len(ppMap) > 0 {
-		t.Errorf("Unexpected result: expected no mappings, got = [%+v]", ppMap)
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockCtrl := gomock.NewController(t)
+			defer mockCtrl.Finish()
+
+			mockCommandRunner := utilmocks.NewMockCommandRunner(mockCtrl)
+			runner = mockCommandRunner
```
iliatsuprik: I suppose we can move lines 639-643 out of the test loop (that's done like this in most of the tests), so testing part would be shorter

packages/zypper_test.go
```diff
@@ -598,11 +598,238 @@ Conflicts   : [5]
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockCtrl := gomock.NewController(t)
+			defer mockCtrl.Finish()
+
+			mockCommandRunner := utilmocks.NewMockCommandRunner(mockCtrl)
+			runner = mockCommandRunner
```
iliatsuprik: same, can we move mockCommandRunner creation out of the loop. Can we also do this for other tests in this file?

iliatsuprik (approved):

packages/zypper_test.go
```diff
@@ -598,11 +598,237 @@ Conflicts   : [5]
...
+			patches:     []*ZypperPatch{{Name: "patch1"}},
+			expectedCmd: exec.Command(zypper, "info", "-t", "patch", "patch1"),
+			mockErr:     errors.New("generic error"),
+			wantErr:     errors.New(`error running /usr/bin/zypper with args ["info" "-t" "patch" "patch1"]: generic error, stdout: "", stderr: "stderr"`),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.expectedCmd != nil {
```
burov: I guess can be replaced with https://github.com/GoogleCloudPlatform/osconfig/blob/master/util/utiltest/utiltest.go#L208

packages/zypper_test.go
```diff
@@ -598,11 +598,237 @@ Conflicts   : [5]
...
+			stdout:  []byte("stdout"),
+			mockErr: errors.New("generic error"),
+			wantErr: errors.New(`error running /usr/bin/zypper with args ["--gpg-auto-import-keys" "--non-interactive" "install" "--auto-agree-with-licenses" "patch:patch1" "package:pkg1"]: generic error, stdout: "stdout", stderr: "stderr"`),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockCommandRunner.EXPECT().
+				Run(testCtx, utilmocks.EqCmd(expectedCmd)).
```
burov: testCtx -> ctx, see other cases as well

packages/zypper_test.go
```diff
@@ -598,11 +598,237 @@ Conflicts   : [5]
...
+		{
+			name:    "exit code 1, want exit error propagated",
+			mockErr: exitErr,
+			wantErr: exitErr,
+		},
+		{
+			name:    "generic error, want wrapped error",
+			stdout:  []byte("stdout"),
+			mockErr: errors.New("generic error"),
+			wantErr: errors.New(`error running /usr/bin/zypper with args ["--gpg-auto-import-keys" "--non-interactive" "install" "--auto-agree-with-licenses" "patch:patch1" "package:pkg1"]: generic error, stdout: "stdout", stderr: "stderr"`),
```
burov: Extract error to variable and give it reasonable name

packages/zypper_test.go
```diff
@@ -598,11 +598,237 @@ Conflicts   : [5]
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockCommandRunner.EXPECT().
```
burov: Same comment about util

packages/zypper_test.go
```diff
@@ -598,11 +598,237 @@ Conflicts   : [5]
...
+		{
+			name:        "all, want all flag in command",
+			options:     []ZypperListOption{ZypperListPatchAll(true)},
+			expectedCmd: exec.Command(zypper, append(zypperListPatchesArgs, "--all")...),
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockCommandRunner.EXPECT().
```
burov: Same comment about util

packages/zypper_test.go
```diff
@@ -598,11 +598,237 @@ Conflicts   : [5]
...
+				Run(testCtx, utilmocks.EqCmd(tt.expectedCmd)).
+				Return([]byte(""), []byte(""), nil).
+				Times(1)
+
+			_, _ = ZypperPatches(testCtx, tt.options...)
+		})
+	}
+}
+
+func TestParseZypperPatchInfo_Errors(t *testing.T) {
```
burov: Avoid defining separate tests fro handling only negative cases, test should cover both success/fail case.

bromivipo: Merged `TestParseZypperPatchInfo_Errors` with two existing functions `TestParsePatchInfo`  and `TestParsePatchInfo_differentFormatsOfConflictPkgsVersions` into `TestParseZypperPatchInfo`. Changed the way mappings are verified and also shortened patchInfo mocks without losing coverage for edge cases.

packages/zypper_test.go
bromivipo: fixed

packages/zypper_test.go
bromivipo: fixed

packages/zypper_test.go
bromivipo: Fixed both in this and one other test

packages/zypper_test.go
bromivipo: Removed global testCtx usage across the file and initialized local context in each test function instead

packages/zypper_test.go
bromivipo: Fixed

packages/zypper_test.go
```diff
@@ -525,84 +443,313 @@ Interactive : ---
...
+		},
 	}
-	if len(ppMap) > 0 {
-		t.Errorf("Unexpected result: expected no mappings, got = [%+v]", ppMap)
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			utiltest.SetExpectedCommands(ctx, mockCommandRunner, tt.expectedCommands)
+
+			_, _ = ZypperPatches(ctx, tt.options...)
```
burov: Why we're not testing return parameters ?

bromivipo: Added the return parameter checks. Initially, I didn't test the return values here because I viewed this test primarily as a check for proper command construction with various options.

petercieslak (approved):

PR #1019, opened 2026-07-06
Add unit tests for scalibr.go

bromivipo:
This PR contains unit tests for packages/scalibr.go

Initial coverage 73.8%, current coverage 88.5%


packages/scalibr_test.go
```diff
@@ -132,12 +151,20 @@ func withZypperDisabled(t *testing.T) {
 
 func TestScalibrIntegration(t *testing.T) {
 	withZypperDisabled(t)
```
ganochenkodg: I suppose remove this function and use here utiltest.OverrideVariable(t, &ZypperExists, false)
I see it's used in only one place

packages/scalibr_test.go
```diff
@@ -132,12 +151,20 @@ func withZypperDisabled(t *testing.T) {
...
+
 	tests := []struct {
+		name     string
+		setup    func(t *testing.T)
 		provider InstalledPackagesProvider
 		wantErr  error
 		wantPkgs Packages
 	}{
 		{
+			name:  "successful scan, expect packages",
```
ganochenkodg, line 160: pls add wantErr: nil

packages/scalibr_test.go
```diff
@@ -149,16 +176,103 @@ func TestScalibrIntegration(t *testing.T) {
...
+			provider: scalibrInstalledPackagesProvider{
+				osinfoProvider: stubProvider{},
+				extractors:     []string{"invalid/extractor"},
+				scanRootPaths:  []string{arrangeVirtualRoot(t, "./testdata/debian.dpkg-status", "/var/lib/dpkg/status")},
+				dirsToSkip:     []string{},
+			},
+			wantErr: errors.New("unknown plugin \"invalid/extractor\""),
+		},
+		{
+			name:  "skipped directory, expect no packages",
```
ganochenkodg, line 197: pls add wantErr: nil

burov (approved):

paulinakania (approved):
LGTM

packages/scalibr_test.go
```diff
@@ -124,41 +137,138 @@ func (stubProvider) GetOSInfo(ctx context.Context) (osinfo.OSInfo, error) {
...
+type errorProvider struct{}
+
+func (errorProvider) GetOSInfo(ctx context.Context) (osinfo.OSInfo, error) {
+	return osinfo.OSInfo{}, errors.New("osinfo error")
 }
 
 func TestScalibrIntegration(t *testing.T) {
-	withZypperDisabled(t)
+	utiltest.OverrideVariable(t, &ZypperExists, false)
+	mockCtrl := gomock.NewController(t)
```
ekremenetskii, line 148: correct me if I'm wrong, but it seems like we are using the parent `t` for the `mockCtrl` here. if a mock fails, wouldn't the error be reported for the entire `TestScalibrIntegration` rather than the specific `t.Run` test case?

maybe we should move the initialization of `gomock.NewController` and `mockCommandRunner` inside the t.Run loop (around line 266) so that each sub-test has its own controller?

packages/scalibr_test.go
```diff
@@ -124,41 +137,138 @@ func (stubProvider) GetOSInfo(ctx context.Context) (osinfo.OSInfo, error) {
...
+	t.Cleanup(mockCtrl.Finish)
+	mockCommandRunner := utilmocks.NewMockCommandRunner(mockCtrl)
+
 	tests := []struct {
+		name     string
+		setup    func(t *testing.T)
```
ekremenetskii, line 154: i might be missing something here, but since `mockCommandRunner` is shared across all test cases, aren't the `.EXPECT()` calls from different cases piling up on the exact same mock object (for example, on lines 213 and 237)? it looks like this only works right now because the tests run strictly sequentially

wouldn't it be safer to isolate them? we could change the signature to `setup func(t *testing.T, runner *utilmocks.MockCommandRunner)` and pass a freshly created mock from inside `t.Run`. what do you think?

PR #1034, opened 2026-08-04
Add Configuration for SCALIBR

bromivipo:
This PR adds configuration support for SCALIBR extended inventory collection in `agentconfig`.

### Changes Included:
* **Metadata keys added:** `osconfig-extended-inventory-enabled`, `osconfig-extended-inventory-collection-interval`, `osconfig-extended-inventory-extractors-allowed`.
* **Parsing & Overrides:** Added `setExtendedInventory` to support instance-level overrides for project settings, along with a `parseSlice` helper for extractors.
* **Tests:** Unit tests added in `agentconfig_test.go` for defaults, project settings, and instance overrides.

### DIFF
--- a/agentconfig/agentconfig.go
+++ b/agentconfig/agentconfig.go
@@ -62,12 +62,11 @@ const (
 
 	prodEndpoint = "{zone}-osconfig.googleapis.com.:443"
 
-	osInventoryEnabledDefault                  = false
-	guestPoliciesEnabledDefault                = false
-	taskNotificationEnabledDefault             = false
-	debugEnabledDefault                        = false
-	extendedInventoryEnabledDefault            = false
-	extendedInventoryCollectionIntervalDefault = 10
+	osInventoryEnabledDefault       = false
+	guestPoliciesEnabledDefault     = false
+	taskNotificationEnabledDefault  = false
+	debugEnabledDefault             = false
+	extendedInventoryEnabledDefault = false
 
 	oldConfigDirLinux = "/etc/osconfig"
 	cacheDirLinux     = "/var/lib/google_osconfig_agent"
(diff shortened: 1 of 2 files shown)


iliatsuprik (approved):

agentconfig/agentconfig.go
```diff
@@ -122,25 +124,28 @@ var (
...
+	numericProjectID                    int64
+	osConfigPollInterval                int
+	debugEnabled                        bool
+	taskNotificationEnabled             bool
+	guestPoliciesEnabled                bool
+	osInventoryEnabled                  bool
+	scalibrLinuxEnabled                 bool
+	guestAttributesEnabled              bool
+	traceGetInventory                   bool
+	extendedInventoryEnabled            bool
```
bromivipo: If we're updating the SCALIBR flow, am I right in understanding that we can drop scalibrLinuxEnabled and rely on extendedInventoryEnabled for both linux and windows?

agentconfig/agentconfig.go
```diff
@@ -122,25 +124,28 @@ var (
...
+	extendedInventoryCollectionInterval int
```
bromivipo: Is this field necessary, or could we reuse the existing osConfigPollInterval ticker instead of adding a new interval configuration? Since we discussed merging the results of extended and legacy scans, it makes sense for them to run simultaneously.

agentconfig/agentconfig.go
petercieslak: No, `scalibrLinuxEnabled` is a flag to enable extraction using `osv-scalibr` and `extendedInventoryEnabled` is used to enable extraction with extended set of extractors (other than currently supported).

agentconfig/agentconfig.go
petercieslak: That's a good point, it seems that we can reuse the existing ticker

bromivipo: removed it

petercieslak (approved):

agentconfig/agentconfig.go
ganochenkodg: please correct me if i'm wrong. if we have scalibrLinuxEnabled = true - we just start to use current default extractors ("os/cos", "os/dpkg", "os/rpm"). with additional extendedInventoryEnabled = true we start to use other extractors by default, language extractors for example. and extendedInventoryExtractorsAllowed is used to filter those additionally activated extractors. @petercieslak is that correct?

my other theory - extendedInventoryExtractorsAllowed must be non-empty to use any additional extractors with extendedInventoryEnabled, otherwise will be no effect.

petercieslak: The first description is correct, with scalibrLinuxEnabled we use the default extractors. Additionally extendedInventoryEnabled allows us to use extendedInventoryExtractorsAllowed in the extraction process. The extendedInventoryExtractorsAllowed don't have to be non-empty - if that's the case we just use the default extractors.

burov (approved):

PR #1038, opened 2026-08-13
scalibr linux new metadata support

ganochenkodg:
Add support for new extended inventory metadata keys on linux

### DIFF
--- a/agentconfig/agentconfig.go
+++ b/agentconfig/agentconfig.go
@@ -33,8 +33,11 @@ import (
 	"sync"
 	"time"
 
+	"testing"
+
 	"cloud.google.com/go/compute/metadata"
 	"github.com/GoogleCloudPlatform/osconfig/clog"
+	"github.com/GoogleCloudPlatform/osconfig/util/utiltest"
 	"golang.org/x/oauth2/jws"
 )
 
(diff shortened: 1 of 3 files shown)


packages/packages_linux.go
```diff
@@ -280,3 +288,12 @@ func NewInstalledPackagesProvider(osinfoProvider osinfo.Provider) InstalledPacka
 		osinfoProvider: osinfoProvider,
 	}
 }
+
+func contains(slice []string, s string) bool {
```
burov: I guess can be replaced with https://pkg.go.dev/slices#Contains

burov (dismissed):

packages/packages_linux.go
ganochenkodg: fixed

petercieslak (dismissed):

petercieslak:
What about tests for this change?

ganochenkodg:
> What about tests for this change?

Added, but it required to add an additional function in main code, check pls

petercieslak (dismissed):

agentconfig/agentconfig.go
```diff
@@ -880,3 +883,15 @@ func DisableCloudLogging() bool {
 func UniverseDomain() string {
 	return getAgentConfig().universeDomain
 }
+
+// SetTestScalibrConfig sets SCALIBR configuration flags for testing
+func SetTestScalibrConfig(t *testing.T, scalibrLinux bool, extendedEnabled bool, allowedExtractors []string) {
```
burov: Production code should not depends on the testing code, testing.T should be added only to the code that is suffixed with _test.go

burov (changes requested):

agentconfig/agentconfig.go
ganochenkodg: Fixed

petercieslak (approved):

PR #1044, opened 2026-08-25
Add scalibr os benchmarks

ganochenkodg:
New test TestScalibrBenchmark to test os/dpkg, os/rpm, os/cos and os/snap extractors. Sample command to run the test:
```
sudo go test -v -tags=benchmark -run=TestScalibrBenchmark ./packages/ 2>&1 | grep ' | '
```

Sample output on debian 13:
```
    scalibr_bench_test.go:211: | Scenario | Avg Scan Time | Avg Heap Alloc | Peak RAM RSS | Peak CPU % | Mean CPU % | Pkgs Found |
    scalibr_bench_test.go:212: | --- | --- | --- | --- | --- | --- | --- |
    scalibr_bench_test.go:183: | `All os extractors` | 1.531515958s | 138.47 MB | 22.58 MB | 199.8% | 115.9% | 442 |
    scalibr_bench_test.go:183: | `os/dpkg` | 1.481452628s | 124.24 MB | 22.60 MB | 204.1% | 115.0% | 442 |
    scalibr_bench_test.go:183: | `os/rpm` | 1.455361571s | 119.27 MB | 21.58 MB | 204.6% | 114.5% | 0 |
    scalibr_bench_test.go:183: | `os/cos` | 1.442813717s | 119.29 MB | 21.09 MB | 195.3% | 116.0% | 0 |
```

### DIFF
--- a/packages/scalibr_bench_test.go
+++ b/packages/scalibr_bench_test.go
@@ -1,0 +1,224 @@
+//  Copyright 2026 Google Inc. All Rights Reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+
+//go:build benchmark
+
+package packages
+
+import (
+	"context"
+	"fmt"
+	"runtime"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/GoogleCloudPlatform/osconfig/osinfo"
+	"github.com/GoogleCloudPlatform/osconfig/util/utiltrace"
+)
+
+type cpuSample struct {
+	totalCPUTime time.Duration
+	time         time.Time
+}
+
+// getCPUTime returns total (user + system) CPU time consumed by the process.
+func getCPUTime() (time.Duration, error) {
+	var usage syscall.Rusage
+	if err := syscall.Getrusage(syscall.RUSAGE_SELF, &usage); err != nil {
+		return 0, fmt.Errorf("getrusage error: %w", err)
+	}
+	user := time.Duration(usage.Utime.Sec)*time.Second + time.Duration(usage.Utime.Usec)*time.Microsecond
+	sys := time.Duration(usage.Stime.Sec)*time.Second + time.Duration(usage.Stime.Usec)*time.Microsecond
+	return user + sys, nil
+}
+
+type traceMetricsResult struct {
+	utiltrace.TraceMemoryResult
+	CPUPeakPercent float64
+	CPUMeanPercent float64
+	Duration       time.Duration
+	AllocMB        float64
+}
+
+func traceMetrics(ctx context.Context, interval time.Duration, resultChan chan<- traceMetricsResult) {
+	startTime := time.Now()
+	var memBefore runtime.MemStats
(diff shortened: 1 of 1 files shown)


packages/scalibr_bench_test.go
petercieslak, line 1: Can we move the code to a dedicated `utiltrace` package?

ganochenkodg: Just checked if it's possible. Unfortunately, no:
1. Circular import dependency. The **packages** package already imports github.com/GoogleCloudPlatform/osconfig/util/utiltrace (in trace.go). So **import packages** in util/utiltrace/ produces an error - import cycle not allowed
2. The benchmark relies on scalibrInstalledPackagesProvider, which is an unexported struct defined inside the **packages** package and cannot be accessed outside of it without code refactoring

petercieslak (dismissed):

packages/scalibr_bench_test.go
```diff
@@ -0,0 +1,232 @@
...
+		avgDuration,
+		avgAllocMB,
+		avgPeakRAM,
+		avgPeakCPU,
+		avgMeanCPU,
+		res.pkgsCount,
+	)
+}
+
+func TestScalibrBenchmark(t *testing.T) {
```
burov, line 194: In Golang there a special API to run the benchmarks testing.B, why don't we use that ?

packages/scalibr_bench_test.go
```diff
@@ -0,0 +1,232 @@
...
+	t.Log("### SCALIBR Extractors Benchmark Results")
+	t.Log("=========================================================================")
+	t.Log("| Scenario | Avg Scan Time | Avg Heap Alloc | Peak RAM RSS | Peak CPU % | Mean CPU % | Pkgs Found |")
+	t.Log("| --- | --- | --- | --- | --- | --- | --- |")
+
+	runs := 3
+	for _, bm := range benchmarks {
+		var total benchResult
+		for i := 0; i < runs; i++ {
+			res, err := runSingleBenchmark(ctx, osinfoProvider, bm.extractors)
```
burov, line 218: runBenchmark should be enough

packages/scalibr_bench_test.go
burov: I guess if it help we can expose scalibrInstalledPackagesProvider

packages/scalibr_bench_test.go
ganochenkodg: fixed

packages/scalibr_bench_test.go
ganochenkodg: Even if we expose scalibrInstalledPackagesProvider, moving the benchmark test itself into package utiltrace still causes a circular dependency: packages imports utiltrace in production code (trace.go), so utiltrace cannot import packages.

What we can do here, I think, is moving `traceMetrics` helper (calculating CPU + RAM metrics) into utiltrace, while keeping scalibr benchmark in packages/. `traceMetrics` itself doesnt depend on any packages code and will not produce import cycles.

packages/scalibr_bench_test.go
ganochenkodg: I chose `testing.T` over `testing.B` because this suite is an OS-level macro profiler rather than an in-memory micro-benchmark:

1. testing.B automatically scales `b.N` (1, 2, 5, 10, 50...) as i know until it reaches -benchtime. scalibr scanning is a heavy fs operation; running dozens of consecutive scans may take an excessively long time and skew memory measurements. In Go, heap memory and RSS are not immediately reclaimed by the OS between iterations, so high `b.N` would measure accumulated garbage rather than the true memory footprint of an isolated scan. In our setup, we run a fixed number of iterations (3) with runtime.GC() between runs.

2. If testing.B executes dozens of iterations, all package database files,i think, remain cached in the Linux page cache after the first run, masking realistic I/O behavior.

So I think testing.T is more suitable here, but if it's necessary I can refactor this code to try to use testing.B instead

petercieslak (approved):

PR #1052, opened 2026-09-10
Add support for all possible SCALIBR os extractor for linux

ganochenkodg:
Add support for os/snap extractor + test

### DIFF
--- a/packages/packages.go
+++ b/packages/packages.go
@@ -104,6 +104,12 @@ type Packages struct {
 	Zypper             []*PkgInfo            `json:"zypper,omitempty"`
 	ZypperPatches      []*ZypperPatch        `json:"zypperPatches,omitempty"`
 	COS                []*PkgInfo            `json:"cos,omitempty"`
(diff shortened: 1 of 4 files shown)


ganochenkodg:
@petercieslak @burov which of these extractors do we need to add in osconfig? Snap looks ok as for me, spack - not sure. And I'm absolutely not sure about Alpine/Arch/Gentoo/Nixos support as they are not supported by GCP now and functionality of such VMs will be restricted.

petercieslak:
> @petercieslak @burov which of these extractors do we need to add in osconfig? Snap looks ok as for me, spack - not sure. And I'm absolutely not sure about Alpine/Arch/Gentoo/Nixos support as they are not supported by GCP now and functionality of such VMs will be restricted.

We should not support extractors for images not supported in GCP. From the provided list only Snap should be added IMO.

burov (approved):

petercieslak (approved):

PR #1055, opened 2026-09-18
Fix failing to report REBOOTING state during postPatchReboot due to c…

petercieslak:
…ontext canceled for Windows VMs


agentendpoint/patch_task.go
```diff
@@ -196,10 +196,6 @@ func (r *patchTask) rebootIfNeeded(ctx context.Context, prePatch bool) error {
 		return nil
 	}
 
-	if err := r.reportContinuingState(ctx, agentendpointpb.ApplyPatchesTaskProgress_REBOOTING); err != nil {
```
burov, line 199: Why do we need move that ?

petercieslak: This block is moved after `r.saveState()` is called. For more context, currently there is a race condition between `main_windows.go` which calls `ctx.cancel` and `rebootIfNeeded()`, which would use the canceled context for reporting REBOOTING state. In this case `reportContinuingState()` would fail and `r.saveState()` is never reached, meaning we will fail to report the state. The fix is to move this block after `r.saveState()` and use `context.WithoutCancel()` for `reportContinuingState()`

